Issue #5, opened 2017-03-28
Basic SSO Implementation


chpapa:
# Description

Have official plugins for common SSO support; And also make it easy to enable and config them from Skygear Portal.

# Related Issues

# Portal Design

## Portal Features

- List of official plugins:
  - Google
  - Facebook
  - Wechat
  - OAuth
- Have an interface for on/off and configuration of ID / Secret.

## UX Design
**Put (link to) Portal Design here**

# API Design

## Scenario

**To be expanded**
* Sign in with web popup
* Sign in with web redirect
* Sign in with limited capability devices (TV/Command line)
* Sign in with Mobile
* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, show an error and tell users to login with another `AuthProvide B`
* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, tell users to login and link with new AuthProvider
* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, assume it is two different accounts (will break the assumption of Skygear, which each users got unique email address)
* At User Setting page, add another `AuthProvider`

## Sample Code

**Put sample code of how you vision this API will be used, consider different type of developers and different abstract level**

## API Design

This is the API I am proposing

### JS

- `loginWithOAuthProvider(providerID, options, [accessToken])`
  - Create or login a new skygear user, associated with the provider
  - `providerID` - A string that identify the login provider
  - `options` 
    - uxMode - Either `popup`(default), or `redirect`
    - clientID
    - scope
    - redirectUrl - when uxMode is `redirect`, skygear will redirect the user to this url after auth. If it is null, back to the current URL
  - `accessToken` - Optional. Pass in access token if client already has it, skygear will try to login directly instead of going through the OAuth flow.
  - This function returns a skygear user, and an access token of the service.
- `associateAccountWithProvider(providerID, options)`
  - Add a new auth provider to the user by going through the auth flow
  - This API requires user to be logged in already, return error otherwise
  - `providerID` - A string that identify the login provider
  - `options` 
    - uxMode - Either `popup`(default), or `redirect`
    - clientID
    - scope
    - redirectUrl - when uxMode is `redirect`, skygear will redirect the user to this url after auth. If it is null, back to the current URL
  - This function returns a skygear user, and an access token of the new service.
- `getOAuthTokens()`
  - Return a promise of tokens

        getOAuthTokens().then(function(tokens){
          //tokens['com.facebook'] is FB's access token
        });

[...]

PR #63, opened 2017-05-14
Add specification for basic SSO

b123400:
connect https://github.com/SkygearIO/features/issues/5

### DIFF
--- a/features/5-basic-sso/api.md
+++ b/features/5-basic-sso/api.md
@@ -195,13 +195,13 @@ It will be used to generate urls when user auth.
 Since the map is in plugin, user should be able to extend it easily.
 
 ```
-@oauth_auth('com.custom.website')
-def auth():
-  return "http://example.com/auth"
+@skygear.provides("oauth", "com.facebook")
+class FacebookOAuthProvider
+  def auth_url(self):
+    return "http://example.com/auth"
 
-@oauth_access_token('com.custom.website')
-def access_token(code):
-  return "http://example.com/access_token?code="+code
+  def access_token_url(self, code):
+    return "http://example.com/access_token?code="+code
 ```
 
 Pseudo code of `oauth:handle_access_token`:
@@ -254,6 +254,26 @@ def handleAccessToken(token):
   - Otherwise,
     - Login or create new users according to the provider and access token
   - Return the user
+
+### API for integrating other services
+
+- New decorator `@skygear.provides("oauth", <provider id>)`
+  - Register a new class that returns urls for OAuth flow
+  - Example in the above session
+  - The goal is to let users to integrate any website that support OAuth
+
+### Functions for working with 3rd party services after auth
+
+- New method `container.getOAuthTokens()`
+  - Return a dictionary of OAuth tokens of the current user
+  - Pseudo code
+
+        container.getOAuthTokens()['com.facebook'].access_token
+
+- New method `container.refreshToken(provider_id)`
+  - A convenient method that refreshs access_token if needed, and if possible.
+  - Updates database after refresh
+  - Returns the new token
 
 # Database Scheme
 
--- a/features/5-basic-sso/api.md
+++ b/features/5-basic-sso/api.md
@@ -229,7 +229,9 @@ def handleAccessToken(token):
 
 ## APIs
 
-- `oauth:auth_url`
+### APIs for auth flow
+
+- New lambda `oauth:auth_url`
   - Accepts a provider id and options
   - Return an url for auth
   - Example pseudo return value
@@ -238,7 +240,7 @@ def handleAccessToken(token):
         client_id={app-id}
         &redirect_uri=http%3A%2F%2Fskygear.dev%2Foauth%2Fhandle_code%3Fprovider%3Dcom.facebook%26user_id%3D123
 
-- `oauth:handle_code`
+- New handler `oauth:handle_code`
   - A handler that accepts code from 3rd party service
   - Exchange code with access token
   - Create user if needed
@@ -247,7 +249,7 @@ def handleAccessToken(token):
 
         http://skygear.dev/oauth/handle_code?provider=com.facebook&user_id=123&code=223344
 
-- `oauth:handle_access_token`
+- New lambda `oauth:handle_access_token`
   - Accepts a provider id, and access token
   - If this handler is called with a user logged in
     - Associate the user with auth provider and access token
@@ -278,6 +280,26 @@ def handleAccessToken(token):
 # Database Scheme
 
 No changes.
+Since we are going to use the existing auth provider to handle auth data, I expect those related data to be saved in `_user` table's `auth` column. There is no specific format for the data.
+
+I am proposing the following format for data in the auth column.
+
+- Data from OAuth providers will be saved with a "oauth." prefix in id
+- The data inside will be almost the same as the [OAuth Access Token Response](https://tools.ietf.org/html/rfc6749#section-4.2.2), e.g.
+  - `access_token`
+  - `token_type`
+  - `expires_at`, calculated from `expires_in`, an absolute timestamp
+  - `refresh_token`
+
+      {
+        "oauth.com.facebook": {
+          "access_token": "...",
+          "token_type": "bearer",
+          "expires_at": 1495052619,
+          "refresh_token": "...."
+        }
+      }
+
 
 # Others Supplement Information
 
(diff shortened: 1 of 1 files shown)


features/5-basic-sso/api.md
```diff
@@ -0,0 +1,207 @@
...
+  - Accepts a provider id, and access token
+  - If this handler is called with a user logged in
+    - Associate the user with auth provider and access token
+  - Otherwise,
+    - Login or create new users according to the provider and access token
+  - Return the user
+
+# Database Scheme
+
+No changes.
```
rickmak: How to save the token?

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,207 @@
...
+## APIs
+
+- `oauth:auth_url`
+  - Accepts a provider id and options
+  - Return an url for auth
+- `oauth:handle_code`
+  - A handler that accepts code from 3rd party service
+  - Exchange code with access token
+  - Create user if needed
+  - Pass the user back to client
```
rickmak: I am gear towards OAuth standard flow is implementing at skygear-server. And the plugin is adding `handle_access_token` with actual implementation is calling standard OAuth implementation. i.e. I will not use `handle_access_token` for all SDK callback of fb/google. I will have `handle_fb_access_token` and `handle_google_access_token` (Which is calling handler_code from plugin)

I am not sure does it make sense to your proposal.

`Create user if needed` skygear-server already have this code.  I don't want plugin implement once again.

rickmak: In addition, I am thinking `UNIQUE_EMAIL_FOR_ACCOUNTS` behavior is core and implement at skygear-server. Not plugin.

features/5-basic-sso/api.md
b123400: I plan to save it in the same `_user.auth` column along with normal auth data, is it ok?

features/5-basic-sso/api.md
b123400: Are you expecting OAuth related url to be hardcoded in `skygear-server`?

e.g. something like this

```
func facebookAuthUrl() String {
  return "https://www.facebook.com/v2.9/dialog/oauth?client_id={app-id}&redirect_uri={redirect-uri}"
}

func exchangeCodeForAccessTokenForFacebook(code) String {
  url := "https://graph.facebook.com/v2.9/oauth/access_token?
   client_id={app-id}
   &redirect_uri={redirect-uri}
   &client_secret={app-secret}
   &code={code}"
  return performReques(url)
}
```

features/5-basic-sso/api.md
rickmak: I think we will need a format and example for that. And @chpapa may not agree to save those things unstructured at `_user.auth`.

features/5-basic-sso/api.md
rickmak: I don't get why it direct to hard code question.

Is it related to how to implement one OAuth related endpoint to multiple OAuth providers?

b123400: Because I want to know more exactly, what do you mean by "And the plugin is adding handle_access_token with actual implementation is calling standard OAuth implementation"

features/5-basic-sso/api.md
b123400: Your comment [here](https://github.com/SkygearIO/features/issues/5#issuecomment-300395577) mentioned:

> I am proposing `oauth:handle_code` is implement at Skygear-server, while other `oauth:handle_access_token` can be model using current `auth:login` with auth provider.

Since we will be using auth provder, I think it make sense to save it in `_user.auth`.

b123400:
I updated the proposal to include more details from offline discussion yesterday

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,254 @@
+# API Design Overview
+
+Have official plugins for common SSO support; And also make it easy to enable and config them from Skygear Portal.
+
+Portal needs to have an interface for on/off and configuration of ID / Secret.
+
+## Scenario
```
chpapa: Would expect the Scenario to be expanded. For example, we have included 3rd Party SDK as an expanded use case during the discussion.

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,254 @@
...
+* Sign in with web popup
+* Sign in with web redirect
+* Sign in with limited capability devices (TV/Command line)
+* Sign in with Mobile
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, show an error and tell users to login with another `AuthProvide B`
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, tell users to login and link with new AuthProvider
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, assume it is two different accounts (will break the assumption of Skygear, which each users got unique email address)
+* At User Setting page, add another `AuthProvider`
+
+# Sample Codes for Use Cases
```
chpapa: Would expect more sample code to illustrate different use cases too.

features/5-basic-sso/api.md
chpapa: I agree it seems `UNIQUE_EMAIL_FOR_ACCOUNTS` should be on server side, but just an impression, haven't give much thought in.

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,254 @@
...
+## Scenario
+
+* Sign in with web popup
+* Sign in with web redirect
+* Sign in with limited capability devices (TV/Command line)
+* Sign in with Mobile
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, show an error and tell users to login with another `AuthProvide B`
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, tell users to login and link with new AuthProvider
+* Login by `AuthProvider A`, but the email already got another account from `AuthProvider B`, assume it is two different accounts (will break the assumption of Skygear, which each users got unique email address)
+* At User Setting page, add another `AuthProvider`
```
chpapa: As discussed in Product Meeting, will need a use case/samples code of using oauth_token in Cloud Functions.

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,254 @@
...
+
+```
+@oauth_auth('com.custom.website')
+def auth():
+  return "http://example.com/auth"
+
+@oauth_access_token('com.custom.website')
+def access_token(code):
+  return "http://example.com/access_token?code="+code
+```
```
rickmak: If we following authProvider convention, we are using class decorator for group of relating function.

https://docs.skygear.io/guides/advanced/auth-provider/python/

b123400: Oh yes this is much better

b123400:
Updated.

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,338 @@
...
+  skygear.getOAuthTokens().then(function(tokens){
+    //tokens['com.example'] is FB's access token
+  });
+})
+```
+
+Cloud code to add an OAuth provider
+
+```
+@skygear.provides("oauth", "com.facebook")
```
rickmak: `skygear.OAuthProvider` ?

features/5-basic-sso/api.md
```diff
@@ -0,0 +1,338 @@
...
+
+  def access_token_url(self, code):
+    return "http://example.com/access_token?code="+code
+```
+
+Access OAuth tokens in Cloud code:
+
+```
+container = SkygearContainer(user_id='1', api_key='master')
+tokens = container.getOAuthTokens()
```
rickmak: This def of `getOAuthTokens` will be in core? Or plugin.

We need a logical separation of where the code belongs to.

features/5-basic-sso/api.md
b123400: I don't understand. You mean it should be `@skygear.OAuthProvider("com.facebook")`?

features/5-basic-sso/api.md
b123400: I hope to put it in plugin so everything is in the same place

features/5-basic-sso/api.md
rickmak: Yes, naming problem. Agree?

features/5-basic-sso/api.md
rickmak: Seem anti-pattern if we implement it in plugin. Since the code will not in py-skygear, it is not convention to inject method in python.

Probably you will have a helper method?

features/5-basic-sso/api.md
b123400: `@skygear.OAuthProvider("com.facebook")`
- Capital letter O makes it look like a class
- It is not consistent with how we register auth provider. Why is normal Auth provider using `@skygear.provide` but not OAuth?

features/5-basic-sso/api.md
b123400: Oh I get what you mean, right, we at least need an extra `user:oauth_token` lambda from `skygear-server`, client side SDK needs that as well.

b123400:
Updated, added a new function so provider can provide an url for refreshing tokens.
So it looks like this now.

```
@skygear.provides("oauth", "com.facebook")
class FacebookOAuthProvider
  def auth_url(self):
    return "http://example.com/auth"

  def access_token_url(self, code):
    return "http://example.com/access_token?code="+code

  def refresh_token_url(self, refresh_token):
    return "http://example.com/token?grant_type=refresh_token&refresh_token=" + refresh_token
```

rickmak:
Follow up at https://github.com/SkygearIO/features/pull/73

Issue #25, opened 2017-03-28
Forgot password and welcome email in Skygear portal


chpapa:
# Description

Plugins to implement forget password and welcome emails features

There is an existing plugins: https://github.com/skygeario/forgot_password/

# Design/Specification

* Workflow:
  * Call `forgotPassword(email)` from client-side
  * The respond is always success even if the email address was not found, if `FORGOT_PASSWORD_SECURE_MATCH` is set to TRUE. (Hence users are expected to display something like *If a matching account was found an email was sent to user@email.com to allow you to reset your password.* after forgot password was requested.
  * `FORGOT_PASSWORD_SECURE_MATCH` is set to FALSE by default, which will return ResourceNotFound if the email doesn't match.
  * If `FORGOT_PASSWORD_WELCOME_EMAIL` is set to TRUE, send an *Welcome Email* when new user signed up.
* Portal Setting (Under `User Auth` > `Forgot Password` / `Welcome Email`)
  * Setting (Enable, Sender, From, Reply To)
  * Welcome Email Template (sent when new user signed up and enabled)
  * Forgot Password Email Template (sent when password reset was requested)
  * Reset Password HTML Template
  * Reset Password Success HTML Template
  * Reset Password Error HTML Template
  * Each Email Template / HTML Template in Portal have a "Try" button for a preview.
  * Each template should have the following fields:
    * From Email Address
    * Subject
    * Reply to
    * Content
    * Redirect URL (only applicable for emails)
    * URL Lifetime (only applicable for HTML templates)
  * Content are expected to be an HTML / plain text editor with placeholder variable. The list of placeholder variable should include all the fields under the User table, and something like `[[url]]` to include link to the confirmation page.

## Sample code

```
SKYContainer.default().forgotPassword(email: “chpapa@gmail.com”)
skygearContainer.forgotPassword(“chpapa@gmail.com”)
skygear.forgotPassword(“chpapa@gmail.com”)
```

# Related Issues

- Server Issues
  - SkygearIO/forgot_password#9
  - SkygearIO/forgot_password#10
  - SkygearIO/forgot_password#11
  - SkygearIO/forgot_password#12
- Client Issues
  - SkygearIO/skygear-SDK-Android#123
  - SkygearIO/skygear-SDK-JS#192
  - SkygearIO/skygear-SDK-iOS#92
- Portal Issues
  - https://trello.com/c/M7QhTYqM/118-skygear-forgot-password-and-welcome-email-portal-setting
[...]

PR #68, opened 2017-05-20
Add design of the new user profile

ben181231:


features/user-profile.md
```diff
@@ -0,0 +1,83 @@
...
+        be discovered.
+-->
+
+# Changes on SDK
+
+The SDKs would be expected to have the following changes:
+
+- All log in / sign up method would returns **a user record**, instead of
+  a `_user` object.
+- remove `getUserByEmails()`, `getUserByUsernames()`, `discoverUserByEmails()`
```
cheungpat: Suggestion: Since user discovery by email and and username appears to be common use case, I think we should retain these two function. Alternative is to create helper functions that return a query object that returns profile when executed.

features/user-profile.md
```diff
@@ -0,0 +1,83 @@
...
+  and `discoverUserByUsernames()` in Skygear Container. These operations can be
+  replaced by record query operation on user record
+- remove `saveUser()` method. It can be replaced by record save operation on
```
cheungpat: Suggestion: `saveUser` implements a use case that allows the developer to change user role. I think we should remove `saveUser` but I also think we should add helper functions that could change the user role.

features/user-profile.md
```diff
@@ -0,0 +1,83 @@
...
+  user record.
+
+# Database Scheme
+
+<!-- TODO: Provide migration SQL -->
```
cheungpat: Comment:  Migration SQL is part of the implementation IMO (i.e. Not needed here)

ben181231:
This PR is ready to review.

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+| User  | AnyUser  | email       | ReadOnly    | Discoverable   |
+| User  | AnyUser  | status      | ReadOnly    | Discoverable   |
+| User  | AnyUser  | profile_pic | ReadOnly    | NotQueryable   |
+| User  | AnyUser  | game_score  | ReadOnly    | Queryable      |
+| User  | AnyUser  | birthday    | NoAccess    | NotQueryable   |
+
+## Authentication Object
+
+The new authentication object (i.e. `_auth` object) is to replace the old
+ `_user` object, to avoid ambiguity with the user record.
```
cheungpat: Also create a view called `_user` iff migrating the schema from an old version. The `_user` view should resemble the `_user` table in previous schema version.

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+userQuery.equalTo('status', 'online');
+skygear.publicDB.query(userQuery)
+.then((users) => {
+  console.log(`Found ${users.length} high score users online`);
+});
+```
+
+## New sign up / log in API
+
+The usage of new sign up / log in API is shown as followed:
```
cheungpat: s/as followed/as follows/

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+
+```js
+const skygear = require('skygear');
+
+// Use Case 1: sign up and setting user profile
+skygear.signup(
+  { 'username': 'user001' }
```
cheungpat: missing comma at the end

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
+# API Design Overview
+
+This document introduces the new Skygear User Record (Profile), which is
+re-designed from user record and user object, utilizing user profile discovery.
+The design depends on [the field-based ACL](field-based-acl.md) feature.
+
+## New User Record
+
+The newly designed user record contains different information of user.By
```
cheungpat: s/user.By/user. By/

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+setting field-based ACL, user record can serve as both public profile and
+private profile.
```
cheungpat: this should read “user record can contain public attributes and private attributes”

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+The following sample code shows how to use user discovery API for different
+purposes:
+
+```js
+const skygear = require('skygear');
+
+// Use Case 1: discover user by username
+skygear.discoverUserByUsername('cheungpat')
+.then((users) => {
+  if (users.length) {
```
cheungpat: probably `users.length === 0`?

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+  }
+).then((user) => {
+  console.log(`Successfully sign up with user ID: ${user.id}`);
+}).catch((err) => {
+  console.error(`Failed to sign up: ${err.message}`);
+});
+
+// Use Case 2: sign up with custom data
+skygear.signup(
+  { username: 'user002', role: 'doctor' },
```
cheungpat: need to describe the behavior when there are multiple auth fields, esp. for uniqueness

features/user-profile.md
```diff
@@ -0,0 +1,226 @@
...
+- `container.discoverUserByUsernames(usernames)`
+- `container.discoverUserByEmail(email)`
+- `container.discoverUserByEmails(emails)`
+
+## New Sign Up and Log In API
+
+The new sign up API is proposed to enable: 1) sign up with custom data;
+2) update user profile during sign up. The new sign up API will perform:
+
+1. Discover user with the provided data
```
cheungpat: Need to describe what happens if the column for the auth data does not exists

ben181231: If the column of the auth data does not exists, it will be regarded as "no user is found" on dev mode, regarded as request error on production mode.

features/user-profile.md
ben181231: Added "Authentication Record Configurations" for setting uniqueness of the auth fields. (Better to generalize this configuration to all record, in another feature issue)

features/user-profile.md
```diff
@@ -0,0 +1,258 @@
...
+Since Skygear Server does not have a way for developers to set the uniqueness
+constraint of record fields (except writing the `before_save` hook of the
+record), configurations are proposed for authentication record.
+
+Authentication Record Configurations should contains: 1) Authentication Record
+Type; 2) Authentication Record Fields. They are suggested to be provided by
+environment variables during implementation:
+
+- `AUTH_RECORD_TYPE`: the record type for authentication, default: `user`
+- `AUTH_RECORD_FIELDS`: the record field names of the record for
```
cheungpat: I’d recommend you do this:

AUTH_RECORD_FIELDS_1=slug
AUTH_RECORD_FIELDS_2=username,role
AUTH_RECORD_FIELDS_3=email,role 
(or if the above doesn’t exists)
AUTH_RECORD_FIELDS=username,email

features/user-profile.md
```diff
@@ -0,0 +1,258 @@
...
+  authentication. The format will be a comma separated list while each list
+  item is either a field name or another comma separated list wrapped with a
+  pair of brackets. (e.g. `slug,(username,role),(email,role)`). The default
+  value would be `username,email`.
+
+The authentication record field configuration defines the following behavior:
+
+- the list of unique constraints on the corresponding table of the
+  authentication record.
```
cheungpat: when is the unique constraint created?

ben181231: Every time when skygear server starts on dev mode.

features/user-profile.md
ben181231: Then developers need to maintain the array index in the env. var. name ?

chpapa:
@cheungpat @ben181231 @rickmak 

Sorry for late comments there, I have two questions/thoughts:

1. For the auth_record setting, I kind of guessed why it was designed like this and how it is going to be used technically. But do we want to have some idea of how the final portal might look like so we make sure this model is easy to use? I would imagine we need to handle these:
    1. Common case: Login by email/login by username/login by phone no.
    2. Special case: HouseDoctor's (username, role)
2. I have barely go over it before, but have we thought about shall we call it "discoverUser" and "Discoverable" (in #69 )? Is it the best name? Have we considered "Getable" or what? Just want to make sure we try to use a name that's easy to grab.

features/user-profile.md
cheungpat: Yes?

features/user-profile.md
cheungpat: i.e. check if the unique constraint exists, and then create it if not found?

if the default is `username,email`, what happens when the skygear server starts on the dev mode for the first time?

features/48-user-profile-discovery/user-profile.md
```diff
@@ -0,0 +1,258 @@
...
+## User discovery API
+
+The following sample code shows how to use user discovery API for different
+purposes:
+
+```js
+const skygear = require('skygear');
+
+// Use Case 1: discover user by username
+skygear.discoverUserByUsername('cheungpat')
```
rickmak: skygear.discoverUserByUsername('cheungpat')> return single user?
skygear.discoverUserByUsernames(['cheungpat'])> return array of  user?

features/48-user-profile-discovery/user-profile.md
```diff
@@ -0,0 +1,258 @@
...
+Since Skygear Server does not have a way for developers to set the uniqueness
+constraint of record fields (except writing the `before_save` hook of the
+record), configurations are proposed for authentication record.
+
+Authentication Record Configurations should contains: 1) Authentication Record
+Type; 2) Authentication Record Fields. They are suggested to be provided by
+environment variables during implementation:
+
+- `AUTH_RECORD_TYPE`: the record type for authentication, default: `user`
+- `AUTH_RECORD_FIELDS`: the record field names of the record for
```
rickmak: I don't understand this....

rickmak: @ben181231 Any comments?

features/48-user-profile-discovery/user-profile.md
ben181231: Yes. Or we may just keep `discoverUserByUsernames()` by the design principle: "There should be only one obvious way to do things".

features/48-user-profile-discovery/user-profile.md
ben181231: This is a environment variable to tell Skygear which combinations of fields can be used to log in / sign up. And Skygear should keep the uniqueness of those combinations.

features/48-user-profile-discovery/user-profile.md
chpapa: On the Design Principle, should be if find one users / find a list of users is "one thing" or "different things"?  I think it might justify for some use cases which need a list of users quickly.

However I'm not sure if the word "discover" is good as mentioned above.

features/48-user-profile-discovery/user-profile.md
chpapa: `AUTH_RECORD_TYPE` maybe a bit redundant, do users really want to / have a case to change the record type from `user` to other stuff?

`AUTH_RECORD_FIELDS` maybe need a better name?

ben181231: Expected that most of the user won't set `AUTH_RECORD_TYPE`. Since the current implementation allows changing the record type associating with the `_user` (default is `user` record), this env. var. just exposes it to developers, providing flexibility.

For `AUTH_RECORD_FIELDS`, any good suggestions ?

chpapa: @ben181231 I think @cheungpat suggestion somewhere (I can't find it now...) of changing it to: `AUTH_RECORD_FIELDS` / `AUTH_RECORD_FIELDS_1` / `AUTH_RECORD_FIELDS_N` is better.

And in that case, maybe we can call it (here are a list of my brain storm):

* `AUTH_RECORD_UNIQUE_IDENTIFIER` (Too Long? And also UID might mixed up with the real user_id for users?)
* `AUTH_RECORD_KEYS` / `AUTH_RECORD_INDEXS` (Borrow the concept from Relational DB)
* `AUTH_RECORD_UNIQUE_KEYS` (又 unique 又 keys 應該明?)

BTW, a reminder, our implementation should still have a unique user_id (use a UUID?), don't use this environment variable as keys...

chpapa: Just to leave a note of reminder, I think this change will change #25 forgotPassword too. @ben181231 @rickmak 

Forgot Password now basically assume one parameter of email address is enough to uniquely identify the users, and trigger the forgot password process.

With this change, the API will need to take arbitrary arguments (is it even possible in ObjC and Java?), to uniquely identify the current users.

Thoughts?

PR #69, opened 2017-05-20
Add api design of field-base ACL

ben181231:


features/field-based-acl.md
```diff
@@ -0,0 +1,190 @@
...
+| User  | Owner    | gender | ReadWrite |
+
+- Make gender field of user record private to every one, readable to friends
+  and updatable to owner
+
+| Class |  UserRole   | Field  |   Level   |
+|-------|-------------|--------|-----------|
+| *     | AnyUser     | *      | ReadWrite |
+| User  | AnyUser     | gender | NoAccess  |
+| User  | Ref:Friends | gender | ReadOnly  |
```
cheungpat: need more info about `Ref:Friends`

features/field-based-acl.md
```diff
@@ -0,0 +1,190 @@
...
+1. **Any Users**: any logged-in users
+
+## Base ACL
+
+Since the field-based ACL is allow-based, the following entities would be added
+serving as a base ACL.
+
+| Class | UserRole | Field |   Level   |
+|-------|----------|-------|-----------|
+| *     | AnyUser  | *     | ReadWrite |
```
ben181231: This refers to all fields in all records. So even new fields or new records are added by JIT, this rule should also be applied.

> Is this implicit or explicit?

I think it is open to implementor. For explicit declaration, Skygear server would throw an exception if no rules is found during accessibility checking. But it gives a flexibility to update the base ACL. For implicit inference, Skygear server would make the assumption in code level and not open to update.

features/field-based-acl.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Accessibility levels
+
+The field-based accessibility is divided into 5 levels:
+
+- `ReadWrite`: fields that available for read and write
+- `Queryable`: Fields that can be used to perform any query predicates and
+  will be returned after query.
+- `Discoverable`: Fields that can only be used to perform equality query
+  predicate and will be returned in query.
```
cheungpat: > Also, I think contains operator is not allowed for discoverable-only fields.

You need to allow people to discover user profiles with email like `email IN ["johndoe@example.com", "janedoe@example.com", "johnny.appleseed@example.com"]`.

Interestingly that also means the user can do `gender IN ["male", "female", "indeterminate"]`.

If a field is discoverable, it is possible to enumerate all combinations by making separate requests **if** the number of combinations is small, no matter you support `IN` or not. Therefore, I think we should allow `IN` because the common use case is to discover by email.

ben181231: Okay. I see.

features/field-based-acl.md
```diff
@@ -0,0 +1,243 @@
...
+| User  | Owner    | gender | ReadWrite   | Queryable      |
+
+- Make gender field of user record private to every one, readable and queryable
+  to friends and updatable to owner
+
+| Class |  UserRole   | Field  | AccessLevel | DiscoveryLevel |
+|-------|-------------|--------|-------------|----------------|
+| *     | Public      | *      | ReadWrite   | Queryable      |
+| User  | AnyUser     | gender | NoAccess    | NotQueryable   |
+| User  | Ref:Friends | gender | ReadOnly    | Queryable      |
```
cheungpat: `Ref:Friends` need update to clarify it is not related to friends relation

perhaps: `col:group_members`?

features/field-based-acl.md
```diff
@@ -0,0 +1,243 @@
...
+
+An table would be created for saving field-based access control entities. The
+schema would be like:
+
+```SQL
+CREATE TABLE "_recrod_field_access" (
+    "record_type" text NOT NULL,
+    "record_field" text NOT NULL,
+    "user_role" text NOT NULL,
+    "level" integer NOT NULL,
```
cheungpat: I think you should make this into columns instead of bitfield.

features/field-based-acl.md
chpapa: I remember at some stage someone suggested that we can just make a discover API for this, would that proposal simpler? (a.k.a. Discoverable would meant developer can use the discover API to retrieve record, but not Query at all)

Otherwise I think a strict definition of Query like what proposed here is fine too.

features/field-based-acl.md
```diff
@@ -0,0 +1,243 @@
...
+
+```json
+{
+  "action": "schema:record_field_access",
+  "api_key": "my-api-key",
+  "access_token": "admin-user-token",
+  "record_type": "Photo",
+  "record_fields": ["slug"],
+  "user_roles": ["_any_user"],
+  "level": 15,
```
cheungpat: I think you should make this into fields instead of bitfield.

features/field-based-acl.md
```diff
@@ -182,46 +182,38 @@ development mode by users with admin role.
...
+  "action": "schema:record_field_access:update",
   "api_key": "my-api-key",
   "access_token": "admin-user-token",
   "record_type": "Photo",
   "record_fields": ["slug"],
   "user_roles": ["_any_user"],
-  "level": 15,
+  "writable": true,
+  "readable": true,
+  "comparable": true,
```
cheungpat: good use comparable

ben181231:
@rickmak Updated

features/field-based-acl.md
```diff
@@ -246,73 +141,59 @@ discoverable by others
...
-  "record_fields": ["gender"],
-  "user_roles": ["_any_user"],
-  "for_all_record_types": false,
-  "for_all_record_fields": false
+  "action": "schema:field_access:update",
+  "api_key": "my-master-key",
+  "access": [
+    {
+      "record_type": "User",
+      "record_field": "*",
```
rickmak: `*`? Are we going to save the `*` at DB or fan out the setting on update?

features/field-based-acl.md
```diff
@@ -0,0 +1,190 @@
...
+  "action": "schema:record_field_access",
+  "api_key": "my-api-key",
+  "access_token": "admin-user-token",
+  "record_type": "Photo",
+  "record_fields": ["slug"],
+  "user_roles": ["_any_user"],
+  "level": "Discoverable",
+  "for_all_record_types": false,
+  "for_all_record_fields": false
+}
```
rickmak: The level format?

features/field-based-acl.md
ben181231: Should save `*` to DB in order to cover the fields created later.

Issue #70, opened 2017-05-23
SDK API Grouping


cheungpat:
# Description

Following a discussion from skygeario/skygear-SDK-JS#220,

* A `Container` class that exposes all API.
* The API have to be discoverable by the app developer, probably by way of looking at class doc, or through IDE autocompletion.
* The API have to be grouped into feature-set so that app developer can find an API by narrowing down to the set of API via the grouping. Grouping is achieved by way of putting API of the same feature-set into a separate class to be exposed by the `Container`.
* The API exposed by all languages SDK have to be the same, meaning the same set of API has to be available within the same grouping and exposed via similar mechanisms.
* The API organization is totally app-developer-driven, so how the plugin code is organized doesn’t matter.

# API Design

> Note: this is not real code, do not copy from these code, these are for illustration purpose only.

```objective-c
// Objective-C
#import <SKYKit/SKYKit.h>

SKYContainer *container = [SKYContainer defaultContainer];
[container.auth loginWithEmail:@"email@example.com" password:@"secret" completion:nil];
[container.auth forgotPasswordWithEmail:@"email@example.com" completion:nil];
[container.pubsub subscribeTo:"channel" handler:nil];
```

```java
// Java
import android.content.Context;
import io.skygear.skygear.Container;
import io.skygear.skygear.Pubsub;

Container container = Container.defaultContainer(Context.getApplicationContext());
container.auth.loginWithEmail("email@example.com", "secret");
container.auth.forgotPassword("email@example.com");
container.pubsub.subscribe("channel", new Pubsub.Handler())
```

```javascript
// JS
var skygear = require('skygear');

skygear.auth.loginWithEmail("email@example.com", "secret");
skygear.auth.forgotPassword("email@example.com");
skygear.pubsub.on("channel", function() {});
```

# Open Questions

* What are the list of groupings? (e.g. `auth`, `pubsub`, `db`)
* How the existing APIs are grouped?
* When we roll out the change (i.e. when API are grouped), do we maintain backward compatibility?

IMO these questions are best answered by @chpapa who has better understanding of what cloud developer needs. The grouping are especially important as we need a grouping that makes sense to cloud developer as well as making them future proof when we need to add more APIs in the future.

When we have a list of grouping. I will try to group the existing the APIs and see if everyone agrees on them.

# Related Issues

- Client Issues
  - [ ] SkygearIO/skygear-SDK-JS#231
  - [ ] SkygearIO/skygear-SDK-iOS#104
[...]

PR #71, opened 2017-05-25
Add design of new SDK API grouping

Steven-Chan:


features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,257 @@
...
+      privateDB(Database)
+      cacheResponse(Boolean)
+
+    api:
+      setAdminRole
+      setDefaultRole
+      getDefaultACL
+      setDefaultACL
+      setRecordCreateAccess
+      setRecordDefaultAccess
```
cheungpat: @chpapa mentioned that ACL configuration should goes to public db.

Pros: Explicit that public DB support ACL while private DB does not.
Cons: We will have a `PublicDatabase` and `PrivateDatabase` class (instead of just one `Database` class).

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,257 @@
...
+
+  Database:
+
+    state:
+      cacheStore
+
+    api:
+      getRecordByID
```
cheungpat: Simply `fetch` in JS.

@rickmak thinks user should use `query` instead.

cheungpat:
This PR is for https://github.com/SkygearIO/features/issues/70

features/70-sdk-api-grouping/api.md
chpapa: just want to clarify, is that these ACL setting really have no effect on privateDB? (Sorry I know I should have look at the code but I was told....) if yes I guess we can either just move them to public DB, or if we want to keep the interface of both DB classes, we can throw some sort of error / warning when people call them on privateDB?

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,257 @@
...
+
+Container container = Container.defaultContainer(Context.getApplicationContext());
+String apiKey = container.apiKey();
+String currentAccessToken = container.auth().currentAccessToken();
+Map publicDBCache = container.db().public().cache();
+```
+
+# Changes on SDK
+
+Proposal:
```
rickmak: Is this proposal universal to all SDK?

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,257 @@
...
+## Java
+
+```java
+// Call API
+import android.content.Context;
+import io.skygear.skygear.Container;
+import io.skygear.skygear.db.Record;
+
+Container container = Container.defaultContainer(Context.getApplicationContext());
+container.auth().loginWithEmail("email@example.com", "secret");
```
rickmak: Callback missing here.

rickmak:
In addition, I am expecting this PR containing the proposal on how to inject the namespace? What is the standard way to inject a new group of feature?

Or we can create another issue for that.

features/70-sdk-api-grouping/api.md
cheungpat: @Steven-Chan I am under the impression that this proposal is only applicable to JS.

cheungpat:
> In addition, I am expecting this PR containing the proposal on how to inject the namespace? What is the standard way to inject a new group of feature?

I talked with @Steven-Chan about this and my suggestion is that we will modify the `Container` to add new grouping instead of injecting functions/methods to an existing class or by use of reflection.

@Steven-Chan Could you also clarify how the classes will be declared in terms of package organisation? That means Objective-C Cocoapod specs, JavaScript npm packages and Android class packages.

I expect that everything needed by a normal user (i.e. `auth()`, `analytics` etc) will be available when the user import/require/add `skygear`, `SKYKit` or `io.skygear.?` package.

Steven-Chan:
> I expect more implementation detail in this:
> 
> auth broadcasts/or global state. refs: #70 (comment)
> In addition, I am proposing we at lease have a comparing example code from similar services. In this case, firebase? to illustrate how similiar/different are we.

> In addition, I am expecting this PR containing the proposal on how to inject the namespace? What is the standard way to inject a new group of feature?
> 
> Or we can create another issue for that.

The `auth broadcasts/or global state` thing is based on the idea that plugin are able to be injected/imported separately, i.e. skygear user may choose which groups they import.

I suggest that this PR would focus on re-organizing the api. This should be required if we want to implement injecting parts dynamically. But we did not come up a good conclusion on how it may be done on Java yet, so I think we should open a new issue on that.

features/70-sdk-api-grouping/api.md
Steven-Chan: Indeed, those api written are based on JS SDK. That's why `push` is missing.

I would work on each sdk, and add a more detail part for each language. The current one should be considered as the big picture.

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,316 @@
...
+    logout
+    changePassword
+    forgotPassword
+    saveUser
+    whoami
+    getUsersByEmail
+    getUsersByUsername
+    discoverUserByEmails
+    discoverUserByUsernames
+    queryFriend
```
cheungpat: I think `queueFriend` belongs to public db.

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,316 @@
...
+skygear.auth.forgotPassword("email@example.com");
+skygear.pubsub.on("channel", function() {});
+
+var SampleItem = skygear.Record.extend('SampleItem');
+var sampleItem = new SampleItem();
+skygear.db.public.save(sampleItem);
+```
+
+```js
+// Retreieve contexts
```
cheungpat: s/retreieve/retrieve/

features/70-sdk-api-grouping/api.md
```diff
@@ -0,0 +1,316 @@
...
+## Cloud
+
+```
+Container:
+
+  state:
+    cloud(CloudContainer)
+
+
+CloudContainer:
```
cheungpat: I believe cloud is out of scope, I don’t mind that you put it here.

PR #72, opened 2017-05-31
Add cloud function invoke features

cheungpat:
connects #42


features/42-cloud-function-log-isolation/api.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+## Log Isolation
+
+Log isolation refers to obtaining console log from invoked cloud function. Log
+is to be isolated from each invocation of cloud functions.
+
+Since log is printed to stdout, we need to run cloud function in separate
+process so that log for individual invocation can be collected. This requires
+some changes across server and plugin runtime.
```
ben181231: Do we need to handle the state of the plugins? The plugins have their own lifecycles, i.e. register (using `includeme`) and init (using `init` event).

features/42-cloud-function-log-isolation/api.md
```diff
@@ -0,0 +1,256 @@
...
+
+Other options to invoke cloud function:
+
+```
+$ skycli help cloud function invoke
+Usage: skycli cloud function invoke [options]
+    --again  Run with payload from previous invocation
+    --log    Print log to stderr
+
+$ skycli cloud function invoke note.add_info --again
```
cheungpat: how about a file inside `~/.skycli/`?

features/42-cloud-function-log-isolation/api.md
cheungpat: (This is largely from python point of view, I am not sure if it is possible in JS)

The separate process will be forked from the main process after loading all the modules. By that point the `includeme` functions are already called and the in-memory state will carry over to the forked process. The `init` event is initiated by the server, hence the forked process will not receive this event.

features/42-cloud-function-log-isolation/api.md
ben181231: Good idea to put it inside `~/.skycli/`

features/42-cloud-function-log-isolation/api.md
ben181231: That seems great. JS can have a similar trick.

cheungpat:
Removed the log isolation part because that takes more time to study. The rest should be trivial.

cheungpat:
@rickmak

features/42-cloud-function-log-isolation/api.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+`--record-type` and `--trigger` is not required if the name is unique.
+
+An editing session is opened with some content to help the user invoke
+cloud function without entering boilerplate. The editing session should
+be prepopulated with either:
+
+- Templated params according to the record type and trigger
+- Payload from previous invocation for the same cloud function if any
```
rickmak: This looks handy. The previous payload is from the same type, right?

i.e. we previous payload from `skycli get record` will not have effect on the prepopulated content of `skycli cloud function invoke hook`.

So we will save the previous invocation by command group?

cheungpat: The saved previous payload is specific to that cloud function, so it will not be affected by other `skycli` commands nor other cloud function of the same type.

That is,

```
skycli cloud function invoke hook add_info --record-type note --trigger before_save
```

will be different from 

```
skycli cloud function invoke hook add_info --record-type note --trigger after_save
```

rickmak:
@ben181231 please confirm your comment is addressed and approve this PR. I will merge it.

ben181231:
I think this PR is good to me.

PR #77, opened 2017-06-06
[WIP] Add log isolation features

cheungpat:
refs #42


cheungpat:
We may do this later.

PR #81, opened 2017-06-07
WIP: Add feature doc on structured logging

cheungpat:


features/79-structured-logging/structured-logging.md
```diff
@@ -0,0 +1,69 @@
...
+Logger is the primary field for filtering log entries.
+
+Proposing the following list of loggers:
+
+* server - skygear-server log not otherwise covered by other loggers
+* server.sql - skygear-server log for executed sql statement and result
+* server.http - skygear-server log for http request and response body
+* server.transport skygear-server log for plugin transport
+* plugin - plugin log not otherwise covered by other loggers
+* plugin.transport - plugin log for plugin transport
```
rickmak: I am not sure the different of `server.transport` and `plugin.transport`.

I assume `plugin.transport` are referring python or js side logging.

features/79-structured-logging/structured-logging.md
```diff
@@ -0,0 +1,69 @@
...
+* cloudfunc - cloud function log generated by developer code
```
rickmak: cloudfunc? Is it refer to portal trigger cloudfunc?

Seem we will have some plugin name in the structured logging? Which is not mentioned in this PR.

Should be something like the `request_id`. Or put it into the name of the logger, like `plugin.forgot_password.transport`?

I lean toward `plugin.forgot_password.transport`, but just gut feeling. I am open to both option.

features/79-structured-logging/structured-logging.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Log format
+
+Will adopt JSON log format.
+
+## Configuration
+
+TBD: Skygear-server can be configured to turn off logs based on logger and
+level.
```
rickmak: We already have some configuration.

IMO the format itself should already sufficient for this PR. All we need is ON/OFF and LEVEL.

features/79-structured-logging/structured-logging.md
```diff
@@ -0,0 +1,69 @@
...
+* If log entry is not in JSON format, wrap the log entry in JSON format with
+  common stuctured log data.
+* Log processor to print log to container stdout.
+
+## Portal
+
+* Print JSON-formatted log entries in a nice format (table or otherwise)
+* Support filtering using the `logger`
+* Colorize log output based on `logger` (or other fields)
+* Filter log output by selecting interested `logger` (likely multi-select)
```
rickmak: Seem we will want a wireframe to process with the portal. So the scope will be clear.

chpapa:
Just want to say I have read it but don't know what to comment lol. I think the main thing is the logs grouping (the logger in the spec), would it make sense for the filter used by users.

features/79-structured-logging/structured-logging.md
Steven-Chan: I get the feeling that this category is to separate `developer code` from non `developer code`.

PR #85, opened 2017-06-14
Add partial record update proposal

ben181231:


oursky-travis:
@ben181231, thanks for your PR! By analyzing the history of the files in this pull request, we identified @cheungpat to be a potential reviewer.

features/48-user-profile-discovery/field-based-acl.md
```diff
@@ -107,10 +107,50 @@ function recordOperationVeridation (operation) {
...
+Skygear SDKs have not many changes since the setup of field-based ACL would be
+expected on Skygear Cloud Portal or Skygear CLI, which will be specified on
+another document.
+
+The record save API on Skygear SDKs would be updated to support partial record
+fields update on non-atomic request.
+
+```js
+// saving one record
+skygear.publicDB.query(record1)
```
ben181231: Oh... should be `skygear.publicDB.save`.

features/48-user-profile-discovery/field-based-acl.md
```diff
@@ -107,10 +107,50 @@ function recordOperationVeridation (operation) {
...
+.then((savedRecord) => {
```
rickmak: I am suggesting `.then((savedRecord) => {` ==> `.then((savedRecord, error) => {`. For array of records, `.then((savedRecords) => {` ==> `.then((savedRecords, errors) => {`

So the code will become follow:

```
then((savedRecord, error) => {
  console.log(`Successfully saved record: ${savedRecord}`);
  console.warn(`Rejected saving fields: ${error.$rejectedFields}`)
}, (err) => {
  console.error(`Failed to save: ${err}`);
});
```

Since `savedRecord` is expected to be an instance of `Record`, adding `$rejectedFields` to the class will add knowledge to the `Record` class. It is not necessary and not good to make the user need to understand field ACL upfront.

Second, we are designing for a smooth normal flow. The API will encourage people to separate normal code and error handling code. And on first draft of their application, the API should not require them to handle/understand every error before they ship a demo.

ben181231: I also think that is a better interface. However, limited by the interface of [`Promise.prototype.then()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then), the resolve function can only take one argument. We cannot add the rejected fields as an error on the 2nd argument unless we override the `Promise.prototype.then()`.

ben181231:
Updated

features/48-user-profile-discovery/field-based-acl.md
rickmak: Okay, I forget about the restriction.

So maybe, we should do this:

```
then((result) => {
  const { savedRecord, error } = result;
  console.log(`Successfully saved record: ${savedRecord}`);
  console.warn(`Rejected saving fields: ${error.$rejectedFields}`)
}, (err) => {
  console.error(`Failed to save: ${err}`);
});
```

So it will the same pattern for saving an array of records?
`const { savedRecords, errors } = result;`

If we put the `$rejectedFields` knowledge into Record because of js limitation, will are put it for all SDK, i.e. iOS and Android. I doubt is it the correct compromise.

features/48-user-profile-discovery/field-based-acl.md
```diff
@@ -197,6 +235,61 @@ The sample request is shown as followed:
...
+      // record that saved partially
+      "_access": [
+        { "level": "read", "public": true }
+      ],
+      "_created_at": "2017-06-19T06:06:30.125993Z",
+      "_created_by": "3c25ddff-b6d4-4d1d-8e67-45c3d82c90f5",
+      "_id": "note/2",
+      "_ownerID": "3c25ddff-b6d4-4d1d-8e67-45c3d82c90f5",
+      // list of fields rejected from update
+      "_rejected_fields": ["tags"],
```
rickmak: The naming `reject` does not tell any about access control?

We can keep `reject` and make it array of object. It is more extensiable.
`"_rejected_fields": [{"key": "tags", "reason":"access denied"}]`

Or we can make it more explicit by naming is `_access_denied_fields`, so we are more confident that it will only need an array of keys.

cheungpat: I think this design lacks extensibility.

From the perspective of the Field ACL, the `_rejected_fields` field satisfied the need to tell the client side which fields were ignored and failed to save. But this field will only work on Field ACL. It is likely that, down the road, we have other kinds of soft-error that we need to tell the user for.

Since the intention is to report soft-error to the client side, I think the new field should contains a list of error object indicating what kinds of error occurred. The error object is already deserializable by all SDKs.

Something akin to this:

```
"_warning": [{"code": 999, "message": "field permission denied", "info": {"fields": "tags"}}]
```

I am trying to make the new error [`DeniedArgument`](https://github.com/SkygearIO/skygear-server/pull/390/files#diff-e365aa9ef3a5e4c5505c294774a21ae0) accepted. I think the same kind of error is useful for indicating soft-error.

ben181231: Okay. Let me update `_rejected_fields` to `_warnings` and change the interface for extensibility.

PR #98, opened 2017-08-28
[WIP] Specification for basic SSO

carmenlau:
connects #5
Modify and update on #73 #63


Steven-Chan:
What about using js sdk with react-native?

PR #119, opened 2017-10-17
Update the field-base ACL matching spec

rickmak:
refs SkygearIO/skygear-server#479


cheungpat:
@rickmak I think I want to add some clarification to the spec PR and my proposed changes are here: https://github.com/rickmak/features/pull/1 does this make sense to you?

rickmak:
Yes, the update makes it even better.

cheungpat:
thanks @rickmak your initial contribution makes the update much easier to do

PR #121, opened 2017-10-19
Fix field ACL behavior without ACE defined

cheungpat:
Sorry I didn't catch this problem earlier. We discussed in meeting that we should match field ACL behavior with Record ACL behavior in that if the ACL definition is empty, we want to allow access. That's also what the developer would expect because when they have nothing configured in field ACL, they would expect their records to be accessible.


rickmak:
Merge via fce2b4c

PR #127, opened 2017-10-25
Add info on user roles and discovery levels

cheungpat:


carmenlau:
Thanks @cheungpat!

Picked via afadf35

Issue #64, opened 2017-05-16
Support offline message cache and sync for Chat


chpapa:
# Description

* Sync current users' participated conversation message history. When enabled:
  * it works with API like getMessages / getConversations to make sure function return cached messages/conversations
      * it provides callbacks to user' code for these events: `unsynced`, `syncing`, `sync completed`, `error` etc.
      * it listens to events and updates the Messages/Conversations local cache in real-time.
      * it auto-download and cache files in Messages if criteria met (criteria includes max size, MIME types, and if auto-download is enabled)
      * it auto-cache user records in message / conversations
  * it supports offline create conversations/send messages/change participants etc
      * it provides callbacks to user' code for send message success, failed, and a way to retry.
* It should also work with the iOS and Android UIKit

# API Design

# Related Issues

- Server Issues
- Client Issues
    - Fill in: JS SDK related issues
    - Other SDK remaining bugs / test
- Guides Issues

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

Steven-Chan:
# WIP

## SDK API interface

*The following code is pseudo-code only.*

### `Cache`

```
set(message: CachableMessage, forKey: string) -> boolean
get(messageForKey: string) -> CachableMessage
purge(key: string) -> boolean
purgeAll() -> boolean
```

(possible) implementation, each of them would provide individual set of configuration:
- `MemoryCache`
- `LocalStorageCache` (js)
- `RealmCache` (android/ios)
- `CoreDataCache` (ios)
- `SQLiteCache` (android)

### `CachableMessage<T>`

```
- text -> string
- data -> T
```

inherited class:
- `CachableTextMessage: CachableMessage<void>`
- `CachableAssetMessage: CachableMessage<Asset>`

chpapa:
@Steven-Chan @carmenlau 

[...]

PR #128, opened 2017-10-26
Draft: Add chat offline cache spec

Steven-Chan:
connect #64


Steven-Chan:
After an offline discussion with @chpapa and @carmenlau yesterday, I decided to re-write the spec again.

PR #130, opened 2017-10-27
WIP: Add chat app research and offline cache spec

Steven-Chan:
connect #64

### DIFF
--- a/features/64-chat-offline-cache/api.md
+++ b/features/64-chat-offline-cache/api.md
@@ -31,25 +31,25 @@ implemented by Skygear or the developers themselves, and
 
 ### Data Fetching
 
-1. Client App decides to fetch new data from remote server, so it calls API of Skygear Chat Container
-2. Skygear Chat Container calls API of Skygear Chat plugin to fetch data
-3. Skygear Chat Plugin gets results from DB or Skygear Server, and then it returns results
+- (1) Client App decides to fetch new data from remote server, so it calls API of Skygear Chat Container
+- (2) Skygear Chat Container calls API of Skygear Chat plugin to fetch data
+- (3) Skygear Chat Plugin gets results from DB or Skygear Server, and then it returns results
 to Skygear Chat Container
-4. Skygear Chat Container returns result to Client App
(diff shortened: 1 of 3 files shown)


Steven-Chan:
@carmenlau  Last time I mentioned about making message in db a link list data structure, i.e. each message has a reference of next and/or previous message.

I find the message has a sequence already, so we can make use of window function to generate a table to find neighbouring message.

I tried with the following sql.

```sql
select _id, conversation, seq, first_value(seq) over w as next_seq, _created_at, first_value(_created_at) over w as next_time
from message
window w as (partition by conversation order by seq desc rows between 1 preceding and current row);
```

chpapa:
Just want to reminded that for local store, besides finding message in sequence, it is possible for full text search. Might not have to consider it for now

Steven-Chan:
This would probably useful for identifying if there are missing messages between two given messages.

The last time discussion was about how to determine when to fetch messages in client, when there are some old local cache in the client and just received newly fetched result.

Given that the developer want to call `load more` when user scroll to top of the list, when to call `load more`?

```
----  <-- this may be the point to call load more.
 ^
 |
 |  10 messages in local cache
 |
 v
----

(How can the developer tell this gap between the two messages? Maybe caused by a period of offline time, or just occasional bad network env.)

----   <-- this may be the point to call load more.
 ^
 |
 | 10 newly fetched message
 |
 v
----
```

Steven-Chan:
Offline discussion @chpapa about how to keep and retrieve message history from db.

Current `message_history` table only keeps the changing history of a message.

We may make the `message_history` a complete journal, storing a `history sequence`, `message_id`, `action`(create/update/delete) , `data`(updated column and value). The concept is that this history table is able to generate a full message table.

So the client may keep the last fetched `history sequence` and then ask the server to fetch newer history and generate the latest snapshot of the message table.

We considered using postgresql materialized view, however seems postgresql does not provide incremental update, this may have performance issue. At this moment, we may just update the message table and history table within a transaction to keep the two table update to date.

Steven-Chan:
After some offline discussion, we decided to implement a simpler cache with sync version first.

PR #132, opened 2017-10-30
Add chat offline cache spec

Steven-Chan:
connect #64


features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,143 @@
...
+---
+ ^
+ | newly fetched result
+ v
+---
+```
+
+Solution:
+
+- Older cached result is removed, only the newly fetched result is kept
```
chpapa: Why we need to have case 2? Why we can't just cache those older message? Just curious

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Changes Required
+
+### Server
+
+- Add a per conversation sequence to `message` table, to determine distance between two given messages.
```
chpapa: Think about the implementation, I guess it is actually better to keep the auto increment ID global, and use a link list like structure (so each message row have a column called prev-message-id)

Advantage:
- server side implementation is easier. I had write a per column auto increment ID before, it is actually quite tricky to be fast, race condition free, and atomic.
- client side will be a bit more complex and slower, but should be minimal (checking if a number missing in a sequence vs checking if a prev-message-id is missing in sequence are also O(n))
- it is future proof, if one day we have a client got a lot of message and we need to have a distributed lock for ID, we don't need to rewrite this logic. Just make sure we use 64bit int as the column, and use a distributed mono increment sequence

chpapa:
I don't quite understand the sdk part of the latest change request.

One more thing, shall we take this chance to make a variance of getMessage, or even deprecate the old one? The old one use timestamp to get older message. I don't know who got the idea, but we really should change it to message ID, to avoid the possibility of missing a message in between with same timestamp.

chpapa:
Please also include all changes on the api?

chpapa:
I think for getMessages(conversation, limit, beforeMessageID, order), we actually need getMessages(conversation, limit, afterMessageID, order)

This is because:
1. When user disconnect pubsub, and reconnect, they need to get all messages after last disconnect and now. Currently no api to do that. (How we do that in demo app now?)
2. This lead to the question, shall we handle these for the users? At least in utility function?

Now it seems like offline cache is not much easier than sync DB lol

Steven-Chan:
@chpapa 

> Why we need to have case 2? Why we can't just cache those older message? Just curious

This part is about the handling of utility function we provide for the client app. The underlying store does not remove anything for this case.

Removing the older one in merged result can help client app to determine which message is the top of message list, they will know they need to fetch messages before the gap comes. And when they try to fetch the message in the gap, they will still get the older cached message first, but then when they get the fetched result and merge it to the existing result, they will get the complete message list.

I think this would make implementing `load more` function easier.

> we really should change it to message ID, to avoid the possibility of missing a message in between with same timestamp.

Agree.

> I think for getMessages(conversation, limit, beforeMessageID, order), we actually need getMessages(conversation, limit, afterMessageID, order)

I think it's a good idea, but for me the reason is quite different, I would update the example use case.

In short, I think the decision of choosing `beforeMessageID` or `afterMessageID` or `getting the latest N messages by not giving a message ID` depends on how the client handle the UI when it comes back online.

They can choose to make it `scroll to load more new message` or `directly skip to latest message`.
For the first behaviour, we need to provide an information that there are still more recent message, so the user can continuously scroll to get newer messages and know when to end.
For the second behaviour, there may or may not be a gap between new and cached messages. 
For both cases, we will provide utility functions for client app to handle.

Steven-Chan:
> Why we need to have case 2? Why we can't just cache those older message? Just curious

Maybe we can just provide a utility function for checking if two messages are subsequent messages, then the user can decide when they should fetch the messages to fill the gap.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+// app should load messages newer than messageA or messageG, and older than messageD or messageZ
+```
+
+This could be quite complicated, so we may provide scroll view delegate or handler implementation to handle this logic for the developer.
+
+# Changes Required
+
+### Server
+
+Add a `previous_message` to `message` table, to determine if two given messages are subsequent messages.
```
chpapa: Add a `previous_message` *column* to `message` table,

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+
+Deleted messages are also returned in fetch API, but the content of the message would be empty. This is to keep the whole message list complete.
```
chpapa: I suspect it is better to explicitly mark the message as deleted instead of just leave it as empty. For easy to understand and avoid confusion in future.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+
+#### JS
+
+```js
+const successCallback = (conversations, cached = false) => {
+  // display conversations
+};
+
+skygearChat
+  .fetchConversations(successCallback)
```
chpapa: I think in js, it is called `getConversations`.

And I think the parameter is a boolean instead of a function, according to the general API signature above?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+```js
+let currentMessages = [];
+
+const successCallback = (messages, cached = false) => {
+  currentMessages = skygearChat.merge(currentMessages, messages);
+  // display currentMessages
+};
+
+skygearChat
+  .fetchMessages(conversation, beforeMessage, limit, order, successCallback)
```
chpapa: again, I think the last parameter is boolean instead of function?

chpapa: And it is called getMessages in javascript.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+- Add utility functions for handling result returned by chat container
+
+```
+func merge(conversationDeltas: [ConversationDelta], toList: [Conversation]) -> [Conversation]
+func merge(messages: [Message], toList: [Message]) -> [Message]
+func merge(messageDeltas: [MessageDelta], toList: [Message]) -> [Message]
+
+func findBoundaries(ofMessageList: [Message]) -> [(Int, Int)]
```
chpapa: I'm not quite sure if these are really easy to use... maybe consider going "higher level" to provide the TableViewController / CollectionViewController is better... or we could leave this part empty, and try to do it when we apply the offline cache feature on the UIKit and refactor the functions as utility functions with a better use case?

chpapa: I suspect (wild guess!) the `ManagedMessageList` below would solve more common use case.... just guess again. So I think it is your call...

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+
+*All below UI update, I use `this.tableView.reloadTable()`. It is up to client app to decide how to update the UI*
+
+### In conversation list
+
+#### Start subscribing conversation changes
+
+When
+- enter conversation list
+- come offline and online again when in conversation list
```
chpapa: For Pubsub, my understanding is if you called the subscribe functions, it will handle auto-reconnect for you.

So for `subscribeConversations` and `subscribeMessages`, are they also going to auto-reconnect?

If not, it might be hard to make sure user understand it?

If yes, what is "come offline and online again when in conversation list"? Maybe you meant "init the conversation list"?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+
+```
+skygearChat.fetchConversations(completion: func (conversations, cached = false) {
+  this.conversations = conversations
+  this.tableView.reloadTable()
+})
+```
+
+### In conversation OR app come back online inside conversation
```
chpapa: Similar to the question above, what it meant by "come back online inside conversation"?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,373 @@
...
+
+### With ManagedMessageList
+
+#### Enter conversation
+
+```
+this.messageList = new ManagedMessageList(conversation: this.conversation,
+                                          delegate: this)
+
+if (skygearChat.pubsub.isConnected) {
```
chpapa: I think ManagedMessageList should handle it's relation with pubsub internally, instead of asking Developers to write these few lines of code all the time?

features/64-chat-offline-cache/api.md
chpapa: I think it is a bit controversial... maybe need a discussion to see if subscribeConversation / subscribeMessages should behave like subscribe in pubsub.

features/64-chat-offline-cache/api.md
Steven-Chan: This pattern is very like cached query in js, https://docs.skygear.io/guides/cloud-db/offline-storage/js/.

Because promise can only be resolved once, so the cached result is returned in the function passed as `getConversations` parameter.

features/64-chat-offline-cache/api.md
Steven-Chan: The `findBoundaries` is lower level function. I expect `ManagedMessageList` is enough for most common use case. But if `ManagedMessageList` does not suite someone's requirement, they can still use the `findBoundaries` instead of scanning the message list themselves, just more convenient and tell the developers the gap is supposed to be found and handled.

Steven-Chan: `ManagedMessageList` should be using `findBoundaries` internally.

And I am not sure if `findBoundaries` should be private or public yet.

features/64-chat-offline-cache/api.md
Steven-Chan: > So for subscribeConversations and subscribeMessages, are they also going to auto-reconnect?

I think it makes sense. So the two subscription functions should be use regardless of pubsub connection state.

features/64-chat-offline-cache/api.md
Steven-Chan: Then I think the `constructor` should also accept an instance of skygear container.

features/64-chat-offline-cache/api.md
Steven-Chan: Though, actually I think `getConversations` is not a good name though, because it is `fetching` resource from server.

chpapa: Oh right Promise can't be resolved twice.... hmmm, is there anyway to do it better? Or some reference to modern JS lib with similar problems? Felt like one callback is via Promise, another one is via function parameter, is quite weird...

chpapa: Like ... return two Promises? (I probably don't know what I'm saying lol)

features/64-chat-offline-cache/api.md
chpapa: I suggest keeping `findBoundaries` internally at this stage yet. We can always release it for the public in future if we find more use cases, but maintain another set not-quite-well-known public API which we might need to maintain backward compatibility in future is a trouble...

(Haha, shall we write somewhere to reminds ourselves, always release a minimal set of API that's enough for the use cases we designed?)

features/64-chat-offline-cache/api.md
Steven-Chan: I think `Rx` is common to solve this kind of problem

features/64-chat-offline-cache/api.md
chpapa: Sure.

howawong:
My two cents after the discussion:

1. Please keep `merge` and `applyDelta` in protected. Developers do not need to call the API explicitly. And of course, please list out which classes and interfaces are exposed to developers and which classes are for internal implementation only.

2. I think the cache policy is missed here. i.e. LRU, TTL etc.

3. And how media asset is being cached actually ? In the same cache or somehow ?

howawong:
Please also 
4. What does `MessageDelta` store?
5. Please clarify if boundary concept is needed as just discussed.

howawong:
6. Shall we have an simple ASCII flow chart explaining the calls between interface or classes ?
7. What if developers wants to fetch without cache?

Steven-Chan:
@howawong please see the updated spec.

> And how media asset is being cached actually ? In the same cache or somehow ?

I think this should be handled separately, because asset can be save and fetched simply by ID.

> What does MessageDelta store?

Basically an action type (create | update | delete), and the message itself. Each SDK should have this already, for pubsub.

> Please clarify if boundary concept is needed as just discussed.

Yes, I think this still exists, but it seems better just handled by the ManagedMessageList, I updated some description.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+
+#### Enter conversation
+
+```
+this.stickToBottom = true | false
+this.messageList = new ManagedMessageList(container: skygear.defaultContainer(),
+                                          conversation: this.conversation,
+                                          delegate: this)
+
+this.messageList.fetchLatestMessages()
```
chpapa: Why users need to call `fetchLatestMessage()`? There are no return value and parameter which makes me wonder why it was not in the constructor or...?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+    this.tableView.scrollToBottom()
+  }
+}
+```
+
+#### Render table
+
+```
+func numberOfCells() -> Int {
+  return this.messageList.messageCount()
```
chpapa: `messageCount()` will return count of all cached messages right? If yes maybe it should not be used this way right? shall we have a way for users to define they can always move the "cursor" of messageList with a specified length?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+### In conversation list
+
+#### Start subscribing conversation changes
+
+When
+- enter conversation list
+
+```
+skygearChat.subscribeConversations(completion: func (conversationDeltas) {
+  this.conversations = applyDelta(conversationDeltas, toList: this.conversations)
```
chpapa: What is `applyDelta`? Is it easier to just get all conversations managed like message list?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+
+When
+- enter conversation list
+
+```
+skygearChat.fetchConversations(completion: func (conversations, cached = false) {
+  this.conversations = conversations
+  this.tableView.reloadTable()
+})
+```
```
chpapa: Sample code missed how to send message in offline and handle the UI change?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+
+- Add cache store to skygear chat
+
+There should be a common cache logic layer and cache implementation.
+
+#### Cache implementation
+
+Skygear may provide cache implementation for each platform.
+
+- iOS: Core Data
```
chpapa: Is Core Data fast enough now? Because I heard from a layer user, he said Layer's team said their original implementation use Core Data and was very very slow.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,318 @@
...
+
+```
+func set(conversation: Conversation, forID: String) -> Void
+func set(conversation: Conversation, message: Message, forID: String) -> Void
+
+// if not implementated, the common cache logic needs to get all and filter the messages
+func fetchMessage(conversation: Conversation, before/afterMessageID: String, limit: Int, order: String) -> [Message]
+```
+
+#### Common cache logic
```
chpapa: I think ideally we can make the default cache have a simple cache policy of latest cache message first, and allow users to specific a size limit for messages and assets.

But I'm okay to ignore it in this version.

features/64-chat-offline-cache/api.md
Steven-Chan: No, I expect the `ManagedMessageList` cannot access to the cache store directly. It just calls chat api, and resolve the cached result and fetched result for the developer.

chpapa: So what will `this.messageList.messageCount()` return? It is unclear to me...

features/64-chat-offline-cache/api.md
Steven-Chan: I want to leave some flexibility to `ManagedMessageList`.

I expect the managed list contains the logic of how to resolve cached result and fetched result.

The scrollViewDelegate implementation is a bonus to the list to also contains the logic of when to fetch what.

So the reason why `fetchLatestMessage` does not return value is that the value are handled by the managed list. 

And then the list itself without attaching to scroll view delegate does not call any api, so user need to call it themself.

I expect a more complicated version of the `ManagedMessageList` in the future can also accept a list of messages in constructor, and developers can choose to fetch messages before/after those messages before even entering the message list.

This gives some flexibility to the developer to use the managed list to display messages within any time frame that they want.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,242 @@
...
+getResource(params, callback, cachedCallback?)
+```
+
+Resource includes conversation and message.
+
+The cachedCallback should be optional, thus compatible to the old API interace.
+
+### Expected UI Display / Common Use Case
+
+Client App is expected to keep a single list of messages for display in table. And they are required to keep track to the newest and oldest fetched messages to determine when to fetch result from server.
```
chpapa: Should not need to? I can image with the current API, to keep the message list up-to-date, client apps should only need to:

1. call subscribe() and update their message list in real time.
2. call getMessages() everytime when pubsub are connected. (Do we have need a API callback for that?)

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,242 @@
...
+
+Client App should replace old message with new message if the same message id is found.
+
+Deleted messages are also returned in fetch API, with empty content and marked as deleted. This is to keep the whole message list complete. Client App is expected to skip these messages in UI display.
```
chpapa: Client App display the empty message with visual indicator or not should be an application level decision.

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,242 @@
...
+
+### In conversation list
+
+#### Start subscribing conversation changes
+
+When
+- enter conversation list
+
+```
+skygearChat.subscribeConversations(completion: func (conversationDeltas) {
```
chpapa: subscribeConversaion is new right?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,242 @@
...
+The controller should have the following behaviour:
+
+- keep a cursor pointing to a message or the latest message, set a window size of how many messages should be kept in memory
+- when moving the cursor around (scrolling around or jumping around), it should erase messages outside the window and fetch messages newly entered the window
+- fetch the messages to fill the window when enter view, if the cursor is not set, it would stick to the latest message
+- subscribe to new messages, but only append to bottom if the window is on the latest message side
+- display, allow resend, cancel pending sent message, i.e. send by user but not responded by server yet
+- sync latest messages on reconnect if the window is on the latest message side
+
+# Sample Code for Conversation List
```
chpapa: Any sample code for message list?

features/64-chat-offline-cache/api.md
```diff
@@ -0,0 +1,242 @@
...
+```
+skygearChat.fetchConversations(completion: func (conversations, cached = false) {
+  this.conversations = conversations
+  this.tableView.reloadTable()
+})
+```
+
+# Changes Required
+
+Please also see [fetch_message_after_time.md](./fetch_message_after_time.md)
```
chpapa: afterMessage or afterTime?

features/64-chat-offline-cache/api.md
chpapa: And also sample code for createMessage

chpapa: Shall we also include how to handle when pubsub is connected, the users should get fetchMessage_after-time?

Steven-Chan:
@chpapa The pubsub event includes the record, so yes, we don't need to fetch again to cache the event record object.

features/64-chat-offline-cache/api.md
Steven-Chan: SDK now subscribes to everything. 

Each SDK has provided different amount of convenient methods to filter different resources, like message, conversation. I think we may need to unify the subscription interface across different SDK.

chpapa: @Steven-Chan Yea, please unify it in this chance if you can.

Issue #123, opened 2017-10-23
Push Notification Support for Chat


chpapa:
# Description

We should support most common Push Notification setting out-of-the-box.

Requirements:
* Have the following built-in push notification for chat:
  * Send push to User B when User A send message to User B, content is the message
  * Send push to all Users in a group when User A send to the group, content is the message
  * Send push to User B when User A start typing to User B (Snapchat Style)
* Work with UIKit, 
  * When push notification is pressed, it should jump to UIKit Conversation of the message
  * Support Quick Action like Reply in iOS and Android
* Able to customize or disable entirely.

Approach:
* Approach A: Like Parse or Layer, have some sort of Class in Client side for developers to configure the Push
* Approach B: What if we just have a template Cloud Functions repo, for users to clone (or support it from skycli)? The default code should do everything of the above, with in-line comments to tell users how they can customize it. It requires very good and readable function names too.
* Approach C: Extension of Approach B, but we have those hooks and template in Developer Portal. So we have an inline javascript editor in Portal, like "after message sent", with a template functions there to listen to the hooks.

My initial preference is Approach B, but I'm open to it. I'm also not sure how easy would be for Approach B to support UIKit, as the template would need to support both with/without UIKit.

If we go with Approach B, I also suggest besides the standard after_save hook, maybe we need to have new hook for pubsub for typing indicator, or even some specialized hooks for chat: https://github.com/SkygearIO/chat/issues/158

# Portal Design

**Remove this section if the feature have no Portal**

## Portal Features

**Put Portal Feature List here**

# API Design

**Remove this section if the feature have no API**

## Scenario

**Put list of Scenario how developers will use this API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues: https://github.com/SkygearIO/chat/issues/158
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
[...]

PR #136, opened 2017-11-08
Add Chat Push Notification Spec

howawong:
Connects #123
Connects SkygearIO/chat#158

### DIFF
--- a/features/123-chat-push-notification/api.md
+++ b/features/123-chat-push-notification/api.md
@@ -1,0 +1,159 @@
+# Push Notification in Chat
+## Feature Overview
+
+This feature enables developer to send customized push notification via cloud function deployment. The approach is to provide hooks in chat conversation including
+
+* Messages
+  * after\_message\_sent
+  * after\_message\_updated
+  * after\_message\_deleted
+* Typing
+  * typing\_started
+* Conversation
+  * after\_conversation\_created
+  * after\_conversation\_deleted
+  * after\_conversation\_updated
+  * after\_user_entered\_conversation
+  * after\_user_leave\_conversation
+
+In addition, the current JS push notification code snippet is too verbose and can be simplified.
+
+### Task
+- Refactor JS SDK Push Notification Function so that it can be called by developer with less code.  
+	- Original Code: 
+	
+	```javascript
+	var skygear = require('skygear');
+	skygear.push.sendToDevice(
+	  ['2aa4af2a-699a-4e43-8d67-7598757fc7ed'], // Device IDs
+	  {
+	    'apns': {
+	        'aps': {
+	            'alert': {
+	                'title': title,
+	                'body': message,
+	            }
+	        },
+	        'from': 'skygear',
+	        'operation': 'notification',
+	    },
+	    'gcm': {
+	         'notification': {
+	              'title': title,
+	              'body': message,
+	          }
+	    },
+		}
+	);
+	```
+
+	- new Code: 
+
+	```javascript
+	var skygear = require('skygear');
+	skygear.push.sendToDevice(
+	  ['2aa4af2a-699a-4e43-8d67-7598757fc7ed'],
+	  {
+         'title': title,
+         'body': message
+	  }
+	);
+	```
+- Create chat hook annotation in chat SDK
+- Enable chat plugin to call pre-defined hook
+
+### Sample Codes
+
+- Sample Javascript Cloud Function
+
+```javascript
+skygearChat.afterMessageSent((message, conversation) => {
+  var userIds = conversation.participants.map(p => p.user.key.recordID);
+  skygear.push.sendToUser(userIds, {'title': 'Title', 'body': 'Hello World'})
+})
+```
+
+- Python Cloud Function
+
+```python
+
+@chat.after_message_sent
+def after_messange_send_handler(message, conversation):
+   user_ids = [p.id.key for p in conversation.participants]
+   notification = {...}
+	skygear.action.push_users(container, user_ids, notification)
+```
+
+### List of APIs
+
+- Javascript
+	- `skygearChat.afterMessageSent((message, conversation) => {})`
+	- `skygearChat.afterMessageUpdated((message, conversation) => {})`
+	- `skygearChat.afterMessageDeleted((message, conversation) => {})`
+	- `skygearChat.typingStarted((conversation, events) => {})`
+	- `skygearChat.afterConversationCreated((conversation) => {})`
+	- `skygearChat.afterConversationUpdated((conversation) => {})`
+	- `skygearChat.afterConversationDeleted((conversation) => {})`
+	- `skygearChat.afterUserAddedToConversation((conversation) => {})`
+	- `skygearChat.afterUserRemovedFromConversation((conversation) => {})`
+
+- Conversation
+  - `@chat.after_message_sent`
+  - `@chat.after_message_updated`
+  - `@chat.after_message_deleted`
+  - `@chat.typing_started`
+  - `@chat.after_conversation_created`
+  - `@chat.after_conversation_updated`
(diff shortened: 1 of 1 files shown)


features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,159 @@
...
+	    },
+		}
+	);
+	```
+
+	- new Code: 
+
+	```javascript
+	var skygear = require('skygear');
+	skygear.push.sendToDevice(
```
chpapa: I guess `sendToUser` will be more often, can we take User Object as parameter too? @howawong

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,159 @@
...
+- Enable chat plugin to call pre-defined hook
+
+### Sample Codes
+
+- Sample Javascript Cloud Function
+
+```javascript
+skygearChat.afterMessageSent((message, conversation) => {
+  var userIds = conversation.participants.map(p => p.user.key.recordID);
+  skygear.push.sendToUser(userIds, {'title': 'Title', 'body': 'Hello World'})
```
chpapa: As mentioned above, better support `user` instead of `userIds`?

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,159 @@
...
+- Javascript
+	- `skygearChat.afterMessageSent((message, conversation) => {})`
+	- `skygearChat.afterMessageUpdated((message, conversation) => {})`
+	- `skygearChat.afterMessageDeleted((message, conversation) => {})`
+	- `skygearChat.typingStarted((conversation, events) => {})`
+	- `skygearChat.afterConversationCreated((conversation) => {})`
+	- `skygearChat.afterConversationUpdated((conversation) => {})`
+	- `skygearChat.afterConversationDeleted((conversation) => {})`
+	- `skygearChat.afterUserAddedToConversation((conversation) => {})`
+	- `skygearChat.afterUserRemovedFromConversation((conversation) => {})`
```
chpapa: for user added / removed, is it better to also have the user record object as the parameter?

howawong:
@rickmak  @carmenlau  please take a look and comment.

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,158 @@
...
+
+	- new Code:
+
+	```javascript
+	var skygear = require('skygear');
+	skygear.push.sendToUser(
+	  [userBen, userRick],
+	  {
+         'title': title,
+         'body': message
```
carmenlau: Besides title and body, need @howawong to help check if skygear server allow us to add custom data in the push notification. It is necessary to fulfill some use case like, passing conversation id so the app can open the conversation when user enter the app from notification. Or providing image url to have rich notification... etc

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,158 @@
...
+})
+```
+
+- Python Cloud Function
+
+```python
+
+@chat.after_message_sent
+def after_messange_send_handler(message, conversation):
+   user_ids = [p.id.key for p in conversation.participants]
```
carmenlau: Need to be aware `conversation.participants` is ready in the message hook.

howawong:
@chpapa  @carmenlau pointed one issue of push notification. Seems it cannot be simplified because developers want to send customized payload.

features/123-chat-push-notification/api.md
carmenlau: And also the example should filter out the message owener

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,158 @@
...
+- Create chat hook annotation in chat SDK
+- Enable chat plugin to call pre-defined hook
+
+### Sample Codes
+
+- Sample Javascript Cloud Function
+
+```javascript
+skygearChat.afterMessageSent((message, conversation) => {
+  skygear.push.sendToUser(conversation.participants, {'title': 'Title', 'body': 'Hello World'})
```
carmenlau: Better to enrich the js example like py version, instead of static title and body

chpapa:
> @chpapa @carmenlau pointed one issue of push notification. Seems it cannot be simplified because developers want to send customized payload.

@howawong @carmenlau oh yes forgot customised payload in push, but should still better than a undefined array as one parameter in current implementation?

carmenlau:
@chpapa @howawong I think we can defined a common payload, but this common payload should not only have `title` and `message`. And we also need to define how we map our defined payload to `apns` and `fcm` payload. Also in this common payload maybe we cannot avoid that some keys that for iOS only, like `category` or `badge`?

refs:
https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html
https://firebase.google.com/docs/cloud-messaging/xmpp-server-ref

chpapa:
> @chpapa @howawong I think we can defined a common payload, but this common payload should not only have title and message. And we also need to define how we map our defined payload to apns and fcm payload. Also in this common payload maybe we cannot avoid that some keys that for iOS only, like category or badge?
> 
> And I think the hooks part should be fine. Only we need to define more detail about the send notification part.
> 
> refs:
> https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html
> https://firebase.google.com/docs/cloud-messaging/xmpp-server-ref

Take reference to other APIs that send notification? It seems common to have a set of common attributes, and some specialised one for different platforms.

https://documentation.onesignal.com/reference#create-notification
https://docs.urbanairship.com/api/ua/#push-object

features/123-chat-push-notification/api.md
rickmak: And here the format should be the HCF of APNS and FCM. Not just title and body?

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,158 @@
...
+def after_messange_send_handler(message, conversation):
+   user_ids = [p.id.key for p in conversation.participants]
+   notification = {...}
+	skygear.action.push_users(container, user_ids, notification)
+```
+
+### List of APIs
+
+- Javascript
+	- `skygearChat.afterMessageSent((message, conversation) => {})`
```
rickmak: In 90% of the use-case, we are sending the push notification to receiving user. So probably `users` should appear in the function parameter?

features/123-chat-push-notification/api.md
```diff
@@ -0,0 +1,158 @@
...
+	- `skygearChat.afterMessageUpdated((message, conversation) => {})`
+	- `skygearChat.afterMessageDeleted((message, conversation) => {})`
+	- `skygearChat.typingStarted((conversation, events) => {})`
+	- `skygearChat.afterConversationCreated((conversation) => {})`
```
rickmak: Include participant?

features/123-chat-push-notification/api.md
howawong: Which user you are talking about?
message.owner or conversation.participants ? In either case, the user can be fetched from function argument.

features/123-chat-push-notification/api.md
howawong: conversation contains participants.

features/123-chat-push-notification/api.md
rickmak: According to your argument, we don't need `conversation` too. Since `conversation` can retrieve from `message.conversation`.

I think explicit function parameter will be helpful for user to do things without checking the full guide and documentation.

howawong: Agree. I will update the arguments.

rickmak:
@howawong where is the update?

Issue #56, opened 2017-04-19
Sign in with Custom Token


seventhmoon:
# Description
Similar to firebase.auth().signInWithCustomToken(token), which allow custom integration on other sign-in services.

# Portal Features
* Enable/Disable Custom Token function in Portal
* Generate / Download Key for client's App Server to generate JWT Token

# API Design
* Client SDK: Similar to firebase.auth().signInWithCustomToken(token)
* Consider comply with OpenID Connect JWT Spec?
  * Check `iss` and `sub` to match with the project's name
  * Check `aud` match with Skygear
  * Check `exp` is not expired
  * Use `uid` for user id?
  * Use custom claims to populate claims value in User Profile (?)

# Guides Required
* [ ] API Reference (iOS/JS/Android): for Client SDK
* [ ] Guide (iOS/JS/Android): Authenticate with Custom Authentication System
* [ ] Guide: Create Custom Token [Ref](https://firebase.google.com/docs/auth/admin/create-custom-tokens), provide recommend 3rd party library and sample codes for common languages.

# Scenario
Integrate with their apps' own authentication server.

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

chpapa:
@b123400 Can you also look at this feature issue and see if it is related with #5 which you're working on? cc: @rickmak

seventhmoon:
It is close, but seems not the same.
https://firebase.google.com/docs/auth/web/custom-auth
https://github.com/firebase/functions-samples/tree/master/linkedin-auth

Or #5 is mean using cloud functions to do this kind of thing?

[...]

PR #143, opened 2017-11-27
Login with custom token spec

carmenlau:
connect #56 
connect SkygearIO/skygear-server#493

### DIFF
--- a/features/56-login-with-custom-token/api.md
+++ b/features/56-login-with-custom-token/api.md
@@ -1,0 +1,160 @@
+## Feature Overview
+
+Allow user to integrate their authentication server to skygear server by logging in with custom token
+
+## Guides
+
+### Sample Codes (Client SDK)
+
+**JS**
+
+```js
+skygear.auth.loginWithCustomToken("<custom_token>")
+.then(function(skygearUser) {
+
+}).catch(function(error) {
+
+});
+```
+
+**iOS (swift)**
+
+```swift
+SKYContainer.default().auth.login(withCustomToken: "<custom_token>") { (user, error) in
+    if error != nil {
+        print("error login with custom token: \(error)")
+        return
+    }
+    print("sign up successful")
+    print("user record: \(user)")
+    print("email: \(user!["email"])")
+}
+```
+
+**Android (java)**
+
+```java
+skygear.getAuth().loginWithCustomToken("<custom_token>", new AuthResponseHandler() {
+    @Override
+    public void onAuthSuccess(Record user) {
+        Log.i("Skygear loginWithCustomToken", "onAuthSuccess: Got token: " + user.getAccessToken());
+    }
+
+    @Override
+    public void onAuthFail(Error error) {
+        Log.e("Skygear loginWithCustomToken", "onAuthFail: Reason: " + error.getMessage());
+    }
+});
+```
+
+### Sample Codes (User's server)
+
+1. Get the token secret from portal, and setup environment variable in your server `SKYGEAR_CUSTOM_TOKEN_SECRET`
+
+2. Generate create token
+
+**Python (With SDK)**
+
+```python
+from skygear.utils.auth import create_custom_token
+
+custom_token = create_custom_token("your server user id", profile)
+
+```
+
+**Rudy (Without SDK)**
+
+- Install [ruby-jwt](https://github.com/jwt/ruby-jwt)
+
+```ruby
+custom_token_secret = "Custom token secret from portal"
+user_id = "unique identifier of user in your server, string between 1-36 characters long"
+now_seconds = Time.now.to_i
+
+payload = {
+    :sub => user_id,
+    :iat => now_seconds,
+    :exp => now_seconds+(60*60), # expire within 1 hour
+    :skygear_profile => {}
+}
+
+JWT.encode payload, custom_token_secret, "HS256"
+```
+
+### List of APIs (Client)
+
+**py-skygear**
+
+```
+skygear.utils.auth.create_custom_token(principle_id, profile={})
+```
+
+**JS**
+
+```js
+skygear.auth.loginWithCustomToken(customToken: String): Promise<Record>
+```
+
+**iOS (swift)**
+
+```obj-c
+- (void)updateWithCustomToken:(NSString *)customToken
+            completionHandler:(SKYContainerUserOperationActionCompletion _Nullable)completionHandler;
+```
+
+**Android (java)**
+
+```java
+public void loginWithCustomToken(java.lang.String customToken,
+                                 AuthResponseHandler handler);
+```
+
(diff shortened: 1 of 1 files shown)


features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+- New handler `sso:custom_token:login` for sdk call
+    - Create skygear server token and login user
+    - Create new user when not found
+    - Will read the `skygear_profile` in jwt token, create profile for signup, update profile for login
+
+### Custom token claims (JWT)
+
+| Claims                                 | Description |
+|----------------------------------------|-------------|
+| sub (Subject)                          | Unique identifier of user in user's server, string between 1-36 characters long (Required) |
```
chpapa: Would we be more specific on what's UID? Is it the Skygear user ID? (But it is optional I suppose?)

Would we have another claims for the email address of the users?

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+    - Value will be used in user's server and skygear server, user's server create the token and skygear server verify it
+    - Need to be auto generated when create app in cloud
+- New handler `sso:custom_token:login` for sdk call
+    - Create skygear server token and login user
+    - Create new user when not found
+    - Will read the `skygear_profile` in jwt token, create profile for signup, update profile for login
+
+### Custom token claims (JWT)
+
+| Claims                                 | Description |
```
chpapa: Any security implication, if we don't specify `alg` and `aud`?

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
+## Feature Overview
+
+Allow user to integrate their authentication server to skygear server by logging in with custom token
+
+## Guides
+
```
rickmak: Please add the `Scenario`. i.e. Developer already have his own user system, but he want to get one of Skygear plugin, chat for example.

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+});
+```
+
+### Sample Codes (User's server)
+
+1. Get the token secret from portal, and setup environment variable in your server `SKYGEAR_CUSTOM_TOKEN_SECRET`
+
+2. Generate create token
+
```
rickmak: I would add a description on where to add this line. Normally during the user login. probably suggest the developer return the custom token to client-side SDK during signup.

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+**py-skygear**
+
+```
+skygear.utils.auth.create_custom_token(principle_id, profile={})
+```
+
+**JS**
+
+```js
+skygear.auth.loginWithCustomToken(customToken: String): Promise<Record>
```
rickmak: JS will have the `createCustomToken` for cloud function.

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+
+JWT.encode payload, custom_token_secret, "HS256"
+```
+
+### List of APIs (Client)
+
+**py-skygear**
+
+```
+skygear.utils.auth.create_custom_token(principle_id, profile={})
```
rickmak: Is it client API?

chpapa: It is server side API I think?

However, I'm not sure if it is a good idea to put it in py-skygear. Let's say I run my own python server, I think it would be confusing to import py-skygear and only use that function.

Maybe we don't need it at this point? 計 popularity, 其實有 PHP / Javascript / RoR create_jwt_token plugins 可能仲多人用過 python...

An alternative is we just provide sample code / sample project for each popular languages / frameworks.

features/56-login-with-custom-token/api.md
cheungpat: @chpapa The `alg` is part of the JWT authentication header which is inserted by all compatible JWT library. We do need to tell developer which algorithm to use though. Including `alg` here will confuse developers that they need to insert this field in the JWT claims.

Having `aud` is probably a good idea but this should be optional and the value should be user-configurable. Suppose the developer has two skygear apps and the authentication server generates different custom tokens for each app. Having a different `aud` disallow one custom token to be used in the other app.

chpapa: @carmenlau said we actually enforce HS256 for alg?

features/56-login-with-custom-token/api.md
```diff
@@ -0,0 +1,160 @@
...
+## Implementation Details
+
+### Changes on SDK
+
+- Create custom token function for user server side use in py-skygear and js sdk, see sample code
+- Login with custom token for client side use in js, ios and android core sdk
+
+### Changes on API at skygear-server
+
+- New environment variable `CUSTOM_TOKEN_SECRET`
```
cheungpat: `CUSTOM_TOKEN_SECRET` or `SKYGEAR_CUSTOM_TOKEN_SECRET` (either way?)

features/56-login-with-custom-token/api.md
cheungpat: `HS256` sounds good!

features/56-login-with-custom-token/api.md
carmenlau: This should be user's auth server user id. I think user will be more convenient that they don't have to store extra id for this feature and reuse their user id. So we have not restricted the id to be uuid or skygear user id .

features/56-login-with-custom-token/api.md
carmenlau: I will update this to provide example for generating the token from jwt libs

features/56-login-with-custom-token/api.md
carmenlau: This env var is in skygear server, so I think we should not need to prefix it with `SKYGEAR_`?

features/56-login-with-custom-token/api.md
carmenlau: For consistency, I think I will create example for how to create jwt token in js. I think we don't have much use case that user have to create custom token in cloud function. In case they need, they still can use the js example

features/56-login-with-custom-token/api.md
carmenlau: The `skygear_profile` claim is for populating developer system user info to skygear. So the email address should be there. I will update the example

carmenlau:
Updated the spec. Welcome for comments :)

Changes:
- Remove server sdk createCustomToken function
- Add js and python example for developer to generate JWT in their server
- Rename claim `skygear_profile` to `skyprofile`, for populating profile to skygear user profile

rickmak:
Let merge it if no further comments by today.

PR #144, opened 2017-11-29
Update Push Notification Feature Spec

howawong:
1. Hook Related Spec moved to hook.md
2. Update api.mid for cross platform push notification call

refs #123

### DIFF
--- a/features/123-chat-push-notification/api.md
+++ b/features/123-chat-push-notification/api.md
@@ -49,11 +49,8 @@ Main Payload is a collection of payloads.
 |body|string|message body|null|
 |title|string|message title|null|
 |subtitle|string|message subtitle|null|
-|sound|string|message sound|null|
-|title\_loc\_key|string|key to localized string|null|
-|title\_loc\_args|list of string|arguments of string replacement of title\_loc\_key|null|
 |priority|string|priority of message|"normal"|
-|expiry|int|time to live in seconds| 2419200 (4 weeks)|
+|ttl|int|time to live in seconds| 2419200 (4 weeks)|
 
 
 ### Custom Payload
@@ -65,13 +62,40 @@ Main Payload is a collection of payloads.
 |android\_body\_loc\_key\_args|string| |null|
 |android\_led\_color|string| |null|
 |android\_icon|string| |null|
+|android_sound|string|message sound|null|
 |ios\_badge|integer| |null|
 |ios\_content\_available|boolean| |null|
 |ios\_mutable\_content|boolean| |null|
 |ios\_launch\_image|string| |null|
 |ios\_category|string| |null|
+|ios_sound|string|message sound|null|
+|android_title\_loc\_key|string|key to localized string|null|
+|android_title\_loc\_args|list of string|arguments of string replacement of title\_loc\_key|null|
+|ios_title\_loc\_key|string|key to localized string|null|
+|ios_title\_loc\_args|list of string|arguments of string replacement of title\_loc\_key|null|
+|ios\_action\_loc\_key|string|key to action|null|
+|ios\_loc\_key|string|key to message string in Localizable.strings|null|
+|ios\_loc\_args|string|Variable string values to appear in place of the format specifiers in loc-key.|null|
 
+#### Example Push Notification Payload
 
+```json
+{
+  "message": {
+    "title": "Hello Rick",
+    "body": "Ditto",
+    "ttl": 99999
+  },
+  "custom": {
+    "android_sound": "wakeup.mp3",
+    "android_icon": "icon-1",
+    "ios_sound": "wakeup.m4a",
+    "ios_badge": 1,
+    "ios_launch_image": "launch.png",
+    "ios_category": "general"
+  }
+}
+```
 
 ## Implementation Details
 
@@ -100,6 +124,7 @@ function sendToUser(users, {
 
 2. SDK supports `users` in either list of user ids and list of user objects.
 3. SDK (in Android and iOS) should have payload builder class for auto complete.
+
 
 ### Changes on API at skygear-server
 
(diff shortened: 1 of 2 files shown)


features/123-chat-push-notification/api.md
```diff
@@ -1,158 +1,108 @@
...
+### Message Payload
+
+|key|type|description|default|
+|---|----|-----------|-------|
+|body|string|message body|null|
+|title|string|message title|null|
+|subtitle|string|message subtitle|null|
+|sound|string|message sound|null|
+|title\_loc\_key|string|key to localized string|null|
+|title\_loc\_args|list of string|arguments of string replacement of title\_loc\_key|null|
```
rickmak: How will the loc used? Will it be different from android/iOS?

features/123-chat-push-notification/api.md
```diff
@@ -1,158 +1,108 @@
...
+|priority|string|priority of message|"normal"|
+|expiry|int|time to live in seconds| 2419200 (4 weeks)|
```
rickmak: I would suggest to make it a determine datetime. It will make debugging easier and determine.

features/123-chat-push-notification/api.md
```diff
@@ -1,158 +1,108 @@
...
-### Changes on API at skygear-server
 
-Nil
+### Custom Payload
+|key|type|description|default|
+|---|----|-----------|-------|
+|android\_channel\_id|string| |null|
+|android\_click\_action|string| |null|
+|android\_body\_loc\_key\_action|string| |null|
+|android\_body\_loc\_key\_args|string| |null|
```
rickmak: I don't get why loc_key is android specific

features/123-chat-push-notification/api.md
```diff
@@ -1,158 +1,108 @@
...
+function sendToUser(users, {
+  'message': {
+    'title': 'Hello World',
+    'body': 'Hello from oursky.'
+  }
+});
+```
+
+2. SDK supports `users` in either list of user ids and list of user objects.
+3. SDK (in Android and iOS) should have payload builder class for auto complete.
```
rickmak: Actaully when we design the new push notification, we will want to include web platform. i.e. browser push notification.

features/123-chat-push-notification/api.md
```diff
@@ -1,158 +1,108 @@
...
+|ios|boolean|send to android device if true|true|
+
+### Message Payload
+
+|key|type|description|default|
+|---|----|-----------|-------|
+|body|string|message body|null|
+|title|string|message title|null|
+|subtitle|string|message subtitle|null|
+|sound|string|message sound|null|
```
rickmak: I think sound is ios/android specific?

rickmak:
Any update on this?

features/123-chat-push-notification/api.md
howawong: I think keeping it as seconds make more sense (and easier to trace) as in FCM & APNS it is also in number of seconds.
https://firebase.google.com/docs/cloud-messaging/concept-options

features/123-chat-push-notification/api.md
howawong: I think `body_loc_key` only exists in android?

features/123-chat-push-notification/api.md
howawong: I asked @carmenlau that we wanted to target at iOS and Android at this moment, am I right? And if we want to also support web push notification.

1. Do I need to scope out parameters for browsers?
2. Is it really supported by Skygear server?

features/123-chat-push-notification/api.md
rickmak: In the link you provide, there is ttl and expiry.

If the name is ttl, I am fine with the second.
if the name is expiry, better to make it Epoch Unix time stamp.

howawong: ok

features/123-chat-push-notification/api.md
rickmak: Yes. But i think why I don't get it is because `action-loc-key` is iOS specific. but we don't make it `ios-action-loc-key`.

So may be the real question is when to have device prefix, when not.

features/123-chat-push-notification/api.md
rickmak: Okay, agree we can focus on two main target platform for now.

features/123-chat-push-notification/api.md
howawong: hmmm, i did not include `action-loc-key`, did you mean `title_loc_key`? because I think it can be unified. If not, then I just split them into `android_title_loc_key` and `ios_title_loc_key`. The rule of thumb is, if a field can be unified / across platform, then a prefix skipped, else I would add a platform prefix.

howawong:
android picture = android_image
and
ios badge type = ios_category
ios badge count = ios_badge

rickmak:
After you update to add more prefixed attributes, I feel it is too much and difficult to digest as an end user.

In the link you provided: https://firebase.google.com/docs/cloud-messaging/concept-options
and in a service some of our client is using: https://docs.jiguang.cn/jpush/server/push/rest_api_v3_push/

Both are grouping the platform-specific settings to an other level, not at root level. I think this way is more readable and easy to code against on our side too.

Other references we can take are https://documentation.onesignal.com/reference#create-notification and https://docs.urbanairship.com/api/ua/#api-request-format


I agree with you on making the most common attributes appear in `message`.

For platform-specific, let try `platform` key? i.e.

```
{
    "message": {
        "title": "Greeting",
        "body": "Hi, Howa"
    },
    "platform": {
        "android": {
            "alert": "Greeting",
            "title": "Howa don't like to Android",
            "priority": -2,
        },
        "ios": {
            "alert": "Greeting",
            "sound": "default",
            "badge": "+1",
            "priority": 5,
        }
    },
}
```

I think it will make the API more easy to read/code against.

rickmak:
We need two core developer to approve the features. @carmenlau

howawong:
JSONs are fixed.
As discussed, if an user wants to support localization, then he just needs to put specific key-value pair in platform payload.

PR #149, opened 2018-01-09
Amend chat SDK with separate cache logic for message operation

cheungpat:


features/64-chat-offline-cache/api.md
```diff
@@ -174,39 +195,70 @@ fetchMessages(conversation_id, limit, before_message?, order?)
...
+```
+
+On JS, this function should return a Promise. On other platforms, a
+callback object / block is specified as the last parameter. The function
+should resolve/return a list of pending and failed operations.
+
+Operations can be removed from the cache store with this function:
+
+```
+func cancelMessageOperations(operationObject);
```
carmenlau: It should be `cancelMessageOperation`?

chpapa:
> As we finally used beforeDate for fetchMessages. I found you updated some of them already, but some of them are still using beforeMessage. Would like to have your help to update using beforeDate!

`beforeDate`? Sorry but who comes up it? I think we should use `beforeMessage(ID)`, if not at least `beforeTime` as the current APIs are... what's the matter with `beforeDate`? @carmenlau

cheungpat:
@carmenlau

carmenlau:
Looks good, thanks @cheungpat!

carmenlau:
Ohoh, @cheungpat Found that the SDK sections some of the api is still using `beforeDate`. And the JS example `fetchMessages` cache and no-cache version are not correct too. Could you help check? thanks!! 🙏🏻

https://github.com/cheungpat/skygear-features/blob/7482778d1e7b463baf5af494088c7934bee2ab0a/features/64-chat-offline-cache/api.md#sdk

carmenlau:
And also the api name now become `fetchOutstandingMessageOperations` :) @cheungpat

cheungpat:
updated @carmenlau

Issue #50, opened 2017-04-07
Disable users


chpapa:
# Description

There are cases to temporary disable a user. For example, violation of the term of use; awaiting settle of a payment, etc.

Currently, a developer can do two things to achieve the goal.
1. Change the password of the user. (Usually add a `!` in front of the password has to make it invalid)
2. Mark the user as disabled at `user` record.

We want to provide a built-in way to do it.

# CMS Design

## CMS Features

- A button to disable a user temporarily, with these parameters:
  - disable until Date Time (Time Zone)
  - A message to display when the user tries to login again
- A button to enable the user back
- A button to "Require user to change password in next login"

## UX Design
**Put (link to) Portal Design here**

# API Design

### APIs

These three API requires a special "user_management" role to call:
* `skygear.auth.disableUser(user, [message="Account blocked"], [expire_at=never])`
* `skygear.auth.enableUser(user)`
* `skygear.auth.changePasswordatLogin(user)`

After `disableUser` or `changePasswordatLogin` was called, the user's subsequent API call will return `NotAuthenticated`.

### Server Design

* skygear-server provider after user login hooks, and plugins listen to the hooks to block user login
* have a simple boolean variable at _user or UserProfile(which is better?) to define blocking status

## Scenario

**Put list of Scenario how developers will use this API**

## Sample Code

**Put sample code of how you vision this API will be used, consider different type of developers and different abstract level**

# Open Questions

* Should we return NotAuthenticated for blocked users, or better another error code that's for the exact purpose?

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
[...]

PR #158, opened 2018-01-19
Add disable user spec

cheungpat:
connects #50


cheungpat:
need to add audit event

features/50-disable-user/api.md
```diff
@@ -0,0 +1,75 @@
...
+## Sample usage:
+
+To disable user account:
+
+```javascript
+// to disable a user account
+var user = 'EA04BEBD-1BFD-4FDD-A7D8-615F007A3B32';
+var reason = 'Account disabled because of TOS violation.';
+var now = new Date();
+var expiry = new Date(now.getTime() + 60*5); // 4 mintues
```
cheungpat: 5 minutes

Issue #22, opened 2017-03-28
Support email verification after sign up


chpapa:
# Description

Handle the use case of User Profile have an email field, and users need to verify by emails before using certain functions.

# Portal Design

## Portal Features

* Portal Setting (Under `User Auth` > `Email Verification`)
  * Setting (Enable, Sender, From, Reply To)
  * Verify Email Template (sent when email verification requested)
  * Verify Confirmation HTML Template (display the HTML when the confirmation link in **Verify Email** is clicked)
  * Email Address Change Notification Email Template (optional, sent to old email address when email change)
  * Each Email Template / HTML Template in Portal have a "Try" button for a preview.
  * Each email template should have the following fields:
    * From Email Address
    * Subject
    * Reply to
    * Content
    * Redirect URL (only applicable for emails)
    * URL Lifetime (only applicable for HTML templates)
  * Content are expected to be an HTML / plain text editor with placeholder variable. The list of placeholder variable should include all the fields under the User table, and something like `[[url]]` to include the link to confirmation result page.

## UX Design
**Put (link to) Portal Design here**

# API Design

## Scenario

* Simple blogging apps that require email verification to post new blog post
  * Send a verification email when non-verified users logged in and tried to post a new blog post
  * Display a page say "You need your email to be verified before a new post" if user's email was not verified.
  * After the user clicks on the link of verified email, it should show a page say "Verified email", and users can click elsewhere to continue what they do.

## Sample Code

When user click on an action that require email verification:
```
if !currentUser.email_verified
  skygear.userUtils.sendEmailVerfication('actionX')
```

The above code will send an email, and the verification links after clicked, will comes with actionX as the parameter. `actionX` will pass as a string as query string to the redirected URL after the verification link is clicked, or ignored if users supply an HTML template.

## API Design

**Put Design and Specification here with links**

* Call `sendEmailVerification(currentLink)` from client side
* `User.email_verified` return `true` or `false`

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
[...]

Issue #20, opened 2017-03-28
Support SMS verification


chpapa:
# Description

Handle the use case of User Profile have an phone field, and users need to verify by SMS before using certain functions.

Related with #19 Login by SMS?

# Design/Specification
* Support Twillio as the only SMS gateway is acceptable.
* Workflow:
  * Call `sendSMSVerification()` from client side (support any fields or a default field for phone no?)
  * Call `verifySMSVerification(code)` from client side, return `success` or `failure`
  * `User.verifiedBySMS()` return `true` or `false`

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

chpapa:
Code completed, guides and portal will be done in #200

Issue #19, opened 2017-03-28
Login by SMS


chpapa:
# Description

Login by SMS, kind of like Telegram / Village. Where users have no username / email (both are optional), users can sign up or login simply by a verification on SMS.

# Design/Specification

WIP

* Support Twilio as the only SMS provider is acceptable.

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

tensiuyan:
@chpapa is this like one time login?

chpapa:
@tensiuyan what do you meant by one time login?

seventhmoon:
I think @tensiuyan means, is the SMS going to send a one time password for log-in (like https://1cm.hk.chinamobile.com/site/login.html?lang=en), or is the SMS is to verify the user owns the number for registration use (and have to provide a password for registration).

chpapa:
Oh it is like one time login. For SMS verification there is another issue #20

rickmak:
@limouren and @josephchow102 are working on an implementation.

rickmak:
I updated a bit on the API design and found it will more appropriate to design the feature to be `passwordless`. We may take reference to Auth0: https://auth0.com/docs/connections/passwordless

PR #159, opened 2018-01-19
Add email and SMS verification

cheungpat:
connects #20
connects #22

### DIFF
--- a/features/22-email-verification/api.md
+++ b/features/22-email-verification/api.md
@@ -99,13 +99,13 @@ for verification again.
 
 ```javascript
 // email
-skygear.auth.sendEmailVerification('johndoe@example.com')
+skygear.auth.sendEmailVerification()
 .then(() => {
   // email sent
 });
 
 // SMS
-skygear.auth.sendPhoneVerification('+85221559299')
+skygear.auth.sendPhoneVerification()
 .then(() => {
   // SMS sent
 });
@@ -115,7 +115,7 @@ To check for verification:
 
 ```javascript
 // verify email 
-skygear.auth.verifyEmail('johndoe@example.com', '31DF2310FAA1')
+skygear.auth.verifyEmail('31DF2310FAA1')
 .then(() => {
   console.log(skygear.auth.verified); // true
   console.log(skygear.auth.verifiedRecordKeys.keys()); // ['email']
@@ -136,28 +136,3 @@ skygear.someAction()
 ## Portal UI
 
 (to be expanded)
-
-## Future extension
-
-The following section describe how the verify by Email/SMS API will be
-extended to work with other use cases.
-
-These use cases will be covered by future specifications.
-
-### Login with Email/SMS
-
-The flow for logging in with Email/SMS will be accomplished by
-first requesting a verification, and then use the code to login.
-
-The app should call `sendEmailVerification()` (or equivalent API for SMS),
-and then use the obtained code to call `loginWithEmailAndVerifyCode()`, the
-latter of which is a wrapper around `verifyEmail()` above. The SDK should call
-the same or similar underlying server actions.
-
-### Reset Password
-
-The flow for reset password is accomplished by first requestign a verification
-and then using the code to reset the password.
-
-The app should call `sendEmailVerification()` (or equivalent API for SMS),
-and then use the obtained code to call `resetPassword()`.
--- a/features/22-email-verification/api.md
+++ b/features/22-email-verification/api.md
@@ -140,6 +140,10 @@ skygear.someAction()
 });
 ```
 
+## Portal UI
+
+(to be expanded)
+
 ## Future extension
 
 The following section describe how the verify by Email/SMS API will be
--- a/features/22-email-verification/api.md
+++ b/features/22-email-verification/api.md
@@ -11,6 +11,7 @@ The use case and requirement apply to verify-by-email and verify-by-sms.
 * The developer may want to deny unverified user from accessing any server API except auth related ones.
 
 ## Requirements
+
 * Server will have configuration options for:
 	* enable/disable email verification (boolean)
 	* enable/disable SMS verification (boolean)
@@ -22,14 +23,39 @@ The use case and requirement apply to verify-by-email and verify-by-sms.
 * Server will add API for sending/resending verification email
 	* when called by an admin or with master key, the API can send verification email to any unverified user, this is designed for cloud code use
 	* when called by non-admin unverified user, this API resends the verification email, this is designed for app use
-* The email is sent via SMTP protocol that is configurable. The SMS is sent via a SMS gateway that is configurable.
-* Server  will add API for checking email verification code
+* Server will add API for checking email verification code
 * Server will add a request preprocessor to check if the user is verified and return an error (if verification is required).
 * Signup/login API to return verification status of a user
 * Signup API to send verification email if verification is enabled.
 * JS, iOS and Android SDK to add API for resending verification email and checking verification code.
 
+### Criteria for Verifying a User
+
+The developer is able to specify which information is required to be verified
+in order for a user to become verified. For example, an app that requests for
+email and phone may require either or both email and phone is verified before
+the user become verified.
+
+### Verification Code Persistence and Consumption
+
+For each verification email/SMS, the code should be randomly generated and
+the code will persist into the database. Each code has an associated expiry time
+and can only be used once.
+
+### Modify Authentication Keys
+
+When modifying authentication keys (auth record keys, i.e. email or phone) using
+the Record DB API, the user may become unverified. API will be provided so that
+the app can decide whether to update the email or phone upon verification.
+
+### SMS and Email Provider
+
+SMS and email are sent via a provider, which is extensible in server code only.
+Email provider will support SMTP initially and SMS will support Twilio and
+Nexmo. Provider is configurable.
+
 ## Sample usage
+
 Flow:
 
 1. The user install an app.
@@ -47,14 +73,14 @@ Flow:
 When signing up:
 
 ```javascript
+// email
 skygear.auth.signupWithEmail("johndoe@example.com", "password")
 .then((user) => {
   console.log(skygear.auth.verified); // false
 });
-```
 
-```javascript
-skygear.auth.signupWithPhone("johndoe@example.com", "password")
+// phone
+skygear.auth.signupWithPhone("+85221559299", "password")
 .then((user) => {
   console.log(skygear.auth.verified); // false
 });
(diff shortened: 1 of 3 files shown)


features/22-email-verification/api.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Requirements
+* Server will have configuration options for:
+	* enable/disable email verification (boolean)
+	* enable/disable SMS verification (boolean)
+	* set verification as required/optional (boolean)
+	* smtp server settings
+	* SMS gateway settings
+	* email template
+	* SMS template
```
chpapa: I guess we will need a web template (kind of like forget-password) after user click the verification links? It could be a web template or a redirection URL (if anything could be done to make deep-linking in app easier, should be included too)

features/22-email-verification/api.md
```diff
@@ -0,0 +1,111 @@
...
+	* For verification code, the user will type the verification code from the email address/SMS an enter it to the app to complete verification.
+	* For verification link, the user will be redirected to a webpage, which automatically send the code to the server to complete verification.
+* The App is able to check the verification status of the user using the SDK and cloud function so that it can display other info and/or disable features.
+* The developer may want to deny unverified user from accessing any server API except auth related ones.
+
+## Requirements
+* Server will have configuration options for:
+	* enable/disable email verification (boolean)
+	* enable/disable SMS verification (boolean)
+	* set verification as required/optional (boolean)
```
chpapa: What will happen on API level if the verification is required / not required?

features/22-email-verification/api.md
```diff
@@ -0,0 +1,111 @@
...
+
+```javascript
+skygear.auth.signupWithEmail("johndoe@example.com", "password")
+.then((user) => {
+  console.log(user.verified); // false
+});
+```
+
+```javascript
+skygear.auth.signupWithPhone("johndoe@example.com", "password")
```
chpapa: Will it change the UNIQUE user profile data environment variable / need related settings?

features/22-email-verification/api.md
```diff
@@ -0,0 +1,111 @@
...
+* App may request the user to verify the email address/phone number upon signup.
+* The app need will send an email/SMS to the user with the email address/phone number provided when logging in. The app will want to customize the content of the email/SMS.
+* Both email/SMS verification supports verify by a verification code or by a verification link.
+	* For verification code, the user will type the verification code from the email address/SMS an enter it to the app to complete verification.
+	* For verification link, the user will be redirected to a webpage, which automatically send the code to the server to complete verification.
+* The App is able to check the verification status of the user using the SDK and cloud function so that it can display other info and/or disable features.
+* The developer may want to deny unverified user from accessing any server API except auth related ones.
+
+## Requirements
+* Server will have configuration options for:
```
chpapa: Another configuration I can think of is the complexity of verification code. We can start with a simple one, or a setting to allow a simple and a complicated one.

Reason is most of the time, for apps, it is pretty simple, could be 4 - 6 digits, for ease of input from SMS to apps.

For web, usually is a bit more complicated.

cheungpat: Yes! I also think we should have different style of verification code something like `123456` for SMS and `352119F5C0B0` for email. Configurable of course.

features/22-email-verification/api.md
cheungpat: I am thinking that if verification is required, then all the API (except for the auth related) will deny the request. We still want auth related API to go through because we still need the user to login so that they can ask the server to resend the verification email/SMS. (Otherwise, a user will be able to request a verification email for another user, which sounds bad : )

features/22-email-verification/api.md
cheungpat: I aim not to change the environment variables currently associated with login.

features/22-email-verification/api.md
chpapa: Thanks understood. I guess a bit of explain in the spec will helps since it is not very trivial.

features/22-email-verification/api.md
```diff
@@ -0,0 +1,112 @@
...
+skygear.auth.signupWithEmail("johndoe@example.com", "password")
+.then((user) => {
+  console.log(skygear.auth.verified); // false
+});
+```
+
+```javascript
+skygear.auth.signupWithPhone("johndoe@example.com", "password")
+.then((user) => {
+  console.log(skygear.auth.verified); // false
```
chpapa: Please add how users can check SMS / Email verification status specifically.

cheungpat:
* whether the request for verification link will overwrite one another?
* may create a table for verification code

portal?

* need UI for configuring verification email?
* need UI for sign up with email or phone or username?

spec need to mention

* may need to verify both *email* and *phone*
* twilio and nexmo

cheungpat:
oh, and you maybe interested in discussing whether this feature is to be implemented in the core or the plugin : )

features/22-email-verification/implementation.md
```diff
@@ -0,0 +1,216 @@
...
+```
+
+
+## Plugin
+
+Handler and Lambda will be allowed to register a handler or a lambda
+that requires user to be verified.
+
+```python
+@skygear.handler(name, verify_required=True)
```
carmenlau: For the current `user_reqired`, I think it should also be rejected, if `VERIFY_REQUIRED` and the user is not verified?

features/22-email-verification/implementation.md
```diff
@@ -0,0 +1,216 @@
...
+* `verified` (boolean)
+  If the user is verified with one of the keys, this returns true.
+
+* `verified_record_keys` (object, string: any)
+  Returns which piece of auth record key is verified.
+
+APIs that include authResponse are (but not limited to):
+
+* `auth:login`
+* `auth:signup`
```
carmenlau: Also the sso api
`sso:oauth:login`
`sso:oauth:signup`
`sso:custom_token:login`

features/22-email-verification/implementation.md
```diff
@@ -0,0 +1,216 @@
...
+* `verify_key` (string)
+  The auth key, either `email` or `phone`.
+
+* `user_id` (string, optional)
+  The user to verify. If the request is authenticated with access token
+  only the current user is accepted. If request is authenticated with master key
+  any user is accepted. Default to current user. If the request is not
+  authenticated, the user will be looked up from `verify_data`.
+
+* `verify_data` (string, optional)
```
carmenlau: Would like to know what will happen if `verify_data` is different from the current user email or phone, those will be udpated?

features/22-email-verification/api.md
```diff
@@ -47,14 +73,14 @@ Flow:
...
 skygear.auth.signupWithEmail("johndoe@example.com", "password")
 .then((user) => {
   console.log(skygear.auth.verified); // false
 });
-```
 
-```javascript
-skygear.auth.signupWithPhone("johndoe@example.com", "password")
+// phone
+skygear.auth.signupWithPhone("+85221559299", "password")
```
david90: is password optional when signing up with phone Number?

features/22-email-verification/implementation.md
cheungpat: the check will not be successful

features/22-email-verification/implementation.md
cheungpat: thanks

features/22-email-verification/api.md
cheungpat: this will be covered later with #19

cheungpat: Common use case mentioned by PM says that it is common for apps that login with phone number to have no password for logging in. Therefore, this API for signing up with phone number *and* password is going to be less commonly used. But the spec currently is grown to the point it is not likely we will finish this feature in the upcoming milestone if we add login with email/SMS into this spec.

I am happy to add this spec for login with email/SMS if we found time to implement this towards the end of the upcoming milestone.

rickmak:
1. `VERIFY_CRITERIA` should not required if `VERIFY_KEYS` are using the AUTH_REACORD_KEY format.
1. SSO may need modification according to this spec.

features/22-email-verification/implementation.md
howawong: and can I specify type of verification in server request? Or there is only one type of verification in one skygear server?

howawong: and do we have role ACL for verified user and unverified user?

howawong: and in (python cloud function )[https://docs.skygear.io/guides/cloud-function/lambda/python/],
there is `@skygear.op(name, user_required=False, key_required=False)`
do we need to add `user_verified=False/True`?

features/22-email-verification/api.md
```diff
@@ -1,73 +1,133 @@
...
 
-### Modify Authentication Keys
+### Extendsible Verification Provider
 
-When modifying authentication keys (auth record keys, i.e. email or phone) using
-the Record DB API, the user may become unverified. 
+A few provider will be provided to verify user data. Provider is configurable
+via environment variable. The provider should be extensible so that developer
+having custom user data can provide their own provider. The provider is
+extensible in python cloud code only.
```
chpapa: I think it is okay to support Python only for now. But just open an issue in JS Cloud Code to indicate we might want to support it in future?

features/22-email-verification/api.md
```diff
@@ -1,73 +1,133 @@
...
 ### Criteria for Verifying a User
 
-The developer is able to specify which information is required to be verified
+The developer is able to specify which user data is required to be verified
 in order for a user to become verified. For example, an app that requests for
 email and phone may require either or both email and phone is verified before
 the user become verified.
 
-### Verification Code Persistence and Consumption
+### Configuration
```
chpapa: I think similar to forgotPassword, we will need to have configurations for the HTML or URL of the verification success/failure page?

features/22-email-verification/api.md
```diff
@@ -115,10 +172,10 @@ To check for verification:
 
 ```javascript
 // verify email 
-skygear.auth.verifyEmail('31DF2310FAA1')
-.then(() => {
+skygear.auth.verifyWithCode('email', '31DF2310FAA1')
```
chpapa: do we really need the field parameter (first parameter) in `verifyWithCode`? (A question, no idea)

features/22-email-verification/api.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Configuration
+
+The following is configurable:
+
+* which user data field supports verification
+* for each user data field:
+  * provider
+  * settings for the provider
+    * for email: SMTP server settings, username and password
```
chpapa: I think similar to forgotPassword, we will need to have configurations for the HTML or URL of the verification success/failure page?

features/22-email-verification/api.md
```diff
@@ -0,0 +1,195 @@
...
+* whether non-auth API is blocked for unverified user
+* whether verification link/code is sent automatically upon signup
+* whether verification link/code is sent automatically upon updates
+
+### Extendsible Verification Provider
+
+A few provider will be provided to verify user data. Provider is configurable
+via environment variable. The provider should be extensible so that developer
+having custom user data can provide their own provider. The provider is
+extensible in python cloud code only.
```
chpapa: I think it is okay to support Python only for now. But just open an issue in JS Cloud Code to indicate we might want to support it in future?

features/22-email-verification/api.md
```diff
@@ -0,0 +1,195 @@
...
+.then(() => {
+  // SMS sent
+});
+```
+
+To check for verification:
+
+```javascript
+// verify email 
+skygear.auth.verifyWithCode('email', '31DF2310FAA1')
```
chpapa: do we really need the field parameter (first parameter) in verifyWithCode? (A question, no idea)

features/22-email-verification/api.md
```diff
@@ -0,0 +1,195 @@
...
+
+```javascript
+skygear.someAction()
+.then(() => {
+  // the promise is rejected and this callback is not called
+}, (err) => {
+  console.log(err.name); // 'VerificationRequired'
+});
+```
+
```
chpapa: I think we also should include (could be done separately later) the spec for cloud functions:

1. Support verified_required as a parameter for lambda functions, so users can limit lambda functions to verified users easily.

2. Support `after_user_verified` as a call back for cloud functions when user is verified.

chpapa: It doesn't meant we have to implement than rightaway, but I guess that would be more complete for the verification feature for Skygear

cheungpat:
@carmenlau is there anything that need changing for this PR?

features/22-email-verification/implementation.md
carmenlau: @howawong 
> non admin user calls verify_user and unverify_user
The response will be the same as the other apis that require master key and admin, it should be deny

> an I specify type of verification in server request
The type of verification is defined by environment variable, one skygear server can have multiple verification method (email, phone...etc). But one user field can only define one verification method. You can see the detail in configuration section.

> do we have role ACL for verified user and unverified user
We don't have this plan yet, but we can plan for this if needed in future

> do we need to add user_verified=False/True
To avoid confusion, we will not add this now. But we will have verified flag in the context, so developer can achieve it by checking the context

carmenlau:
@cheungpat 
1. Need update the spec for handling verified by link that we will support success and failure templates and redirect like forgot password ?
https://github.com/SkygearIO/forgot_password/blob/master/README.md

2. Per discussion, we will only support using py for writing provider, so no update is needed for this part.

3. See if we want to remove field name in `verifyWithCode`?

4. Confirmed that to avoid confusion, we will not add `verified_required` at the moment. Developer can still check the context for the verified flag. So no update is needed too

5. I will add another issues for the planning the new server hook and put `after_user_verified` there, as we will need more planning for that

cheungpat:
updated @carmenlau

carmenlau:
Thanks @cheungpat !!

Issue #150, opened 2018-01-15
Enable security audit feature in skygear cloud


carmenlau:
# Description

- Support showing server audit log in portal
- Support password policy configuration

# Portal Design

- Audit log page
- Password policy configuration page

# Portal features

- Show server audit log
- Setting policy configuration updating server env

## Infrastructure

- To discuss how to support app's logging in skygear cloud

PR #161, opened 2018-01-22
Security audit logging

cheungpat:
connects #150

### DIFF
--- a/features/150-audit-log/api.md
+++ b/features/150-audit-log/api.md
@@ -1,0 +1,107 @@
+# Security audit logging on Skygear Cloud
+
+## Use cases
+
+* For administrators of an app to download audit log for further investigation.
+
+## Requirements
+
+* Portal to build UI for configuring audit logging
+* Portal to build UI for retrieving log files
+	* Only showing download links for retrieval. No need to provide searchable UI.
+* Controller
+	* Will discover log files and generates link for client retrieval
+* Log collection daemon
+	* The log collection daemon is going to collect the log generated from Skygear Server.
+	* The daemon will upload the log entries to a persistent storage for long term storage and for retrieval
+
+## Portal UI
+
+### Enable Audit Log
+
+The portal will have a new section for audit log. The user enables the audit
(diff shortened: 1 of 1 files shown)


features/150-audit-log/api.md
```diff
@@ -0,0 +1,31 @@
+# Security audit loggin on Skygear Cloud
```
cheungpat: s/loggin/login/

cheungpat:
* how long the s3 retain log
* how often the s3 will update the log

chpapa:
> how long the s3 retain log

I guess this one need to be configurable.

cheungpat:
> I guess this one need to be configurable.

yea, i guess this should be configurable, but since the log is not retained by the skygear server, the configuration will apply to the cloud daemons instead. To make things easier can we say the cluster will have the same audit log retention period that is long enough for most apps?

chpapa:
> To make things easier can we say the cluster will have the same audit log retention period that is long enough for most apps?

My gut feeling is yes.

features/150-audit-log/api.md
```diff
@@ -0,0 +1,107 @@
...
+
+The portal is expected to call this API to obtain a list of audit-logs for
+display.
+
+The controller is expected to check that the user is an owner or collaborator
+of an app. Then the controller lists all objects with a certain prefix
+from the S3 bucket that is storing the audit logs.
+
+* S3 bucket: TBD
+* Prefix: `audit-logs/<app>/<date>/`
```
rickmak: Will the format change to this is better for query `<month>/<week>/<day>`

features/150-audit-log/api.md
```diff
@@ -0,0 +1,107 @@
...
+log events might not be immediately available.
+
+## Controller API
+
+The Controller API does not need to update to enable audit logging. Portal
+will add environment variable to config.
+
+The Controller will add the following APIs:
+
+### `GET /apps/<id>/audit-logs/<date>`
```
rickmak: Skycli can provide command to download a date range.

Issue #152, opened 2018-01-16
[CMS] Role models and policy setting spec


carmenlau:
# Description

**Add 2-3 sentence description of your feature**

# Portal Design

**Remove this section if the feature have no Portal**

## Portal Features

**Put Portal Feature List here**

# API Design

**Remove this section if the feature have no API**

## Scenario

**Put list of Scenario how developers will use this API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

chpapa:
I'm imaging applying a policy-based ACL for the CMS, it could be something like:

## Should support in this phase

### Record based access
action: CREATE / READ / UPDATE / DELETE
resources: `resources:entity:<entity name>`

(reason why it is `entity`, not `record type`, is because the ACL should reflect the entity defined in the CMS's config file, not each record class in Skygear DB)

e.g.
* Allow listing of records, and create new records
action: `create|read`
resources: `resources:entity:jobs`

### Permission to access certain functions
action: (CRUD, or sth else is better?)
resources: `resources:push`

Set a user to disallow access to `resources:push` will disable the push section

## Support in future?

* `resources:entity:<entity name>:<ids>:<field name>`
* `resources:push:<user ids>`
    * actions:
        * SEND: send push notifications
        * LIST: list all push notifications sent to the users
* `resources:user-management:group:<group ids>`
* `resources:user-management:users:<user ids>`
    * actions
        * Block: block users
[...]

PR #162, opened 2018-01-22
Pending for review: CMS roles and policy settings spec

Steven-Chan:
connect #152

### DIFF
--- a/features/152-cms-role-models-and-policy-settings/api.md
+++ b/features/152-cms-role-models-and-policy-settings/api.md
@@ -27,6 +27,25 @@ The whole feature is divided into two parts.
   - CMS user-roles are defined in a separate table.
 - Server will provide API for getting cms config for current user.
 
+#### New `cms-config.yaml` format
+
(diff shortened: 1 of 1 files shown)


Steven-Chan:
@carmenlau @cheungpat 

I have updated the Requirements, please see if it aligns with what we have discussed this morning.

chpapa:
I took a brief look, since I happen to stumble upon it.

I suspect a better phase 1 is actually either:

a. use policy, but the policy server and implementation is in the skygear-cms backend
b. use the split config file, but the skygear-cms backend can block unauthorised request if some table were not configured.

The reason is if it is pure UI side, it is easily hackable and will probably got users complain / clients security audit problems quickly. So even for stage 1 it must not be a UI-only thing.

For backend, if we already have the code in place to block access to a table if the config doesn't have it, than b is better. If not, I think to avoid duplicate efforts, we better go straight to a.

To avoid complicating things to have a policy server inside skygear-server, we may use a library and implement it to only block on CMS plugins side first....

just my two cents.

PR #167, opened 2018-01-29
cms: Basic user management

Steven-Chan:
connect #165


Steven-Chan:
- Remove the three plugin api
    - pass the cms role to client
        - python generated html: pass from plugin env
        - `skygearCMS.start({ adminRole: 'Admin' })`

Steven-Chan:
@carmenlau updated

PR #171, opened 2018-02-07

howawong:
Connects SkygearIO/chat#199

### DIFF
--- a/features/64-chat-offline-cache/fetch_participants.md
+++ b/features/64-chat-offline-cache/fetch_participants.md
@@ -36,35 +36,37 @@ And returns a list of user records.
 
 ```json
 {
-    "result": [{
-        "_access": null,
-        "_created_at": "2018-01-11T07:18:53.791103Z",
-        "_created_by": "23db86bd-4c92-44ee-abcd-eff0217d3207",
-        "_id": "user/23db86bd-4c92-44ee-abcd-eff0217d3207",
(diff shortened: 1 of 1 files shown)


carmenlau:
Three remarks in offline discussion:

1. To ensure the api is scalable, fetch participants api should support both get all users and get users by page. Now the uikit will get all participants after entering the conversation, we will keep it as is and use this api to get all participants. Unless we find performance issue in uikit, we can further optimize later. But we should keep the sdk api scalable in this stage.

2. Getting participants will sorted users by joining time, howa will further check if it is possible.

3. Make sure query user are through skygear query, so it won't bypass the acl

rickmak:
The joining time will be available in `user_conversation` table.

howawong:
After the discussion, I would like to close the PR because we cannot reach a conclusion on pagination and caching on participant list. There are few issues which makes it not practical.

1. It is difficult to maintain the order of the items. For example, if an user is being invited or kicked out of a conversation. The relative order of the items are not stable which cache will be invalidated often.

2. Server contains  sorting related information which might not be cached by SDK/client side. So order might not be identical and cannot be reproduced at SDK in trivial.

3. It is hard to merge a desired cached response from two different responses from plugin side.

Therefore, it is not practical to cache the participant list. Instead, we would like to cache the user object in chat SDK so that user information can be loaded right after message is loaded from cache.

howawong:
I will send a separate PR on user object caching.

PR #176, opened 2018-02-14

howawong:
Connects SkygearIO/chat#199

### DIFF
--- a/features/64-chat-offline-cache/fetch_participants.md
+++ b/features/64-chat-offline-cache/fetch_participants.md
@@ -1,0 +1,82 @@
+#Feature Overview
+
+## Objective
+Allow fetching participants' User records offline given an array of IDs.
+
+## Description
+
+Currently, user records are fetched from publicDB everytime when conversation is fetched, which makes messages fetched from offline cache cannot get the correct sender and avatar immediately. Also, related logic only exists in chat UI Kit, but not in chat SDK.
+
+Therefore, we need a seperate API in chat SDK which
+
+1. Accepts an array of user IDs,
+2. Retrieve a list of user records from publicDB and store in the cache and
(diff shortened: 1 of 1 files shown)


features/64-chat-offline-cache/fetch_participants.md
```diff
@@ -0,0 +1,82 @@
...
+     * @param users cached users
+     */
+    void onGetCachedResult(@NonNull Map<String, ChatUser> chatUsersMap);
+}
+```
+
+#### getChatUsers API
+
+```java
+public void getChatUsers(@NonNull List<String> userIds, GetChatUsersCallback callback) {
```
carmenlau: Would like to know why we user `ChatUsers` in android but `Participants` in ios and js? Just checked a bit and found we had `addConversationParticipants` in android, it would be better to use the same terms.

features/64-chat-offline-cache/fetch_participants.md
```diff
@@ -0,0 +1,82 @@
...
+    /* clang-format off */ NS_SWIFT_NAME(fetchParticipants(participantIDs:completion:));
+```
+
+
+### Javascript
+
+#### getParticipants API
+
+```javascript
+getParticipants(participantIds).then((participantsMap) => {
```
carmenlau: How can we get the cache callback from js?

features/64-chat-offline-cache/fetch_participants.md
howawong: I tries to, but in Android, it is already called ChatUser before this spec and unless we really want to change the name, otherwise we can just keep it.

rickmak:
As per discussion, we will go for using participant across SDK.

PR #179, opened 2018-03-01

howawong:
Connects SkygearIO/chat#199

### DIFF
--- a/features/64-chat-offline-cache/fetch_participants.md
+++ b/features/64-chat-offline-cache/fetch_participants.md
@@ -1,0 +1,104 @@
+# Feature Overview
+
+## Objective
+Allow fetching participants' objects offline given an array of IDs.
+
+## Description
+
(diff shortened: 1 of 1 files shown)


features/64-chat-offline-cache/fetch_participants.md
```diff
@@ -0,0 +1,104 @@
...
+    /* clang-format off */ NS_SWIFT_NAME(fetchParticipants(participantIDs:completion:));
+```
+
+
+### Javascript
+
+#### getParticipants API
+
+```javascript
+getParticipants(participantIds).then((participantsMap, cached) => {
```
carmenlau: It is not possible to call the promise twice and we are not passing the cached callback this way.

Please check the offline cache spec 

https://github.com/SkygearIO/features/blob/master/features/64-chat-offline-cache/api.md

Issue #100, opened 2017-09-04
DB Lambda Functions which return record objects instead of JSON


rickmak:
The current lambda is JSON in and JSON out. However, in many use case, developer want records in and records out; or more often, JSON in and records out. Currently, there is no suggested way or tools to facilitate that.

Proposal:
* Support another type of Cloud Functions (let's say `DB Lambda Functions`) which is a decorator in py-skygear, or something like `.DBop()` in js.
* In `DB Lambda Functions`, we support return record object or array of record objects from skygear query. In the case of Python, would have a util for sqlalchemy result.
* In Client side, the returned record objects can be used directly just like other record objects.
* In Client side, it might be affected by #112 Ideally we should support both Records and custom subclass.

# API Design

## Scenario

This is like the number one painpoint of Skygear usage for Oursky Developers

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

chpapa:
@carmenlau I want to ask if this feature have already covered part of the problem at #4 ? My understanding is partly, as this issue convert JSON <=> Skygear Records. However if users opt to use SQL / ORM, the ORM won't output the right JSON that we can convert to Skygear Records here.

carmenlau:
[...]

PR #187, opened 2018-03-21
Add Lambda support for Skygear Data Type

cheungpat:
connects #100


cheungpat:
@carmenlau Updated, I think this pull request is ready for merging as we have reached consensus in product meeting. Let me know if I have missed anything : )

features/100-lambda-params/api.md
```diff
@@ -0,0 +1,217 @@
...
+
+Since Skygear Record was not considered a data type, the protocol does
+not have support for specifying a Record object. To support using Record object
+in parameters and return value, a new `$type=record` will be implemented.
+
+### Breaking Changes
+
+When upgrading cloud function runtime (i.e. python or node), the function
+arguments will replace `$type` JSON objects with native Skygear Data Type.
+If the developer sends `$type` JSON objects from SDK, they will need to update
```
chpapa: Just want to confirm what I heard from product meeting, we are pretty sure right now most users don't do it? (Maybe a quick survey in office?)

chpapa:
Great work and overall looks good. 

May I know if there are any use cases (or considered?) for Cloud Functions that return an asset object? @cheungpat @carmenlau

limouren:
For `$type=record`, would it cover record's `$transient` field?

features/100-lambda-params/api.md
carmenlau: Did a quick survey, no one in the office will use `$type` in their lambda response, it is safe for us to do it.

carmenlau:
### For asset return
@chpapa 
Discussed with @cheungpat, also asked some colleagues in office, they sometime will return asset from the lambda. To facilitate the user to returning lambda, if user return record with asset or return asset directly. We will signed the asset, and user will be able to give the signed asset in lambda result.

### For $transient
@limouren 
If the returned record in lambda response with `$transient`, we will still return it to the SDK and user will be able to access the `$transient` in record. But if user sending request to lambda with record that contain `$transient`, we will remove and ignore it just like we did record save.

Let me know if there is any further questions or comments? Otherwise I will merge this spec and start the development! Thanks :)

Issue #40, opened 2017-03-29
Scheduled job


cychiuae:
# Description

Something like rqscheduler

# Questions

**Put a list of questions here before if you cannot provide a complete design / specification yet**

# Design/Specification

**Put Design and Specification here with links**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

chpapa:
@cychiuae May you describe some use cases of how you use it in the context for Skygear? :)

PR #188, opened 2018-03-21
Support for job queue

cheungpat:
connects #40

### DIFF
--- a/features/40-job-queue/api.md
+++ b/features/40-job-queue/api.md
@@ -48,6 +48,21 @@ non-trivial because:
   developer need low-level access.
 * Difficult to provide the same level of persistence guarantee in Redis.
 
+## Life time of a job
+
+1. Cloud function call function to enqueue a job.
+2. Plugin send request to server.
+3. Server persist job information to job queue
+4. Server periodically check job queues for job waiting to dispatch.
+5. Lock the job item in the job queue.
+6. Invoke the job like a lambda function.
+7. Save job status to the job queue.
+
+If retry is enabled, the server keep invoking the lambda based on the retry
+count. The lambda is invoked until a retry limit is reached.
+
+Server will run maintenance task to clean job queue for outdated jobs.
+
 ## Sample Usage
 
 Usage will be like this:
@@ -55,7 +70,7 @@ Usage will be like this:
 ```python
 @skygear.job('user:send-email')
 def send_email(user_id):
-    # send logic
+    # send email logic
     pass
 
 @skygear.op('user:register')
@@ -67,7 +82,7 @@ def register_user(email):
                       run_at=tomorrow_datetime())
     print('job_id: ' + job.id)
 
-    # job can be cancelled
-    job.cancel()
-
+@skygear.op('user:deregister')
+def deregister_user(email):
+    cancel_job('send-email:' + email)
 ```
--- a/features/40-job-queue/api.md
+++ b/features/40-job-queue/api.md
@@ -1,0 +1,73 @@
+# Job Queue
+
+## Use Cases
+
+* For cloud function that need to send an email to the user, it is preferable
+  to send the email outside the request lifecycle (i.e. async) because sending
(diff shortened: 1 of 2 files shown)


cheungpat:
Would be great if there are feedback on this article, this is intentionally brief because I would like to gather more feedback before I give more concrete examples, feel free to ask for clarification if there is anything you want to know more : )

chpapa:
Thanks @cheungpat ! I have a few thoughts after reading the PR.

1. I'm a bit confused on use cases between Job Queue and Async Cloud Functions. For example in the document, it said "send an email" "liking a post" as use cases for job queue, but I thought async cloud functions already serve the purpose? I think we need to identify the use cases, as we will also need to write guides and tell people which use cases to use which features.
1. I think some requirements are "good to have" and not a must in version one. Such as cancellation and retrying. (Of coz if do it in one go is easier, let's do it!)
1. I'm more inclined to use Redis (finally) for two reasons:
    1. Seems there are many other features on wish-list that will need Redis. Maybe we would need it anyway at some point.
    1. I'm worried that we will regret using PostgreSQL as queue just like all the trouble we have had using MySQL as queue before. It is not well understood on how to use PGSQL as queues, how to fix problems and tune for performance. In fact, even que author wrote `I don't recommend that you try to run Que in production yet` before and it seems not really popular anyway https://gist.github.com/chanks/7585810; In contrast we use redis as queue all the time, it's popular for the usage; I understand the implication is we might need to do multi-tenant for redis by namespace on application code, but feel like that's easier than troubleshooting at PGSQL for something we are not familiar with. We can use AOF instead of RDB for better persistence?
1. Will we plan to make it possible to enqueue job coded in JS from PY or vice-versa?

cheungpat:
> it said "send an email" "liking a post" as use cases for job queue, but I thought async cloud functions already serve the purpose?

When you are talking about async cloud function I think you are talking about database hooks. Database hooks have support for async call but this capability is only supported in database hooks (e.g. lambda). I am not aware of a method to support a use case such as performing an async job after running a lambda function.

---

The quote in its entirety: 

> I don't recommend that you try to run Que in production yet - the multi-threaded worker pool in particular is still a bit twitchy - but I'm very interested in comments and suggestions on it. My goal is to make it as reliable as possible while keeping its current speed.

Since the author coded Que in Ruby and the cited problem is about multi-threaded worker pool, the problem isn’t really relevant for us.

---

> Will we plan to make it possible to enqueue job coded in JS from PY or vice-versa?

This will be supported.

cheungpat:
> it said "send an email" "liking a post" as use cases for job queue, but I thought async cloud functions already serve the purpose?

When you are talking about async cloud function I think you are talking about database hooks. Database hooks have support for async call but this capability is only supported in database hooks (e.g. lambda). I am not aware of a method to support a use case such as performing an async job after running a lambda function.

---

The quote in its entirety: 

> I don't recommend that you try to run Que in production yet - the multi-threaded worker pool in particular is still a bit twitchy - but I'm very interested in comments and suggestions on it. My goal is to make it as reliable as possible while keeping its current speed.

Since the author coded Que in Ruby and the cited problem is about multi-threaded worker pool, the problem isn’t really relevant for us.

---

> Will we plan to make it possible to enqueue job coded in JS from PY or vice-versa?

This will be supported.

chpapa:
> When you are talking about async cloud function I think you are talking about database hooks. Database hooks have support for async call but this capability is only supported in database hooks (e.g. lambda). I am not aware of a method to support a use case such as performing an async job after running a lambda function.

Understood. Than for consistent and simpler syntax, would it make more sense to just have lambda function support async? It could be a simple wrapper to job dispatch the function as a job? (And it also consistent with DB Hooks in terms of syntax)

> Since the author coded Que in Ruby and the cited problem is about multi-threaded worker pool, the problem isn’t really relevant for us.

My point being it doesn't seems the lib is popular, nor using PGSQL as job queue is well understood in general / in our experience.

Issue #79, opened 2017-06-07
Structured logging


cheungpat:
# Description

* Log output should be structural

# Portal Design/Features

* Console log need to have color to make different kinds of log easier to see
* Support for filtering for log output

# Related Issues

- Server Issues
  - SkygearIO/skygear-server#227
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
  - When you are done with the code, apply the "workflow/code-complete" label.
- Guides
  - Write or modify guides and get them merged in https://github.com/skygeario/guides
  - When the PR of guides is merged, check this checkbox and apply the "workflow/guides-complete" label.

tensiuyan:
The items covered in this issue is very similar to the suggestions I have collected from our colleagues. I would love to further expand the list as follow:

1. Add color to logs to identify logs from difference sources
2. Support log history 
3. Support logs filtering by source, by category and by time

cheungpat:
Created annotated log output with different scenarios:
https://docs.google.com/spreadsheets/d/1iuwco5tPTPp7AWwEaHPP9XeY92P6Tf1zGCIzNf3CmjA/edit#gid=0
Cloud code and raw log:
https://github.com/cheungpat/skygear-cloud-app-annotated-logs

chpapa:
@cheungpat Great! Exaclty what we discussed last time :)

From these logs, can we summarize some "default" criteria that we should give users in the UI? To help us facilitate the UI design?

For example:
* (`javascript` && `exception/error`) || (`python` && `exception/error`) = Cloud Code Error
* (`javascript` && `user log`) || (`python` && `user log`) = Cloud Function Logs

etc

PR #193, opened 2018-03-22
Structured logging spec

carmenlau:
connect #79

### DIFF
--- a/features/79-structured-logging/api.md
+++ b/features/79-structured-logging/api.md
@@ -1,0 +1,65 @@
+# Structured Logging
+
+## Feature Overview
+
+* Support production logs filtering
+* Show logs with color to make different kinds of log easier to see
+
+## Requirements
+
+* Support different filtering to make developer easier to discover the logs they want
+    1. filter by log level (log level: warning, error, info, debug)
+    1. filter by source
+    1. filter by tag (Self defined tags, e.g. Server: db, request, handler..., Py, JS: plugin name)
+    1. group by request ID
+
+* Use color to make the log easier to see
+    1. Different color based on log level
+
+
+## Implementation Details
+
+### skygear-server, py-skygear, skygear-SDK-JS
+
+- logger function to print structured logs
+- Wrap exceptions to structured logs
+- skygear-server create id per request and pass request id to plugin through context
+- Refine all logs and reduce redundant logs in server and all skygear plugins
+
+### Portal
+
+- Filtering will be done in client side
+- Group logs by request ID option, those logs without request id will not be shown in this mode
+- With log level coloring
+- Increase logs size (Current: 1000)
+
+### skycli
+
+- Filtering will be done in client side, filters are passed by arguments
+
+    `skycli --tail 1000 --filters level=info --filters source=server logs`
+
+- With log level coloring
+
+### Cloud
+
+- Need to be able to retrieve logs from completed pod
+
+## Structured log format
+
+```js
+{
+    "time": 1521712900,
+    "level": "info", // debug / info / warn / error / panic / fatal
+    "tag": "db", // 
+    "source": "server",  // server / py / js
+    "request_id": "uuid",
+    "message": "Executed SQL successfully with sql.Queryx",
+    "fields": {          // extra information of logs, optional
+        "args": [],
+        "error": "<nil>",
+        "executionCount": 2,
+        "sql": "SELECT record_type, record_field, user_role, writable, readable, comparable, discoverable FROM \"app_chatdemoapp\".\"_record_field_access\"",
+    }
+}
+```
--- a/features/79-structured-logging/api.md
+++ b/features/79-structured-logging/api.md
@@ -8,14 +8,42 @@
 ## Requirements
 
 * Support different filtering to make developer easier to discover the logs they want
-    1. filter by log level (log level: warning, error, info, debug)
-    1. filter by source
-    1. filter by tag (Self defined tags, e.g. Server: db, request, handler..., Py, JS: plugin name)
-    1. group by request ID
+    1. filter by log level
+        - warning
+        - error
+        - info
+        - debug
+    1. filter by component
+        - Server
+        - Chat plugin
+        - CMS plugin
+        - Forgot password plugin
+        - SSO plugin
+    1. filter by tag
+        - db
+        - requests
+        - handler
+        - pubsub
+        - Other with user input (User can defined his own tag if they use SDK provided logging functions to print log)
+    1. All message / System logs or Errors
+        - All message (No filter applied)
+        - System logs or Errors (All logs sent out not in JSON format, including system logs or exceptions. If user use `print` or `console.log` without using SDK log function in their cloud function, those logs will also be shown here)
+
+* A toggle to allow user group logs by request ID (Logs without request ID will not be shown when this is ON)
 
 * Use color to make the log easier to see
     1. Different color based on log level
 
+## Use cases
+
+1. When developer is developing their own cloud code, they will only want to see the logs of their cloud code
+    - Developer can define tag when calling logs function and filter by custom tag
+    - If developer use print, console.log or their own logger, can choose `System logs or Errors`
+2. For debugging error, developer will want to find the exceptions log easily
+    - choose `System logs or Errors`
+3. When their are multiple requests at the same time, now the logs are mixed up and makes the logs
+difficult to read.
+    - Enable group by request ID
 
 ## Implementation Details
 
@@ -50,12 +78,13 @@
 ```js
 {
     "time": 1521712900,
-    "level": "info", // debug / info / warn / error / panic / fatal
-    "tag": "db", // 
+    "level": "info", // debug / info / warn / error / critical
     "source": "server",  // server / py / js
+    "component": "server", // server / chat / forgot_password / skygear_content_manager / skygear_sso
+    "tag": "db", // db, request, handler, pubsub
     "request_id": "uuid",
     "message": "Executed SQL successfully with sql.Queryx",
-    "fields": {          // extra information of logs, optional
+    "fields": { // extra information of logs, optional
         "args": [],
         "error": "<nil>",
         "executionCount": 2,
--- a/features/79-structured-logging/api.md
+++ b/features/79-structured-logging/api.md
@@ -92,3 +92,55 @@ difficult to read.
     }
 }
 ```
+
+## Sample code
+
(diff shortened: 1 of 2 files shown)


carmenlau:
@cheungpat Would need your help to have a look when have time. I also updated the spec with the sample code of py js logging functions. Welcome for any comments. thanks!!

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+        - debug
+    1. filter by component
+        - Server
+        - Chat plugin
+        - CMS plugin
+        - Forgot password plugin
+        - SSO plugin
+    1. filter by tag
+        - db
+        - requests
```
cheungpat: maybe `request` instead of `requests`

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+        - handler
+        - pubsub
+        - Other with user input (User can defined his own tag if they use SDK provided logging functions to print log)
```
cheungpat: s/his/their/

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+    1. All message / System logs or Errors
+        - All message (No filter applied)
+        - System logs or Errors (All logs sent out not in JSON format, including system logs or exceptions. If user use `print` or `console.log` without using SDK log function in their cloud function, those logs will also be shown here)
+
+* A toggle to allow user group logs by request ID (Logs without request ID will not be shown when this is ON)
+
+* Use color to make the log easier to see
+    1. Different color based on log level
```
cheungpat: 👍

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+        - warning
+        - error
+        - info
+        - debug
+    1. filter by component
+        - Server
+        - Chat plugin
+        - CMS plugin
+        - Forgot password plugin
+        - SSO plugin
```
cheungpat: I think a component is `Cloud Code Functions`, which include user-written logs.

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+    - Enable group by request ID
+
+## Implementation Details
+
+### skygear-server, py-skygear, skygear-SDK-JS
+
+- logger function to print structured logs
+- Wrap exceptions to structured logs
+- skygear-server create id per request and pass request id to plugin through context
+- Refine all logs and reduce redundant logs in server and all skygear plugins
```
cheungpat: Add this

- Output JSON formatted logs on the skygear cloud. Output readable log when deployed in docker compose. Skygear Cloud can supply custom environment variable to enable this behaviour.

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Cloud
+
+- Need to be able to retrieve logs from completed pod
+
+## Structured log format
+
+```js
+{
+    "time": 1521712900,
```
cheungpat: need sub-second precisions

features/79-structured-logging/api.md
carmenlau: Sure, do you think it is good if we use `Cloud Code Functions`  as default value of component for log function? So when user call the log function they don't have to put so many argument.

features/79-structured-logging/api.md
carmenlau: Good idea! Will add.

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,148 @@
...
+    1. All message / System logs or Errors
+        - All message (No filter applied)
+        - System logs or Errors (All logs sent out not in JSON format, including system logs or exceptions. If user use `print` or `console.log` without using SDK log function in their cloud function, those logs will also be shown here)
+
+* A toggle to allow user group logs by request ID (Logs without request ID will not be shown when this is ON)
+
+* Use color to make the log easier to see
+    1. Different color based on log level
+
+## Use cases
```
chpapa: Have we considered all common use cases? One thing I'm not sure if below have covered are:

* I think it is pretty common when people deploy new Cloud Functions, there were syntax error or something that prevent the skygear-server from starting. Would the error here belongs to which component and could be easily spotted from logger in this structured log?

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,148 @@
...
+
+- logger function to print structured logs
+- Wrap exceptions to structured logs
+- skygear-server create id per request and pass request id to plugin through context
+- Refine all logs and reduce redundant logs in server and all skygear plugins
+- Output JSON formatted logs on the skygear cloud. Output readable log when deployed in docker compose. Skygear Cloud can supply custom environment variable to enable this behavior.
+
+### Portal
+
+- Filtering will be done in client side
```
chpapa: Is it really a good idea? I'm worried a lot of browsers will hang if there are too many logs and even modern browser are still not very good at handling these.... (or maybe I was wrong)

features/79-structured-logging/api.md
```diff
@@ -0,0 +1,148 @@
...
+        - db
+        - request
+        - handler
+        - pubsub
+        - Other with user input (User can defined their own tag if they use SDK provided logging functions to print log)
+    1. All message / System logs or Errors
+        - All message (No filter applied)
+        - System logs or Errors (All logs sent out not in JSON format, including system logs or exceptions. If user use `print` or `console.log` without using SDK log function in their cloud function, those logs will also be shown here)
+
+* A toggle to allow user group logs by request ID (Logs without request ID will not be shown when this is ON)
```
chpapa: Sorry for my ignorance, is request ID something automatically generated by skygear-server no matter what now, so that meant users can look at what logs are generated by API requests?

carmenlau:
@cheungpat The JS logging libs: 
https://github.com/winstonjs/winston
https://github.com/trentm/node-bunyan

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+  --level       Apply log level filter, comma separated. Support values: debug,info,warning,error,critical
+  --process     Apply process filter, comma separated. Support values: skygear-server,python,js
+  --tag         Apply tag filter, comma separated. Support values: request,auth,db,pubsub,push,auth_plugin,chat_plugin,cms_plugin,sso_plugin,others
+
+Examples:
+  # select skygear server logs and log level is debug
+  skycli --level=debug --process=skygear-server logs
+
+  # select cms and error logs
+  skycli --tags=cms,others logs
```
cheungpat: `--tag=cms,others`

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+
+  # select cloud functions logs
+  skycli --tags=others logs
```
cheungpat: `--tags=others`

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+````
+
+### Cloud
+
+- Need to be able to retrieve logs from completed pod
```
cheungpat: 👍

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Structured log format
+
+```js
+{
+    "time": 1521712900.511476,
```
cheungpat: I think a ISO8601 timestamp would be more descriptive, something like `"2018-04-26T03:09:18.279977682Z"`. Oh if you have a good reference where you draw inspiration from, let me know 😀

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+        "error": "<nil>",
+        "executionCount": 2,
+        "sql": "SELECT record_type, record_field, user_role, writable, readable, comparable, discoverable FROM \"app_chatdemoapp\".\"_record_field_access\"",
+    }
+}
+```
+
+### Python log function
+
+To keep the
```
cheungpat: To keep the...?

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+- We can use filter to inject current request id
+https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information
```
cheungpat: 👍

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+- .error()
+- .critical()
+
+
+### JS log function
+
+```
+const skygearCloud = require('skygear/cloud');
+
+skygearCloud.log.info('message', {
```
cheungpat: update this to pass the context (request id)

features/79-structured-logging/implementation.md
```diff
@@ -0,0 +1,119 @@
...
+To keep the
+- We can use filter to inject current request id
+https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information
+- Use the logger name for the tag, prefix the name with `tag.`. So if user create their own logger
+in their cloud functions. We can recognized the logs are from cloud functions.
+
+```py
+import logging
+
+logger = logging.getLogger('tag.chat_plugin')
```
cheungpat: We have little control over the logger name, for example, sqlalchemy sets its own logger name, so we cannot tell the sqlalchemy log generated from the chat plugin.

My recommendation is that logger name can be anything, and we do not apply semantic to what logger name means. The logger name is included in the log output as `logger="[loggername]"` field.

The `tag` is added to the log record using log filtering (see the python doc that carmen mentioned 👍 ) and supply the `tag` using thread-local variable. The thread-local variable is set at the time when we receive the request.

For example:

1. Server receive call to `chat:send_message` (the name of the lambda is for illustration purpose (i.e. not real name))
2. Python calls `chat:send_message` lambda.
3. In the lambda, we set `tag=chat` as thread local variable.
4. Lambda call log function
5. Filter get `tag=chat` from thread local variable and add it to log record.

For user lambda, we can have `tag=user` by default, so that the user doesn’t need to know about tag when the user use log functions. skycli/portal can filter based on `tag=user` tag.

I hope this makes the implementation easier as we don’t need to interpret the logger name, at the same time we support the use case of separating user log from plugin logs.

carmenlau:
Updated and thanks for the help @cheungpat 🙏🏻

Issue #220, opened 2018-05-23
[CMS] Support asset import


carmenlau:
# Description

**Add 2-3 sentence description of your feature**

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #224, opened 2018-05-25
[CMS] Import asset

Steven-Chan:
connect #220

### DIFF
--- a/features/220-cms-import-asset/api.md
+++ b/features/220-cms-import-asset/api.md
@@ -1,0 +1,183 @@
+# User Requirement
+
+- Import records with skygear asset field
+- Allow cms user to upload file from cms client, when importing records
+- In the csv, the value of asset field would be the file name of uploaded files
+
+# Overview
+
+Summary of design decision here.
+
+### Keep the cms import one single API call
+
+Import feature for each project can be very different if we want to make a
+cutomized cms for the client, due to the different work flow of the client and
+the different target user.
+
+At the current state, CMS provides the import function as a whole feature,
+meaning if the client requires customization beyond what we provided, it is
+likely the developer need to implement one from scratch themselves.
+
+Thus I HIGHLY recommend to keep this feature as simple as possible.
+
+The cms frontend would be responsible for uploading the files and create a
+mapping of file name to asset id for server.
+
+### Accept zip for assets upload
+
+A more natural approach would be to allow cms user to select files to upload.
+However, considering the behaviour of import that if the field value is empty,
+it would replace the original value with null.
+
+Think of this scenario, a cms user import a csv with some files first, then
+the user make some changes to the csv without touching the asset field, if the
+user does not upload the files again, the server would clear the previous value
+of asset field, because the server cannot find the skygear asset.
+
+So to match user expectation, the cms would require users to upload a single csv
+or a zip file that contains all the files.
+
+# UI Update
+
+(The following does not mean the final design)
+
+### Import button clicked
+
+```
+|--------------------------|
+| Import                   |
+|--------------------------|
+| Upload a csv or zip file |
+| here.                    |
+|                          |
+|                 Continue |
+|--------------------------|
+```
+
+- The whole area of `Upload a csv or zip file` is a drag and drop file upload
+  area.
+- The `Continue` button is disabled now.
+
+### After file uploaded
+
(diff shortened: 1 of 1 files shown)


chpapa:
To be honest, I don't quite understand/get the logic of this feature design, so I'm just commenting inline for what I could think of immediately, feel free to elaborate or we can talk face to face if that's easier.

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+
+Import feature for each project can be very different if we want to make a
+cutomized cms for the client, due to the different work flow of the client and
+the different target user.
+
+At the current state, CMS provides the import function as a whole feature,
+meaning if the client requires customization beyond what we provided, it is
+likely the developer need to implement one from scratch themselves.
+
+Thus I HIGHLY recommend to keep this feature as simple as possible.
```
chpapa: I don't get the logic behind this *Thus*, why this feature should be as simple as possible if *"it is likely the developer needs to implement one from scratch themselves"*?

If this section is intended to be the overview of Design Goal of the feature, it is quite confusing...

From what was written, my guess is, it meant to say there are two design goal:

1. Customization beyond what we provided will be developed by Skygear users from scratch themselves, so there should be some APIs to facilitate that, a.k.a. make it easy for users to develop import feature from scratch.
1. This CMS frontend is not meant to make it possible to use with every customization in the world, so the intended use case are: `[blank]`

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+
+The cms frontend would be responsible for uploading the files and create a
+mapping of file name to asset id for server.
+
+### Accept zip for assets upload
```
chpapa: Why zip file only, but not drag and drop of multiple files?

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+
+A more natural approach would be to allow cms user to select files to upload.
+However, considering the behaviour of import that if the field value is empty,
```
chpapa: Can't get what it meant

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+it would replace the original value with null.
+
+Think of this scenario, a cms user import a csv with some files first, then
+the user make some changes to the csv without touching the asset field, if the
+user does not upload the files again, the server would clear the previous value
+of asset field, because the server cannot find the skygear asset.
+
+So to match user expectation, the cms would require users to upload a single csv
```
chpapa: I understand the above but again, can't get *so* part.

Is it really a surprise if user import another CSV file with identical record ID, and if the asset field is empty it will wipe out the asset field? Sounds like an intended behavior?

If the user want to keep the asset field, maybe he should make sure the CSV imported should have the asset ID inplace?

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+| Import                   |
+|--------------------------|
+| Upload a csv or zip file |
+| here.                    |
+|                          |
+|                 Continue |
+|--------------------------|
+```
+
+- The whole area of `Upload a csv or zip file` is a drag and drop file upload
```
chpapa: I think this issue is about import asset? Or we will have import records and import assets on the same interface? (But I think it would be confusing?)

I guess it is easier for layman users to understand import records and import asset ID as two different things:

1. Import assets: upload a butch of files, we will generate an asset ID and keep it filename somewhere.
2. Import records: upload a CSV or whatever, the asset field should either reference the file by it's asset ID, or filename.

It also imply filename should be unique?

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,183 @@
...
+| - 001.jpg                |
+| - 002.pdf                |
+|                 Continue |
+|--------------------------|
+```
+
+- The browser would unzip the file if needed.
+  (Some js zip library: https://github.com/Stuk/jszip)
+- The file list displays files to be uploaded.
+- If a new file a dropped in the area, the file lists would be replaced with
```
chpapa: Why not just "add the new file" but to "replace" it? Or just a typo?

PR #225, opened 2018-05-25
Add 190-consistent-error

cheungpat:


features/190-consistent-error/api.md
```diff
@@ -0,0 +1,95 @@
+# Consistent error handling
+
+## Requirement
+
+* To support a consistent way of handling error
+* To obtain error information from failed API request
+* To improve TypeScript support with regards to data obtained from rejected
+  promise.
+* Support legacy browsers (aka pre-IE-11).
```
chpapa: Product Design-wise, if it induces too much complexity or difficulties, I think we can drop pre-IE-11 browser support.

features/190-consistent-error/api.md
```diff
@@ -0,0 +1,95 @@
...
+* For other kinds of error, it is preferred to wrap such errors into
+  a SkygearError and use an appropriate error code. Reject the promise
+  with a SkygearError.
+
+### Saving multiple records
+
+When saving multiple records, the returned error depends on how many records are
+being saved and whether the operation is atomic.
+
+To ensure a consistent return value and error for different cases, multiple save
```
chpapa: I have an impression that Javascript folks are ok with inconsistent return value, so we should not really need `saveMultiple`? Or is it because of TypeScript or I just got it wrong?

chpapa:
BTW, is that we have checked there are no inconsistent problem with Error in iOS / Android SDK?

features/190-consistent-error/api.md
cheungpat: Well, I guess if we have inconsistent return value, users might be caught off guard when the return value is not what they expected.

TypeScript is also a big reason we should have consistent return value.

cheungpat:
> BTW, is that we have checked there are no inconsistent problem with Error in iOS / Android SDK?

Since Objective-C/Swift and Java has type checking, error isn’t a problem because we always return error. Anyway we can still take a look and see if we always return an Error object when an error has occurred.

cheungpat:
I have updated the document and I have changed this:

* To reduce the scope of the specification, I propose to separate two kinds of API calls.
  * API that calls the Server API: Will always reject promise with SkygearError
  * API that does not call the Server API: Maybe rejected with SkygearError or Error
  * This is to avoid excessive checking of error throughout the codebase. It is expected that Error on the second case is rare and hiding such errors in SkygearError will makes problem more difficult to discover by the developer.
* For function that returns a promise, the function should not throw—it should reject the promise with an Error/SkygearError instead.

cheungpat:
I would like to add that this suggestion is not intended for broad audience and it only applies to our codebase after considering concerns of relevant parties.

chpapa:
@cheungpat @carmenlau as discussed earlier today, I'm concerned about what is the implication for end user (Developers who use Skygear) when they need to learn Skygear.

My understanding from Carmen are:

For user, 
1. Skygear API always return error by reject Promise; They could know what kind of Skygear Error could happen from Skygear Guide / API References
2. When there are exception throw, it is considered a bug of Skygear and an issue could be created.

For Skygear Team,
1. We won't wrap everything with try-catch, as simply wrapping an error object with Skygear Error object make no sense.
2. We will try to catch the common error and reject promise.

****

Another confusion I have is would there be any problem for developers if we reject with SkygearError or Error?

carmenlau:
In the guides, we should teach user to handle SkygearError only. And in most of the common errors, I think we should reject with SkygearError. If user get Error in reject and the case is common, we should also wrap it so that user can handle? So if user get Error in reject, it is not common case and SDK should handle.

For example, if user use function to sign a asset URL. Even we know the library will have Error if the secret configuration is missing, we should not use the Error from the library. Skygear SDK should check and reject with SkygearError in this case.

What do you think? If it looks good, let me update the spec with that.
@cheungpat @chpapa

chpapa:
@carmenlau Looks good and align with my understanding of the discussion last time. (Just for the record, I still thought it is a better idea to have error and promise reject separately but I'm just making sure what written down was aligned with the discussion conclusion)

I want to confirm if these are mistakes in language or concept, may I confirmed that:

> And in most of the common errors, I think we should reject with SkygearError.

It should be: And **all of** the common errors, we should reject with SkygearError.

Just want to make sure our language are consistent with the conclusion so we won't be confused in future.

PR #226, opened 2018-05-29
Add cms import asset

Steven-Chan:
connect #220

### DIFF
--- a/features/220-cms-import-asset/api.md
+++ b/features/220-cms-import-asset/api.md
@@ -1,0 +1,122 @@
+*Things started with (Possibly) are related but can be separate features.*
+
+# User Requirement
+
+- Allow cms user to upload file from the CMS.
+- In the csv, the value of asset field would be the file name of uploaded files or asset id.
+
+# Overview
+
+### New page for file upload
+
+- New page for cms user to upload file(s) to Skygear. A link to the page can be added to the side bar of the CMS.
+- The page supports file upload, download and listing.
+- CMS Plugin maintains the files with a special record type.
+  - When CMS user upload a file in the page, the server would save a record of this record type, which is a mapping of file name to skygear asset ID, e.g. `abc.jpg` or `folderA/folderB/abc.jpg` -> `pqocmponf-dbqohow-wlkdlbub-ansoidwhz-abc.jpg`
+  - The file name is unique.
+  - Assets created outside the CMS would not appear in this record.
+  - (Possibly) The record contains tags for searching.
+
+### Save record assets field when import
+
+When importing records with csv,
+
+- The value of asset field is either the file name or asset id.
+- If the file name or asset id is not found, it would be treated as empty value, which would clear the old value of the original record.
+
+# Use Case
+
+### Import records with asset
+
+1. CMS user creates a csv, fill the csv with record data. Columns for asset are filled with file name.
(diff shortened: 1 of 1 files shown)


features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,123 @@
+*Things started with (Possibly) are related but can be separate features.*
+
+# User Requirement
+
+- Allow cms user to upload file from the CMS.
+- In the csv, the value of asset field would be the file name of uploaded files.
```
chpapa: filename or the asset ID?

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,123 @@
...
+
+### New page for file upload
+
+- New page for cms user to upload file(s) to Skygear. A link to the page can be added to the side bar of the CMS.
+- The page supports file upload, download and listing.
+- CMS Plugin maintains the files with a special record type.
+  - When CMS user upload a file in the page, the server would save a record of this record type, which is a mapping of file name to skygear asset ID, e.g. `abc.jpg` or `folderA/folderB/abc.jpg` -> `pqocmponf-dbqohow-wlkdlbub-ansoidwhz-abc.jpg`
+  - The file name is unique.
+  - Assets created outside the CMS would not appear in this record.
+  - (Possibly) The record contains tags for searching.
```
chpapa: I think this is too much at least for 1st version.

features/220-cms-import-asset/api.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Save record assets field when import
+
+When importing records with csv,
+
+- The value of asset field is either the file name or asset id.
+- If the file name or asset id is not found, it would be treated as empty value, which would clear the old value of the original record.
+  - (Possibly) Fields in CMS config can be marked as `required`. In the case of asset, if the file name is not found and `required` is `true`, error would be thrown.
```
chpapa: another case we discussed, which I'm not sure does it belong to this spec, is to change the original csv import behavior, so that for `optional` field, if the CSV doesn't contain that field, the columns will be intact (or this is the current behavior?)

chpapa:
Looks good to me, minor clarification depends on what we discussed before.

features/220-cms-import-asset/api.md
Steven-Chan: This part is not about the optional column, but is to enforce the value to be non-empty. Let me update the spec to clarify.

The optional column is out of this scope.

Steven-Chan: This part is not about optional column, but optional value, i.e. enforce the value in csv to be non-empty.

I doubt if this feature is a bit ambiguous, in the case of asset and reference, so I think its better take it out in this spec, and open new issue for discussion.

chpapa: As discussed offline, it should belong to another issue, just writing this down for copy & paste.

The cases for handling required columns and missing data are:

| Column Config | CSV's Column status | Result
| ---- | ---- | ----
| Required | Column is missing | Error at import
| Required | One or more data have empty value | Value is NULL or cleared
| Optional | Column is missing | Data are NOT imported and existing row's field will not be override
| Optional | One or more data have empty value | Value is NULL or cleared

cheungpat:
Looks good to me.

Issue #126, opened 2017-10-24
Remove "type/" from record ID in SDKs


carmenlau:
# Description

Currently, record ID consist of this format: `type/uuid` e.g. (`user/1FD43CF9-934E-4CD9-9463-4A45A1F3F9F0`). Since some SDK/Plugin API expects ID to have no type information, this confuses user as to what Record ID is.

This proposal suggest we use only `uuid` for Record ID instead of `type/uuid`.

For Objective-C/Swift, the `SKYRecordID` will be replaced with string type. For APIs that require `SKYRecordID`, they will be replaced with `String type, String ID` parameters.

# API Design

Before:

```
Record ID is in this format: user/1FD43CF9-934E-4CD9-9463-4A45A1F3F9F0

Record Class:
            RecordID recordID (maybe called `id` in JS)

RecordID Class:
            String recordType
            String recordName

Database Class:
            fetch(RecordID recordID)
            fetchMany(RecordID[] recordIDs)
            delete(RecordID recordID)
            deleteMany(RecordID[] recordIDs)
            save(Record[] records)
```

After:

```
Record ID is in this format: 1FD43CF9-934E-4CD9-9463-4A45A1F3F9F0

Record Class:
            String recordType
            String recordID (maybe called `id` in JS)

Database Class:
            fetch(String recordType, String recordID)
            fetchMany(String recordType, String[] recordIDs)
            delete(String recordType, String recordID)
            deleteMany(String recordType, String[] recordIDs)
            save(Record[] records)
```

# Scenario

* This will mostly benefit new learner/user of the SDK

# Open Questions

* Do we need to change the API backward compatible?

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

# Progress Tracker

- [ ] Specification Design Approval
- [ ] Write (code + tests + API docs) then get them merged. **Put All PR there**
  - [ ] Code
  - [ ] Minimal tests
  - [ ] Minimal API Docs
- [ ] Guides. **Put All PR there**
- [ ] Release
  - [ ] Update release notes
  - [ ] Release

# Advice

- Specification Design Approval
  - Once you get LGTM from another Skygear Core Team Member, you can check this checkbox. And apply the "workflow/design-complete" label.
- Coding
  - Use as many PRs as you need.  Write tests in the same or different PRs, as is convenient for you.
  - API doc should goes in the same PR with the code.
  - As each PR is merged, add a comment to this issue referencing the PRs.
[...]

PR #227, opened 2018-05-29
Add spec for removing `type/` from Record ID

cheungpat:
connects #126

### DIFF
--- a/features/126-record-id/api.md
+++ b/features/126-record-id/api.md
@@ -1,0 +1,73 @@
+# Removing `type/` from APIs
+
+## Overview of changes
+
+This document propose what changes are needed to remove `type/` from Skygear API
+so that developer will have consistent experience when needing to specify
+a Record ID.
+
+Throughout this document the different ID formats are described as:
+
+Deprecated ID - "note/0ebd484c-ed82-4b04-aba7-33fc530f3c88"
+Record Type - "note"
+Record ID - "0ebd484c-ed82-4b04-aba7-33fc530f3c88"
+
+### Changes to SDK API
+
+* API that accepts deprecated ID as parameter should accept a Record Type and
+  a Record ID as parameter, both of `string` type.
+* API that accepts a list of deprecated IDs as parameter should accept a Record
+  Type (string) and a list of Record ID (string).
+* API that returns deprecated IDs (or list) should return Record ID (string),
+  the Record Type is either
+  * Will be omitted if Record Type is already known. For example, delete
+    operation will return a Record IDs, and the Record Type is inferred from
+    the Record Type used in the delete operation.
+  * Will return a Reference instead, which has separate Record ID and Record
+    Type fields.
+  * Other cases will need to be discussed.
+* API that supports deprecated IDs with multiple Record Types will no longer
+  be available. Make multiple API calls instead. Saving record is different,
+  because it operates on a list of Records, not a list of deprecated IDs.
+* Record class and encoded payload should have Record Type separated
+  to a different field.
+
+### Changes to Server API
+
+* Parameters which accept deprecated IDs will be deprecated. New parameters
+  will be added for specifying Record Type and Record IDs.
+* Encoded Record payload will have deprecated ID replaced with a Record Type
+  and a Record ID.
+* Reference will have deprecated ID replaced with a Record Type and
+  a Record ID.
+
+### Payload Format
+
+Old:
+
+```json
+{
+	"_id": "note/note1",
+	"_access": null,
+	"_ownerID": "ownerID",
+	"category": {"$id":"category/important","$type":"ref"},
+	"city": {"$id":"city/beautiful","$type":"ref"},
+	"secret":{"$id":"secret/secretID","$type":"ref"},
+	"_transient": {"secret":null}
+}
+```
+
+New:
+
+```json
+{
+	"_recordID": "b7c66af9-cabd-4e12-bcce-0a873dfb6328",
+	"_recordType": "note",
+	"_access": null,
+	"_ownerID": "ownerID",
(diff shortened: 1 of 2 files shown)


features/126-record-id/api.md
```diff
@@ -0,0 +1,73 @@
...
+Record Type - "note"
+Record ID - "0ebd484c-ed82-4b04-aba7-33fc530f3c88"
+
+### Changes to SDK API
+
+* API that accepts deprecated ID as parameter should accept a Record Type and
+  a Record ID as parameter, both of `string` type.
+* API that accepts a list of deprecated IDs as parameter should accept a Record
+  Type (string) and a list of Record ID (string).
+* API that returns deprecated IDs (or list) should return Record ID (string),
```
chpapa: similarly, have we got an overview of our api, that when shall we return object or ID?

features/126-record-id/api.md
```diff
@@ -0,0 +1,73 @@
...
+  the Record Type is either
+  * Will be omitted if Record Type is already known. For example, delete
+    operation will return a Record IDs, and the Record Type is inferred from
+    the Record Type used in the delete operation.
+  * Will return a Reference instead, which has separate Record ID and Record
+    Type fields.
+  * Other cases will need to be discussed.
+* API that supports deprecated IDs with multiple Record Types will no longer
```
chpapa: just to confirm the only APIs fall into this category are `fetch` and `delete`

features/126-record-id/api.md
```diff
@@ -0,0 +1,73 @@
...
+	"_transient": {"secret":null}
+}
+```
+
+New:
+
+```json
+{
+	"_recordID": "b7c66af9-cabd-4e12-bcce-0a873dfb6328",
+	"_recordType": "note",
```
chpapa: In last meeting we discussed, we might want to have `_id` for compatibility, and a reference data point for the decision is how many Oursky developers are using it, have we asked or researched a bit (Say, do a quick search of relevant API across oursky GitHub repos)

features/126-record-id/implementation.md
```diff
@@ -0,0 +1,210 @@
...
+```javascript
+var Note = Record.extend('note');
+var note = new Note({});
+
+// Old
+console.log(note.id);  // 'note/91c02d33-ea8a-41d3-ae6c-a87e064eaa5c'
+console.log(note.recordType);  // 'note'
+
+// New
+console.log(note.recordID);  // '91c02d33-ea8a-41d3-ae6c-a87e064eaa5c'
```
chpapa: Is it intended to rename `id` to `recordID`? Or it is an unconscious choice?

features/126-record-id/implementation.md
```diff
@@ -0,0 +1,210 @@
...
+
+The Record fetch and delete APIs are changed substantially:
+
+```obj-c
+// Old
+- (void)fetchRecordWithID:(SKYRecordID *)recordID
+        completionHandler:(void (^)(SKYRecord *record, NSError *error))completionHandler;
+
+// New
+- (void)fetchRecordWithID:(NSString *)recordID
```
chpapa: Is there any reason to change SKYRecordID to string? (Just asking, no thought from me)

features/126-record-id/api.md
```diff
@@ -0,0 +1,73 @@
...
+
+Throughout this document the different ID formats are described as:
+
+Deprecated ID - "note/0ebd484c-ed82-4b04-aba7-33fc530f3c88"
+Record Type - "note"
+Record ID - "0ebd484c-ed82-4b04-aba7-33fc530f3c88"
+
+### Changes to SDK API
+
+* API that accepts deprecated ID as parameter should accept a Record Type and
```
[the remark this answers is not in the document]
cheungpat: I think we will handle this with the cross platform API consistency issue. I think I am with you that for parameters that take ID, we should also take object.

features/126-record-id/api.md
cheungpat: That depends on the API?

fetchRecord: ID > Record
saveRecord: Record > Record
deleteRecord: (ID or Record) > ID (returning an object after deleting it doesn’t seem intuitive)

features/126-record-id/api.md
cheungpat: Yes.

features/126-record-id/api.md
cheungpat: I didn’t. Let me try ask around.

On the protocol level, I think we will have `_id` for compatibility. For JS SDK, I think keeping `_id` will confuse people, so we should remove that.

features/126-record-id/implementation.md
cheungpat: My idea is that using string ID is easy to understand for most people. For new learner, they don’t need to figure out how to construct a `SKYRecordID`.

I am open to suggestion for change here, as having `SKYRecordID` might help type safety.

features/126-record-id/implementation.md
cheungpat: It is renamed to recordID to keep consistency with other platforms. Android SDK has `recordID` as well.

cheungpat:
I have updated the documents will minor changes.

When the SDK has made changes to separate record type and record ID information, I think it is beneficial if we have the same separation on the server-side API as this will make the two APIs easier to understand. I understand that it is not a requirement for the server-side API to match the client-side API, but I don’t see why matching both sets of API is a bad thing. That’s why I will keep the record type and record ID separated in the payload.

An effort will be made to keep both formats in the Record Payload format, so that old clients are still compatible with new version of server. In the long run, we will remove the deprecated fields.

I think this is the best I can come up with, having considered the opinions of the relevant parties. If you have a better idea. Please feel free to make a pull request and I will be happy to include them.

chpapa:
I don't have any more comment at this stage, thanks!

PR #238, opened 2018-07-10
Add cms validation spec

Steven-Chan:
connect #237

### DIFF
--- a/features/237-cms-validation/api.md
+++ b/features/237-cms-validation/api.md
@@ -1,0 +1,146 @@
+# CMS Validation
+
+## Requirements
+
+- Prevent cms user from entering invalid data.
+- Allow developer to use common and customisable validation in cms config on field level and form level.
+- Display error message, which can be defined in config, for invalid data.
+
+## Common Use Cases
+
+- Non empty text input.
+- Text input with max length.
+- Text input with certain pattern, e.g. credit card number.
+- Number or datetime input value within range.
+- One field cannot be empty if another field is not empty.
+
+## Overview
+
+- Validation is a process that evaluate data with an expression, the result is either success or failure.
+- Validation can be defined in record edit page and new page in cms config.
+- Validation happens after use click save button and before all server api calls are fired.
+- If any validation result is failure, error message should be displayed and no server api calls should be fired.
+- Error message of field validation appears next / below the field, while that of form validation appears below the form.
+- Field validation is executed before form validation.
+
+## Configuration
+
+- One validation item is a dictionary which contains the validation configuration (depends on where it is placed).
+  - Each validation optionally contains a `message` field, which would be the error message if that validation item cannot pass.
+  - Otherwise, a default error message would be shown.
+- `validation` may be a validation item or contains a list of validation items.
+
+```
+new:
+  # page validation
+  validation:
+    - expression: xxx
+    - expression: xxx
+  fields:
+    - name: xxx
+      type: xxx
+      # field validation
+      validation:
+        pattern: /^xxx$/
+        message:
+```
+
+## Expression
+
+- `expression` is the base type of validation, supported by all validation.
+- An expression is a string, which can be evaluated to give boolean value.
+- All other types of validation are predifined expression, i.e. they are all compiled to expression at runtime.
+- Expression evaluation would be based on this library https://github.com/joewalnes/filtrex, while CMS would provide extra predefined functions for common use case.
+
+### Functions
+
+- Passing wrong data type to the function would throw error in validation.
+
+### Expression context: `value`
+
+Developer uses the variable `value` in the expression to get the value from the field or form.
+
+- Field validation
+  - `value`: value of the field
+  - e.g. `value in ('abc', 'def')`
+- Form validation
+  - `value`: value of the form
+  - `get(value: formValue, fieldName: string)`
+  - e.g. `get(value, "full_name") != get(value, "nick_name")`
+
+#### Predefined functions
+
+- String
+  - `length(value: string)`
+  - `upper(value: string)`
+  - `lower(value: string)`
+  - `substring(value: string, from: number, to: number)`
+- Number
+- Datetime
+  - `datetime(str: string)`
+  - `timestamp(data: datetime | string)`
+    - e.g. `timestamp(value) > timestamp('2018-01-01')`
+- JSON
+  - `get(data: json, key: string | number)`
+    - return the value at key
+    - e.g. `get(get(value, "a"), "b")`, `{"a": {"b": 1}}` => `1`
+    - e.g. `get(get(value, "a"), 0)`, `{"a": [1]}` => `1`
+  - `hasKey(data: json, key: string | number)`
+    - return if the key exist
+- Location
+- References
+  - `length(value: reference[])`
+- EmbeddedReference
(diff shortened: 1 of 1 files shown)


features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+## Requirements
+
+- Prevent cms user from entering invalid data.
+- Allow developer to use common and customisable validation in cms config on field level and form level.
+- Display error message, which can be defined in config, for invalid data.
+
+## Common Use Cases
+
+- Non empty text input.
+- Text input with max length.
```
chpapa: since it is "common" use cases, shall we have a predefined validation below too?

features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+- If any validation result is failure, error message should be displayed and no server api calls should be fired.
+- Error message of field validation appears next / below the field, while that of form validation appears below the form.
+- Field validation is executed before form validation.
+
+## Configuration
+
+- One validation item is a dictionary which contains the validation configuration (depends on where it is placed).
+  - Each validation optionally contains a `message` field, which would be the error message if that validation item cannot pass.
+  - Otherwise, a default error message would be shown.
+- `validation` may be a validation item or contains a list of validation items.
```
chpapa: does it meant, to configure multiple validation for the same field, it should be like...

```
new:
  fields:
    - name: xxx
       type: xxx
       validation:
         validation:
           pattern: /^xxx$/
         validation:
           required: true
```

If that's the case, would it be better call it `validations`?

features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Predefined functions
+
+- String
+  - `length(value: string)`
+  - `upper(value: string)`
+  - `lower(value: string)`
+  - `substring(value: string, from: number, to: number)`
+- Number
+- Datetime
```
chpapa: should we have some function for date range operation? Like "with in 7 days" or "within the same month" / "within the same week" etc?

features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+- required: true
+# String / References / EmbeddedReferences equivalent to
+- expression: length() > 0
+# Datetime / Asset equivalent to
+- expression: value != null
+```
+
+Note that, `required: false` is no-op.
+
+### `pattern`
```
chpapa: is it regular expression? If yes call it `regex` (etc) instead of `pattern`?

features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+  - `length(value: reference[])`
+- EmbeddedReference
+  - `get(data: embeddedReference, key: string)`
+- EmbeddedReferences
+  - `length(value: embeddedReference[])`
+  - `get(data: embeddedReference[], index: number)`
+- Asset
+  - `size(value: asset)`
+
+## Predefined validation for field validation
```
chpapa: I think it is better to start with a list of common predefined validation too, besides `required`, e.g.

`max-length` / `min-length`
`email` / `creditcard` / `url`
`image-width` / `image-ratio` etc
`with-in-range` (for numbers)

etc?

carmenlau: ummm... When I review the guides of cms config and recalls me that we used camelcase for predicate too. Maybe we should decide, which case we should use camelcase or underscore. Or should we change all to underscore? 🙈

features/237-cms-validation/api.md
ben181231: So there should be some functions like `getMonth(value)` or `getWeekOfYear(value)`?

features/237-cms-validation/api.md
```diff
@@ -0,0 +1,146 @@
...
+
+Developer uses the variable `value` in the expression to get the value from the field or form.
+
+- Field validation
+  - `value`: value of the field
+  - e.g. `value in ('abc', 'def')`
+- Form validation
+  - `value`: value of the form
+  - `get(value: formValue, fieldName: string)`
+  - e.g. `get(value, "full_name") != get(value, "nick_name")`
```
ben181231: We don't need the `value` param here?

features/237-cms-validation/api.md
Steven-Chan: Configuration for more than one validation.

```
fields:
- name: xxx
   type: xxx
   validation:
     - expression: xxx
     - expression: xxx
```

features/237-cms-validation/api.md
Steven-Chan: `value` for form validation is actually a dictionary, so the function signature is the same as `get` of `JSON`.

Besides, all predefined functions are pure, so I don't think its good to break the rule here.

features/237-cms-validation/api.md
Steven-Chan: > So there should be some functions like getMonth(value) or getWeekOfYear(value)?

Agree, I will add those functions, and with a function `now()`, probably would do what @chpapa mentioned.

features/237-cms-validation/api.md
Steven-Chan: > Maybe we should decide, which case we should use camelcase or underscore. Or should we change all to underscore?

Didn't take care of it when writing this guide, I will update this spec.

features/237-cms-validation/api.md
Steven-Chan: What about both `regex` and `pattern`, while `pattern` can be something like `email`, `creditcard` and `url`, which are some predefined regex.

features/237-cms-validation/api.md
chpapa: @Steven-Chan so that will also be:

```
validation:
  - pattern: /^xxx$/
  - required: true
  - expression: xxx
```
?

BTW that should be `validations`?

features/237-cms-validation/api.md
chpapa: I'm not familiar with the expression language, so someone should try to look at it and consider if it make sense to make `getMonth(value)`/`getWeekOfYear(value)`, or something like `daysAfter(date, n)`.

My problem with `getMonth` or `getDay` etc, is usually it is not straightforward, for example, if you get the day is 28, and month is 5, and wanna make it 5 days later, you can never simply add 5 to day (28+5=33, and obviously there is no May 33rd)

features/237-cms-validation/api.md
chpapa: good idea.

features/237-cms-validation/api.md
carmenlau: maybe we can handle this by the timestamp function?

features/237-cms-validation/api.md
ben181231: I see.

features/237-cms-validation/api.md
Steven-Chan: You're right, like other config keys `fields`, `actions`, they are plural and only accept array. For consistency, I will update the configuration here to follow other config.

Steven-Chan:
@chpapa @carmenlau @ben181231 updated.

carmenlau:
@ben181231 @chpapa Steven updated the spec, please kindly check. If it looks good, I will merge it and we can start working on it :)

Thanks!!

PR #239, opened 2018-07-13
Update spec of new Record ID for backward compatibility

ben181231:
This commit is to update the spec of `record:delete`
action using new Record ID format for backward compatibility.

refs #126


cheungpat:
@ben181231 oh sorry you are right, totally missed that

PR #241, opened 2018-07-19
Update record api consistency spec

carmenlau:
- Use list of result object to represent the result of non-atomic api
- Adding `QueryResult` in iOS and Android for presenting records and overallCount, we had this in JS already
- Adding success callback with the first element in Android, it is used for when user fetch, save or delete single record with the api.


carmenlau:
Replaced by #243

PR #243, opened 2018-07-19
[API Consistency] Update fetch, non-atomic save and delete api

carmenlau:
Update summary:

- For fetching single record, the callback is simple, success callback will be called if record is found, otherwise fail callback will be called
- For fetching multiple records, save and delete records non-atomically, fail callback will be called only when there are operational error. For any individual errors, it will return in the success callback with the result object. The result representation in 3 platforms will be 
    - JS, Union type result object
        - `type FetchResult = Record | Error`
        - `type NonAtomicSaveResult = Record | Error`
        - `type NonAtomicDeleteResult = String | Error`
    - iOS Objective C, individual result object with record and error
        - `SKYFetchResult`, `SKYNonAtomicSaveResult` and `SKYNonAtomicDeleteResult`
    - Android Java and Koltin
        - Generic `RecordResult<T>` that T will be Record in fetch and save, T will be String in delete
- In Android, it is strange that handler always return array even we use single record save and delet api. We updated to use generic in `RecordFetchResponseHandler<T>`, `RecordSaveResponseHandler<T>` and `RecordDeleteResponseHandler<T>` to handle this problem.
- In JS, save and delete api, if user save with array that contain 1 record. We resolved with single record instead of array with 1 record. To solve this, we separate single and multiple records function, so that we can have api input output symmetric.
- Support both delete records by ids or by records in all SDKs

### DIFF
--- a/features/207-sdk-api-consistency/record-api.md
+++ b/features/207-sdk-api-consistency/record-api.md
@@ -32,23 +32,29 @@ https://github.com/SkygearIO/features/pull/227 is considered in this doc.
 
 ```objc
 - (void)performQuery:(SKYQuery *)query
-    completionHandler:
-        (void (^_Nullable)(NSArray *_Nullable results, NSError *_Nullable error))completionHandler;
+          completion:
+            (void (^_Nullable)(NSArray *_Nullable results, NSError *_Nullable error))completion;
 
 - (void)performCachedQuery:(SKYQuery *)query
-         completionHandler:(void (^_Nullable)(NSArray *_Nullable results, BOOL pending,
-                                              NSError *_Nullable error))completionHandler;
+                completion:(void (^_Nullable)(NSArray *_Nullable results, BOOL pending,
+                                              NSError *_Nullable error))completion;
 
 - (void)fetchRecordWithType:(NSString *)recordType
                    recordID:(NSString *)recordID
-          completionHandler:(void (^)(SKYRecord *record, NSError *error))completionHandler;
+                 completion:(void (^_Nullable)(SKYRecord *_Nullable record,
+                                               NSError *_Nullable error))completion;
+
+@interface SKYFetchResult
+
+@property (nonatomic, readonly) SKYRecord *_Nullable record;
+@property (nonatomic, readonly) NSError *_Nullable error;
+
+@end
 
 - (void)fetchRecordsWithType:(NSString *)recordType
                    recordIDs:(NSArray<NSString *> *)recordIDs
-           completionHandler:(void (^)(NSDictionary<NSString *, SKYRecord *> *recordsByRecordID,
-                                      NSError *operationError))completionHandler
-      perRecordErrorHandler:(void (^)(NSString *recordID,
-                                      NSError *error))errorHandler;
+                  completion:(void (^)(NSArray<SKYFetchResult*> *results,
+                                       NSError *operationError))completion;
 ```
 
 #### Android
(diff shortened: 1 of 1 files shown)


features/207-sdk-api-consistency/record-api.md
```diff
@@ -313,17 +488,22 @@ del(records: Record | Record[] | QueryResult): Promise<DeleteResult | DeleteResu
 
 ##### New
 
+- Deleted record is an `Record` Object that contain `_recordID`, `_recordType` and `deleted`,
+`deleted` is `true` and the other part of record is empty.
```
ben181231: I think the key should be `_deleted`.

features/207-sdk-api-consistency/record-api.md
```diff
@@ -240,25 +353,64 @@ saveNonAtomically(records: Record[]): Promise<NonAtomicSaveResult[]>;
...
+                                    NSError *_Nullable error))completion NS_NS_REFINED_FOR_SWIFT;
+
+// results contain deleted record
+- (void)deleteRecordsNonAtomically:(NSArray<SKYRecord *> *)records
+                        completion:(void (^_Nullable)(NSArray<SKYRecordResult<Record*>*> *_Nullable results,
+                                                      NSError *_Nullable error))completion NS_NS_REFINED_FOR_SWIFT;
+```
+
+```swift
+enum SKYRecordResult<T> {
```
ben181231: s/SKYRecordResult/RecordResult/

features/207-sdk-api-consistency/record-api.md
```diff
@@ -32,23 +32,44 @@ https://github.com/SkygearIO/features/pull/227 is considered in this doc.
...
+                 completion:(void (^_Nullable)(SKYRecord *_Nullable record,
+                                               NSError *_Nullable error))completion;
+
+// Generic RecordResult object for fetch, non-atomic save and delete
+@interface SKYRecordResult<ObjectType> : NSObject
+
+@property (nonatomic, readonly) ObjectType value;
+@property (nonatomic, readonly) NSError* error;
+
+@end
```
ben181231: May I clarify that this class will not exist in Swift? Otherwise, the coexistence of `SKYRecordResult` and `RecordResult` would be very confusing.

features/207-sdk-api-consistency/record-api.md
```diff
@@ -240,25 +353,64 @@ saveNonAtomically(records: Record[]): Promise<NonAtomicSaveResult[]>;
...
+
+// Generic RecordResult object for fetch, non-atomic save and delete
+@interface SKYRecordResult<ObjectType> : NSObject
 
+@property (nonatomic, readonly) ObjectType value;
+@property (nonatomic, readonly) NSError* error;
+
+@end
+// results contain deleted record id
 - (void)deleteRecordsWithTypeNonAtomically:(NSString *)recordType
```
cheungpat: See if you want to put NonAtomically to the middle so that `deleteRecordsNonAtomicallyWithType:`

features/207-sdk-api-consistency/record-api.md
carmenlau: I updated both obj-c and swift using the same name `SKYRecordResult`. In obj-c, I added `NS_SWIFT_NAME` to rename the object and having enum implementation in swift. See if it looks good? @ben181231 @cheungpat 

commit: bb27194

features/207-sdk-api-consistency/record-api.md
carmenlau: Updated, thanks!

features/207-sdk-api-consistency/record-api.md
ben181231: 👍

PR #244, opened 2018-07-23
Update cms validation spec, add `when` config to simplify null handling

Steven-Chan:
connect #237


chpapa:
@Steven-Chan I'm reviewing the whole CMS spec, not just the pull request, here are what I thought:

* I don't have much comment for null value handling, I think it looks fine. However, if we only focus on the "writing part of the spec", I would suggest:
  * Write down what field types could have `null` value, for example, I guess string always have an empty string instead of `null`?
  * Use an example about `null` value handling in the paragraph? Currently, the example given is not about `null`, but a case of `admin` string.
  * I think it will help future reader if we describe the whole use case under *null value handling* first (including the default `when` config of `value != null`), and go into details one by one afterwards.

Some other comments on part not relevant:

* Under `pattern`, I don't think it should be "predefined regex", for example we probably won't check email solely by regex?
* Mentioned in earlier PR, I guess `get_hour`, `get_minute` etc won't be enough for validation expressions, I would suggest we should include some common example and see how it was done. For example, how do we make sure a date-time field is within next 45 mins?

Steven-Chan:
@chpapa @carmenlau I've updated the spec.

I've addressed ben's comments, except the date time functions, I left a TODO there.

features/237-cms-validation/api.md
```diff
@@ -211,3 +214,105 @@ validations:
     min: 2000-01-01
     max: 2001-01-01
 ```
+
+### Null value handling
+
+Cases where `null` value appears:
+
+- value cleared for `Datetime`, `Reference`, `EmbeddedReference`, `Asset` field
+- value was originally `null` and not updated when save
```
chpapa: Sorry does it meant:
1. if the value was not originally `null` and not updated, the value will be the old value
2. if the value was originally `null` or not updated when save, the value will be `null`

features/237-cms-validation/api.md
```diff
@@ -211,3 +214,105 @@ validations:
...
+
+# example
+- expression: value == null or length(value) > 10
+```
+
+`when` config is introduced to simplify the expression.
+
+- accepts an string expression which evaluates to give boolean value, like `expression`.
+- the effective boolean expression would be: `not (when) or (expression)`
+- `when` is optional, default `when` expression: `true`
```
chpapa: Does it contradict with the statement "Thus we provide a default when config, value != null, for all validations except required." below?

features/237-cms-validation/api.md
```diff
@@ -211,3 +214,105 @@ validations:
...
+```yaml
+# example
+- type: string
+  validations:
+    - when: value != null and regex(value, "^admin")
+      length:
+        max: 15
+    - when: value != null and not regex(value, "^admin")
+      length:
+        max: 10
```
chpapa: For readability: better show how to mix "expression" with predefined criteria here.

features/237-cms-validation/api.md
Steven-Chan: > if the value was not originally `null` and not updated, the value will be the old value

Yes.

To be more accurate, if the field is not updated, the value will be the old value, no matter what value it was.

> if the value was originally `null` or not updated when save, the value will be `null`

Do you mean:
if the value was originally `null` ~~or~~ and not updated when save, the value will be `null`


In cms, the empty value of each field is different, for example, empty value of string fields is empty string, while that of asset field is null. Thus some field will never produce a null value, once the cms user has edited the field.

At this moment, there is not a generic config for cms developer to specify the nullability a field, except dropdown.

features/237-cms-validation/api.md
Steven-Chan: I think validation can be a generic config across the whole cms, e.g. maybe later some config in user management also support validation. So i write this as a general description.

While currently we only support field validation and form validation, field validation is an exception for this.

carmenlau: Maybe we can mention field validation has different default here to avoid confusion?

carmenlau:
If there is no further comments, I will merge it :)

PR #254, opened 2018-10-24
next: Design SDK callback response for auth gear

kaojohnny:
connects #253 

## Note

1. Mock record type implementation will be in server side.
2. Add a `_auth_user_profile` table.
3. Add a `SimpleUserProfileStore` which confirms `UserProfileStore` interface.

This PR won't cover following topics:
1. Implementation detail of following properties/APIs:
   1. `updateUser(user<Record>)`
   2. `isUserDisabled(user<Record>)`
   3. `isUserVerified(user<Record>)`
   4. `isCurrentUserDisabled`
   5. `isCurrentUserVerified`
2. Handling about a developer uses record gear to save auth info.
3. Migration from without record gear to with record gear.
4. API design of user query of auth gear.
5. Should we have a `UserReocrd` type (in SDK)?


Steven-Chan:
I think the following is missing in the spec:

- how `updateUser` is implemented, how it behaves on SDK and server side, and what happens when record gear is present or absent
- how the changes affect other SDK that use statically typed language

carmenlau:
There are some detail that we have discussed we may also want to cover in the spec.
1. How to resolve inconsistency if developer update user profile by record save?
   Some options:
    1. Developer can only use auth gear to update user, record gear should block developer from saving reserved fields, record save should be denied. In this way we need to figure out how to distinguish request from gear, since both auth gear and sdk will call record gear through public gateway
    1. Allow saving user record too, user should be responsible to the data inconsistent if they don’t do it through auth.updateUser
    1. Maybe we don’t help developer to copy anything to user profile, developer can choose to do it when signup. In current signup function username and profile are different params, if developer need username in profile they need to provide it in the profile params. That can help developer to understand auth_data and user record are different data? And it looks more obvious
    1. Any other options?

1. For the api that need user as params, developer should provide core user or user record? How can we avoid the confusion (Question for option 1 and 3)
    1. Should accept core user since record is optional, for statically typed language we should not have confusion. For JS, maybe we should accept both core user and user record? If user pass record to the function, underlying we can wrap user as core user with id. In this case the core user object will not have information of disabled or verified...etc

1. I think we cannot avoid discussing the query other user case (both admin or non admin)? As I think this is one of the most confusing part, as developer will have user object and user record for option 1 and 3.

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,154 @@
...
+     const phone = skygear.auth.currentUser['phone'];
+     
+     // update record
+     skygear.auth.currentUser['username'] = 'new-username';
+     skygear.auth.currentUser['phone'] = 'new-phone';
+     skygear.auth.updateUser(skygear.auth.currentUser);
+     
+     // it is uncertain for now if developer saves via record gear
+     skygear.auth.currentUser['username'] = 'new-username';
+     skygear.publicDB.save(skygear.auth.currentUser.profile); // ??? what should happened?
```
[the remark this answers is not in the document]
Steven-Chan: > Keep the currentUser as Record.

You have rejected option 1 to make this statement here, but actually I lean towards option 1 more than the other two options.

kaojohnny:
I had added some more discussions to the PR, please help check if it is proper.

kaojohnny:
@Steven-Chan I made some updates, when you have time, please help review again. Thanks.

rickmak:
> Not sure what put in record alike object yet.
Not record alike object. It is a Record object.

> Should we have a UserReocrd type.
I think it is SDK specific.

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,209 @@
...
+    "_updated_by": "<user_id>",
+    "email": "abc@example.com"
+}
+```
+
+To handle the case of record gear missing, auth gear will add a new table to save user's profile.
+
+```
+CREATE TABLE _auth_user_profile (
+  user_id text REFERENCES _core_user(id),
```
carmenlau: Maybe we should add unique key to user_id, every user should only have 1 auth_data

carmenlau: Or we should call it profile for the field name? 🤔

kaojohnny: @carmenlau I added UNIQUE, and update the field name to `data`.

rickmak:
I think @chpapa can do the final read the merge.

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,215 @@
...
+skygear.auth.isCurrentUserVerified;
+skygear.auth.isCurrentUserDisabled;
+
+// current user profile record
+const phone = skygear.auth.currentUser['phone'];
+
+// update record
+skygear.auth.currentUser['username'] = 'new-username';
+skygear.auth.currentUser['phone'] = 'new-phone';
+skygear.auth.updateUser(skygear.auth.currentUser);
```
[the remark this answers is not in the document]
kaojohnny: @chpapa I am still writing down some example codes about how to access user auth info, before I finish, let me add some short responses to you. 

Followings are based on current implementation direction,:

1. for `skygear.db.save(user)` case, this behavior is forbidden, a user can only update user profile via new `skygear.auth.updateUser(...)` API through auth gear to record gear. The reason here is for auth data (username, email) consistency.
2. for other user auth properties (roles, last login at, last seen at, ...),
   - for current user, they can be accessed via new properties, like: `skygear.auth.currentUserLastLoginAt`, `skygear.auth.currentUserSignupAt`.
   - for the case to access other user's auth properties, they can be accessed via new APIs, like: `skygear.auth.getUserLastLoginAt(userRecord)`, `skygear.auth.getUserSignupAt(userRecord)`

Feel free to give me your concerns. Thanks.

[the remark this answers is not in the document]
chpapa: Thanks @kaojohnny !

1. Okay, so we need to think about how to explain this to users...
2. Shall we consider generalize how we treat current users and other users? E.g. change `skygear.auth.currentUserLastLoginAt` to `skygear.auth.getUserLastLoginAt(skygear.auth.currentUser)`?
    * Main reason is just thought it would help make the API references / guide easier? (Is it easier to understand?)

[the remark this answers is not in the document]
kaojohnny: Thanks for your comments @chpapa, I will add them to another issue (https://github.com/SkygearIO/features/issues/256).

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,393 @@
...
+  created_by text,
+  updated_at timestamp without time zone NOT NULL,
+  updated_by text,
+  data jsonb,
+  PRIMARY KEY(user_id),
+  UNIQUE (user_id)
+);
+```
+
+Auth gear will also have a store `SimpleUserProfileStore` which confirms `UserProfileStore` interface:
```
chpapa: s/confirms/implements/?

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,393 @@
...
+
+## update user profile
+
+```javascript=
+var user = skygear.auth.currentUser;
+user["username"] = "new-username";
+skygear.auth.updateUser(user).then((user) => {
+  console.log(user); // updated user record
+  console.log('Username is changed to: ', user["username"]);
+  return skygear.auth.whoami();
```
chpapa: is `skygear.auth.whoami()` needed? Or `updateUser` shall update the current user automatically?

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,393 @@
...
+```javascript=
+console.log(skygear.auth.currentUser["username"]);
+console.log(skygear.auth.currentUser["email"]);
+console.log(skygear.auth.currentUser["gender"]);
+```
+
+# access current user auth info
+
+```javascript=
+console.log(skygear.auth.isCurrentUserVerified);
```
chpapa: As mentioned above, is it better to use `skygear.auth.isCurrentUserVerified` or `skygear.auth.isUserVerified(skygear.auth.currentUser)`?

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,393 @@
...
+```javascript=
+const users = [skygear.auth.currentUser];
+skygear.auth.fetchUserRole(skygear.auth.currentUser).then((roleMap) => {
+    console.log(roleMap); // { "<currentUsrId>": ["admin", "editor"] }
+}, (error) => {
+    console.error(error);
+});
+```
+
+## access aother user's profile
```
chpapa: Is the ACL for other user's profile same as Record's ACL?

chpapa:
@kaojohnny left my comments above. If there will be another PR to address them feel free to close it and continue... cc: @rickmak

features/253-auth-sdk-callback-in-next/api.md
kaojohnny: As I know, they should follow same policy.

features/253-auth-sdk-callback-in-next/api.md
kaojohnny: @chpapa We will discuss user attribute design in another PR, and https://oursky.quip.com/ewfGAfyodkb0 listed draft API design. Feel free to leave comments there. Thanks.

kaojohnny:
Update:

1. remove record gear related spec.
2. remove admin API spec in client SDK.

carmenlau:
@kaojohnny 

1. Do we need api for user to update `username` and `email`?
2. Will we consider adding `avatar_url` and `preferred_lang` later? For this 2 fields, we should have better support when we add them. For example, SDK may want to support having api to upload avatar or support autofill with sso profile avatar... etc. `preferred_lang` should follow language code format but not a free text, so we can have better support in future user management dashboard.

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -25,233 +25,143 @@ skygear.publicDB.save(user).then((user) => {
...
+    email: <email>,
+    metadata: {
+        avatar_url: <avatarUrl>,
+        name: <name>,
+        nickname: <nickname>,
+        birthday: <birthday>,
+        preferred_lang: <preferred_lang>,
+        
+        // any other custom attributes
+        ...
```
Steven-Chan: First thing want to confirm about `AUTH_RECORD_KEYS` and metadata. If developer configure other keys, e.g. phone, as `AUTH_RECORD_KEYS`, would that be a top level key? And if username and/or email are not in `AUTH_RECORD_KEYS`, are they part of predefined attributes in metadata?

Second, I think when considering what are predefined attributes, we need a real use case.
For example, avatar_url are commonly used when link with social account and maybe chat app. However, nickname and preferred_lang seems no real use case. 
It was mentioned before that these attributes are for user portal, however, if we now support custom attributes in meta data, and the support for both predefined and custom attributes should be the same, unless the list of predefined attributes is large enough that cover different area, which is pretty impossible. So I don't think user portal is a strong supporting use case when defining user portal.

kaojohnny:
> @kaojohnny
> 
> 1. Do we need api for user to update `username` and `email`?
> 2. Will we consider adding `avatar_url` and `preferred_lang` later? For this 2 fields, we should have better support when we add them. For example, SDK may want to support having api to upload avatar or support autofill with sso profile avatar... etc. `preferred_lang` should follow language code format but not a free text, so we can have better support in future user management dashboard.

@carmenlau 

1. I do think we need support update username or email, I will update the spec to add an API: `changeAuthRecordKey({"email": "<new_email>"})`.
2. We better define "pre-defined user attributes" by now but can leave them blank temporarily, this is because it could very hard to add "pre-defined user attributes" back after launching first release (especially when app implements similar functions). 
3.  In current spec, I defined `preferred_lang` using RFC 5646 format, ex: 'en', 'zh-TW', 'zh-CN', 'zh-HK'. Do you think this is okay?

Steven-Chan:
> 3\. In current spec, I defined `preferred_lang` using RFC 5646 format, ex: 'en', 'zh-TW', 'zh-CN', 'zh-HK'. Do you think this is okay?

@kaojohnny Do you mean skygear support this as a field type, like location?

features/253-auth-sdk-callback-in-next/api.md
kaojohnny: @Steven-Chan 

> First thing want to confirm about `AUTH_RECORD_KEYS` and metadata. If developer configure other keys, e.g. phone, as `AUTH_RECORD_KEYS`, would that be a top level key? And if username and/or email are not in `AUTH_RECORD_KEYS`, are they part of predefined attributes in metadata?

1. I consider `AUTH_RECORD_KEYS` will represent keys  at top leavel. For instance, if `AUTH_RECORD_KEYS=[["phone"]]`, then user object is:

   ```
   {
     user_id: <id>,
     created_at: <created_at>,
     updated_at: <updated_at>,
     last_seen_at: <last_seen_at>,
     verified: <verified>,
     verify_info: <verified_info>,
     roles: [<role>, <role>, <role>, ...],
     phone: "<a_phone_number>",
     metadata: {
         avatar_url: <avatarUrl>,
         name: <name>,
         nickname: <nickname>,
         birthday: <birthday>,
         preferred_lang: <preferred_lang>,
     }
   }
   ```

2. username/email will not at pre-defined user attributes, though app still can define username/email in custom user data in metadata, they can be treated as normal user attribute, like

   ```
   {
     ...
     phone: "<a_phone_number>",
     metadata: {
         avatar_url: <avatarUrl>,
         name: <name>,
         nickname: <nickname>,
         birthday: <birthday>,
         preferred_lang: <preferred_lang>,

         // custom user data
         username: <username>,
         email: <email>
     }
   }
   ```

> Second, I think when considering what are predefined attributes, we need a real use case.
> For example, avatar_url are commonly used when link with social account and maybe chat app. However, nickname and preferred_lang seems no real use case.
> It was mentioned before that these attributes are for user portal, however, if we now support custom attributes in meta data, and the support for both predefined and custom attributes should be the same, unless the list of predefined attributes is large enough that cover different area, which is pretty impossible. So I don't think user portal is a strong supporting use case when defining predefined attributes.

I can understand your concern; current "pre-defined user attributes" design base on a hypothesize: "We will have user-management dashboard in a short time" (https://github.com/SkygearIO/features/pull/261#issuecomment-454262346). And base on this, so we

1. manage user profile in auth gear
2. pre-defined user attributes

Though I do have troble to define a large enough attribute set to cover all possible use cases.

kaojohnny:
> > 3. In current spec, I defined `preferred_lang` using RFC 5646 format, ex: 'en', 'zh-TW', 'zh-CN', 'zh-HK'. Do you think this is okay?
> 
> @kaojohnny Do you mean skygear support this as a field type, like location?

@Steven-Chan May you elaborate more about what do you mean "support"? Do we have concern on lang support? I think `preferred_lang` would be useful in user management dashboard, returned error message, push notification, ..., etc. Thanks.

Steven-Chan:
> @Steven-Chan May you elaborate more about what do you mean "support"? Do we have concern on lang support? I think `preferred_lang` would be useful in user management dashboard, returned error message, push notification, ..., etc. Thanks.

"support" means supporting `RFC 5646 format` as a field type in server and SDK, otherwise this `preferred_lang` is just a string field.

kaojohnny:
add `changeUserAuthInfoKeys` to support update username or email at 1a284dc.

kaojohnny:
> > @Steven-Chan May you elaborate more about what do you mean "support"? Do we have concern on lang support? I think `preferred_lang` would be useful in user management dashboard, returned error message, push notification, ..., etc. Thanks.
> 
> @kaojohnny "support" means supporting `RFC 5646 format` as a field type in server and SDK, otherwise this `preferred_lang` is just a string field.

@Steven-Chan Thanks. I remember there are some issues about localization support in skygear, so I believe language support is necessary. We can first confirm if the format is enough for future use, and then have another issue for further discussion for language support.

chpapa:
> 1. I do think we need support update username or email, I will update the spec to add an API: `changeAuthRecordKey({"email": "<new_email>"})`.

1. It seems a bit unnatural to call it `AuthRecordKey`? What about simply `changeEmail` or something like that which is more discoverable in the documentation? I thought `email` and `username` are user object's attribute, not metadata? So we bascially only need two API to update user's email and username?

2. Will `AuthRecordKey` support metadata? It seems we should or else it lost the meaning of the configuration.

kaojohnny:
> > 1. I do think we need support update username or email, I will update the spec to add an API: `changeAuthRecordKey({"email": "<new_email>"})`.
> 
> 1. It seems a bit unnatural to call it `AuthRecordKey`? What about simply `changeEmail` or something like that which is more discoverable in the documentation? I thought `email` and `username` are user object's attribute, not metadata? So we bascially only need two API to update user's email and username?

skygear next has a somewhat flexible user auth authentication system, and an app can choose from `username OR email`, `username AND email`, `phone_number OR email`, ... when user signup. At latest update `a284dc`, I think `AuthRecordKeys` is an internal concept so that it may be not easy to understand for other developers, so I change the API name to `changeUserAuthInfoKeys`. User can use it like 

```
changeUserAuthInfoKeys({
  email: <new email>,
  phone_number: <new phone number>
});
```

Hopefully, it easy to find.

> 2. Will `AuthRecordKey` support metadata? It seems we should or else it lost the meaning of the configuration.

Not quite understand your question. May you help elaborate a littele more detail? By default, 

```
AUTH_RECORD_KEYS=[["username"], ["email"]]
```

which means an user would sign up by "username" or "email". The value of `AUTH_RECORD_KEYS` is stored in app config.

chpapa:
@kaojohnny Please refer to the specification:
```
{
    user_id: <id>,
    created_at: <created_at>,
    updated_at: <updated_at>,
    last_seen_at: <last_seen_at>,
    verified: <verified>,
    verify_info: <verified_info>,
    roles: [<role>, <role>, <role>, ...],
    username: <username>,
    email: <email>,
    metadata: {
        avatar_url: <avatarUrl>,
        name: <name>,
        nickname: <nickname>,
        birthday: <birthday>,
        preferred_lang: <preferred_lang>,
        
        // any other custom attributes
        ...
   }
}
```

> so I change the API name to `changeUserAuthInfoKeys`

1. What is the reason we cannot simply use `changeUsername` and `changeEmail`?

I am well aware of the mechanism of the `AUTH_RECORDS_KEYS`, however, I don't see how it would be related to the user data.

BTW, what is `phone_number` in your example? Is it a metadata?

2. Will `AuthRecordKey` support metadata?

I'm literally saying it can I include a metadata field in the config of `AuthRecordKey`?  For example, can I put something like these?

```
AUTH_RECORD_KEYS=[["username"], ["metadata:nickname"]]
```

or simply

```
AUTH_RECORD_KEYS=[["username"], ["nickname"]]
```

carmenlau:
Would like to share my thoughts on the api naming.

In v1 js sdk, we have

```
signup(authData: Object, password: String, data: Object): Promise<Record>
signupWithEmail(email: String, password: String, data: Object): Promise<Record>
signupWithUsername(username: String, password: String, data: Object): Promise<Record>
```

I think we may want the change api follow the signup api, so maybe we can have

```
changeAuthData(authData: Object, password: String, data: Object): Promise<Record>
changeEmail(email: String, password: String, data: Object): Promise<Record>
changeUsername(email: String, password: String, data: Object): Promise<Record>
```

We used the word `authData` in v1 to describe the data that we used for identify user, i think we can keep it. And we can consider will we change `signup` to `signupWithAuthData` for consistent.

carmenlau:
Btw, would like to remind that we had a doc for SDK api consistency, we can check it when we design the api naming. The verb `change` is great :)

https://github.com/SkygearIO/features/blob/master/features/207-sdk-api-consistency/guidelines.md

chpapa:
@kaojohnny we just had a quick discussion with @carmenlau and @Steven-Chan , here are what we discussed; @carmenlau will have a conf call with you to sync on it:

1. At this stage, the only change is we want to move `username` and `email` to metadata, as all other user object data are those "not updatable by user" (except Password). So `username` and/or `email` are merely default we will create for user when they sign up with the `signupWithEmail` or `signupWithUsername`. 
    1. In this case, I'm not sure if we still need a special `changeAuthData` API? Can we simply ask users to update the username and email field of metadata? @carmenlau 
1. The consideration for further development are (but no need for now, so leave it as a remarks in spec?)
    1. We shall support login with different identifiers including password in future. For example, it could be `(username, password1)` and `(email, password2)` both login to the same user. It meant some API changes.
    1. We have a global `AUTH_RECORDS_KEYS` config only. Each user's identifiers are devierd implicitly by if they have some metadata filled in. For example if `AUTH_RECORDS_KEYS=[["email"], ["username", "role"]]`, and if the user doesn't have a `role` in their metadata, they can only login with (email, password). However, in the future, we may want to explicitly let users set each user's available identifiers.
    1. We may need an API to list all "identifiers" available for a user.

Let me know if I missed / didn't capture anything? @Steven-Chan  @carmenlau

carmenlau:
We also discussed about predefined fields in metadata. We may not want to have this in this stage, as in v1 we found it is difficult to enforce user to store those data in a fixed field. For example, if user had external db with the user data, and they want to use the same field name in auth gear metadata. If fixed those fields name is for the future dashboard only, many the fields information should be config of the dashboard like v1 CMS, rather than fixed them.

kaojohnny:
@chpapa @carmenlau Thanks for your comments. I leave some thoughts here first and let's find some other time for future discussion.

1. For moving `username` and `email` to `metadata`, I prefer putting them outside metadata because  
   1. `username` and `email` are frequent use attributes; I though put them at the root level is more friendly to access. I mean `user['username']` vs `user['metadata']['username']`. 
   2. `username` and `email` are like `roles`, `verified` user attributes, they need to update via specific API, like `updateUsername`, `updateEmail`, ..., etc. Unless we also support change `username` and `email` via `updateMetadata`, or it looks weird to me to put them in metadata.
 2. For removing pre-defined user attributes:  
    1. From the previous discussion, I rather think there is no need to config user management dashboard, it is more like a built-in feature for auth gear user, and the configuration could be complex to developers.
    2.  If an app saves user profile in external DB only,  don't know if it easy to config or access remote user attributes in the user management dashboard.

chpapa:
@kaojohnny  In responds to your comments:

1. I agree with the first argument that `user['username']` is better than `user['metadata']['username']`, however I don't agree with 2nd arguments, we can just support updating via `updateMetadata()` and don't need another API for that. I think the problem is whether it is weirder to make User object's fields dynamic (default it have username and email fields, but it could have phone no, or even role or other special attributes), or better if we group all dynamic and user-defined attributes inside metadata.
2. i. We could have "default user attribute name", for example, our Push Notification Dashboard could have a default config of using `preferred_lang` for which language of push to send, or our Auth UI Kit can default to use `avatar_url` for an avatar. And we teach users this is the default we use, so that we don't need any concept of "predefined" attributes.
ii. Don't know why it is relevant, we assume they must save those data in metadata in-order to be accessible by others.

kaojohnny:
@chpapa 

1. Using `updateMetadata` to change auth info key (username, email, phone number, ...) would generate some corner cases. Like:

```javascript
// user signup with username
skygear.auth.signupWithUsername('user', 'password');

// what if user add email in updateMetadata
// does skygear support the user to login with email after the operation?
skygear.auth.updateMetadata({
    username: 'user',
    email: 'user@example.com'
});

// user signup with username and password
skygear.auth.signup({
  'username': 'user',
  'email': 'user@example.com' 
});

// since metadata is free form object, what if user delete email in updateMetadata
// does skygear still support the user to login with email after the operation?
skygear.auth.updateMetadata({
    username: 'new_username',
});
```

We can use error message or documentation to illustrate such cases, though I think this API usage is not intuitive to me. 

May I suggest another way? How about we separate "dynamic" part into two parts? Like

```
{
    user_id: <id>,
    created_at: <created_at>,
    updated_at: <updated_at>,
    last_seen_at: <last_seen_at>,
    verified: <verified>,
    verify_info: <verified_info>,
    roles: [<role>, <role>, <role>, ...],
    authdata {
      username: <username>,
      email: <email>,
    },
    appdata: {
        avatar_url: <avatarUrl>,
        name: <name>,
        nickname: <nickname>,
        birthday: <birthday>,
        preferred_lang: <preferred_lang>,
        
        // any other custom attributes
        ...
    }
}
```

And use `changeAuthData`, `changeEmail` and `changeUsername` to change user authentication ID, and use `updateAppdata` to update user custom attributes.

2. In the spec, I define user metadata something like:

```
CREATE TABLE _auth_user_metadata (
  user_id text REFERENCES _core_user(id),
  avatar_url text,
  name text,
  nickname text,
  birthday timestamp without time zone,
  preferred_lang text
  data jsonb, /* custom user attributes */
  PRIMARY KEY(user_id),
  UNIQUE (user_id)
);
```

The `data` field is a JSONB object which stores all custom user attributes (except fixed ones). If we don't use fixed fields (`avatar_url`, `name`, ...), and store all custom user attributes in a JSONB object, I consider it would be difficult to implement some basic user management dashboard requirements, such as query and listing. I can understand "pre-defined attribute name" can be used in the Auth Kit, or the push notification dashboard, I just want to point out it has some limitation too.

chpapa:
@kaojohnny 

1. I don't see both of the cases you mentioned are edge cases. The first case maybe confusing because our default configuration is both `(username, password)` and `(email, password)` are identifiers to log in the system. So implicitly when user set the email, `(email, password)` is allowed. We can solve it by making the configuration explicit. Separate metadata into authdata won't solve the confusion case above. For the second case, it should be OBVIOUS that removing email would stop the users from logging in as `(email, password)` is the identification pair. It is only confusing to you because you know the implementation details that the data were copied elsewhere.

I have no preference over moving the authdata away, I actually suggested it before, but if we move it I think it is even better if we make it explicit which data are paired together for identification (example below), and I think we should avoid using the term `appdata` as it sounds too much like user profile and mislead people into using it as such.

```
{
    user_id: <id>,
    created_at: <created_at>,
    updated_at: <updated_at>,
    last_seen_at: <last_seen_at>,
    verified: <verified>,
    verify_info: <verified_info>,
    roles: [<role>, <role>, <role>, ...],
    authdata {
      [
          {username: <username>, password: <password>},
          {email: <username>, password: <password>},
          {type: <type>, username: <username>, password: <password>}
      ]
    },
    appdata: {
        avatar_url: <avatarUrl>,
        name: <name>,
        nickname: <nickname>,
        birthday: <birthday>,
        preferred_lang: <preferred_lang>,
        
        // any other custom attributes
        ...
    }
}
```

2. That is a classic example of mixing specification design with implementation. I think for users they don't care (and will never know) how the underly implementation was done. It could be elasticsearch, mongodb, or postgresql jsonb.

BTW, for your limitation. I think PostgreSQL support at least GIN index already. Quick google (https://bitnine.net/blog-postgresql/postgresql-internals-jsonb-type-and-its-indexes/)

kaojohnny:
@chpapa  Here I list some use cases, hope it helps.

Let's suppose user signup with username

```
skygear.auth.signupWithUsername('user', 'password');
```

And we have 

|  | authData in root user object | authData under `user.metadata` | authData under `user.authData` |
| -------- | -------- | -------- | -------- |
| access username | `user['username']` | `user['metadata']['username']` | `user['authData']['username']` |
| change username | `changeUsername(...)` | `changeUsername(...)`<br/>`updateMetadata(...)` | `changeUsername(...)`<br/>`updateAuthdata(...)` |
| add login with email provider | N/A¹ | `updateMetadata(...)` |N/A¹ |
| remove login with username provider | N/A¹ | `updateMetadata(...)`² |N/A¹ |
| update metadata | `updateMetadata(...)` | `updateMetadata(...)` | `updateMetadata(...)` |

¹: We don't have such API yet.
²: User couldn't delete all authData

Merging username, email into `metadata` makes `updateMedata(...)` become a way too powerful API, it has implicit functionalities that didn't match the API's name.

For link current user with second auth provider (email, phone number, ...), I prefer a explicit API to do it. (ex `linkUserWithEmail(...)`, `linkUserWithUsername(...)`)

Personally I don't care "dynamical" data attributes in root user object, as a developer, I think it seldom changes, and can understand it may need a distinct API to change it.

Feel free to give any comments. Many thanks.

chpapa:
@kaojohnny Here is what I think:

1. IMO I don't think combine or separate into `metadata` or `authdata` would have any differences. Maybe you should list out all cases (including using  `(type, username)` as login ID), you will find that both implementations actually could use similar APIs (e.g. `linkUserWithUsername` etc?), and have similar implictness when we, for example what if `updateAuthData` suddenly added `email` when the user object only have `username` before? It will implict linked the user with the email.
1. To further elaborate my suggesiton above, if we truely want to separate auth data from metadata, below are what I think we should do instead. (Slightly modified as I finally have time to write things on desktop instead of mobile now lol)

```
{
    user_id: <id>,
    created_at: <created_at>,
    updated_at: <updated_at>,
    last_seen_at: <last_seen_at>,
    verified: <verified>,
    verify_info: <verified_info>,
    roles: [<role>, <role>, <role>, ...],
    loginIDs { // I think loginID is better than authdata? 
      {
          "username": <username>, // default
          "email": <email>, // default
          "nickname": <nickname>, // user-defined
          "type&username": {"type": <type>, "username": <username>}, // idea 1 for complex type of loginIDs, multiple fields in dict
          // "type&username": <type-username>, // idea 2, we only support string, and user need to multiplex multiple fields into one string on their own
          "crazy": { "crazyID": <crazyID>, "password": <password> } // ok I'm just suggesting, very niche feature that different ID need different password.
      }
    },
    appdata: {
        avatar_url: <avatarUrl>,
        name: <name>,
        nickname: <nickname>,
        birthday: <birthday>,
        preferred_lang: <preferred_lang>,
        
        // any other custom attributes
        ...
    }
}
```

| | sample code |
| --- | --- |
| access username | `user['loginIDs']['username']` |
| change username | `updateLoginID('username', "[username]")`, or `changeUsername("[username]")` |
| add login with email provider | `updateLoginID('email', "[email]")` |
| remove login with username provider | `deleteLoginID('email')` |
| add login with type + username | `updateLoginID('type&username', {"type": "[type]", "username": ["username"]})` |
| the crazy idea of supporting specific password of LoginIDs | `updateLoginIDspecificPassword('email', "[password]")` |

Besides the data structure, we can also consider the following in the design...
* Simply remove the `AUTH_RECORD_KEYS` config. We can still keep it if it helps...
* Consider if the term of loginID is clear enough for APIs, that we can remove shortcut API like `changeUsername`.
* We definitely don't need `linkUserWithUsername/Email` again. The data structure is clear enough that we support multiple LoginIDs

Is it way easier to understands? Maybe try to ask developers who are new to Skygear, tell them we support multiple login IDs for user, and see if they understand the API and usage?

kaojohnny:
@chpapa While I am writing use cases, I found a case would like to discuss first.

Suppose a app has `AUTH_RECORD_KEYS=[["username", "email"], "username"]` setting, and based on current implementation, when user signup with `username` AND `email`, auth gear will create two `loginID`. They are:

```
loginIDs: [
  {username: "username", email: "email@example.com"},
  {username: "username"}`
]
```

And then the user can log in by 1. username 2. username AND email.  The question is, when we update `LoginID`, **do we expect they updated separately?** For example, when user invokes `updateLoginID('email&username', {email: 'email@example.com', username: 'new_username'})`, will it update `loginIDs` as:

```
// only one username updated
loginIDs: [
  {username: "new_username", email: "email@example.com"},
  {username: "username"}`
]
```

If yes, I thought we have to choose the user object structure as:

```
{
    user_id: <id>,
    ...
    loginIDs: {
      "email&username": {username: "new_username", email: "email@example.com"},
      "username": "username"
    },
    appdata: {
        ...
    }
}
```

This is the only solution that can illustrate a user connected to two different usernames. Not sure if this is what we want (two different usernames of a user), would you help provide your insights? I think it is a flexible solution, but also a little complex to the skygear users.

chpapa:
@kaojohnny I think to keep things simple, we should expect them to update the username separately in your case. IMO conceptually it is easier to explain we support multiple login IDs, so if it happens two different loginIDs both have a field named `username`, they are of course different.

Have we thought about is it better to remove `AUTH_RECORD_KEYS`? I'm thinking the only reason we want to keep such a config is for graphical user interface (maybe it is a good enough reason though)

kaojohnny:
@chpapa I have updated this spec to align discussion we had yesterday, when you have time, please help review. Thanks.

features/253-auth-sdk-callback-in-next/api.md
```diff
@@ -0,0 +1,324 @@
...
+- `onUserChanged`
+
+Those requirements can be implemented via cloud function or external user DB.
+
+## API argument naming chages on Client JS SDK:
+
+- `public async requestVerification(recordKey: String): Promise`  
+  `public async requestVerification(loginIDKeys: [String]): Promise`
+- `async login(authData: Object, password: String): Promise<Record>`  
+  `async login(loginIDs: Object, password: String): Promise<User>`
```
chpapa: Shall we have an API for login with LoginID without specifiying which field to match? Something like:
`async loginWithLoginID(loginID: String, password: String)`?

kaojohnny: The usage of `async login(loginIDs: Object, password: String): Promise<User>` is
```
skygear.auth.login({username: "exmaple"}, "123456");
```
auth gear will match `{username: "example"}` with existing principals to authenticate the user and the password.

In your suggestion API `async loginWithLoginID(loginID: String, password: String)`? Do you expect user login by `loginWithLoginID("example", "123456")`? User doesn't need to specifify  `username` here.

chpapa: > In your suggestion API `async loginWithLoginID(loginID: String, password: String)`? Do you expect user login by `loginWithLoginID("example", "123456")`? User doesn't need to specifify `username` here.

yes, but maybe just ignore me here first lol

kaojohnny: Thanks, I ignore this one temporarily.

kaojohnny:
@chpapa Updated.

kaojohnny:
If no one rejects, I will merge this PR later today. Thanks.

Issue #256, opened 2018-11-26
Skygear Auth 2.0


kaojohnny:
# Description

This feature issue is about porting Auth to the new Skygear Next platform.

# Open Questions

- In skygear next, how to update a user's auth info and user profile? 
- In skygear next, how does current user access his own auth info, e.x. last login at, last seen at, disabled state, roles, etc.
- In skygear next, how to access other user's auth info and user profile?

# Related Issues

- Server Issues
SkygearIO/skygear-server#878
SkygearIO/skygear-server#880
SkygearIO/skygear-server#881
SkygearIO/skygear-server#882
SkygearIO/skygear-server#883
SkygearIO/skygear-server#887
SkygearIO/skygear-server#874
SkygearIO/skygear-server#835
SkygearIO/skygear-server#818
SkygearIO/skygear-server#820
SkygearIO/skygear-server#876
SkygearIO/skygear-server#877
SkygearIO/skygear-server#879
- Client Issues
- Guides Issues

kaojohnny:
https://oursky.quip.com/ewfGAfyodkb0

chpapa:
Note from Meeting @kaojohnny 

We will properly need to re-design it since we removed Record Gear in the release, the design shall aim at providing one-way (or even two-way?) sync for auth data, and decide the following:

1. Are we going to provide one-way or two-way sync for auth data?
2. What are the auth data we will need to sync for user's access and the format? (Presume it is SQL first...)
3. Consider the design for syncing auth data to single (and multiple DBs) in future, that might include SQL or non-SQL...

chpapa:
Note from Meeting for new scope:

We will probably need to re-design the whole experience of how auth is used since we removed Record Gear in the release.

There will be two (three?) main ways for users to interact with auth:

1. Client SDK
2. (optional?) SDK for Cloud Functions / gRPC or api call for Cloud Functions
3. Access database sync by auth gear to read (or write?) data

Major considerations:
1. List out possible use cases regarding current auth session / manipulating other users' auth data
2. Determine we will provide one-way or two-way sync for auth data and how? (e.g. One-way sync meant users can "read" all user's disabled status from CF, two-way sync meant users can "read" and "change" all other user's disabled status from CF.)
3. What are the auth data (DB schema) and APIs good enough to satisfy use cases in 1
4. Consider the one-way/two-way sync need to support multiple DB vendors in future.

chpapa:
[...]

Issue #258, opened 2018-11-29
Support User Language Preference in standard user profile


chpapa:
# Description

It is a pretty common requirement in applications, and skygear could help by having a standard attribute.

It is expected to have:
- Each application have a list of languages that are allowed to be selected
- Each application have a fallback/default language if not selected
- APIs aware to return data in preferred language (maybe a API on-top/independent from record API)
- Related features in CMS #229 
- Configurable template for forgot password/welcome email/sms, etc.
- Segment for Push Notifications

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
@kaojohnny another reason maybe some metadata structure in Auth would be good for NEXT?

PR #261, opened 2018-12-03
next: auth gear interaction overview

kaojohnny:
connects #256


features/256-design-for-user-attribute-operations-in-next/api.md
```diff
@@ -0,0 +1,178 @@
...
+- `verified`
+- `xxx_verified` (`xxx_verified` would change when server configuration is changed.)
+
+## If record gear is configured
+
+If record gear is configured, predefined user attributes will be saved as columns of `_record__user` table, and a developer can use a rich API set which is provided by record gear to access user record.
+
+## If record gear is not configured
+
+If record gear is not configured, predefined user attributes will be saved as columns of `_auth_user_profile`. Auth gear will provide limited API to support user attributes access operations (query will not be provided).
```
rickmak: Will ACL provided?

And will their listing of user supported in SDK?

kaojohnny: 1. Default `record` attributes should be included, like "_access", "_created_at", "_updated_at", ...
2. As discuss before, auth gear won't support listing.

rickmak: I mean are we going to support ACL if recrod grear is not configured. I think the answer is no. And please make it explicit no in the feature spec.

2. Yes, we don't support listing. Than we should explicit state that in the spec.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Goal
+
+* Keep auth gear simple, focus on auth related functionalities.
+* Keep client SDK simple, less interaction with auth gear directly.
+* More cloud functions, a developer should create corresponding cloud funtion to fulfill application logic.
+
+## Major changes
+
+* Remove admin API in client SDK.
```
chpapa: More clear should be: Move admin related features from Client SDK to APIs at Cloud Functions?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+* Remove user profile support in auth gear.
+
+## Architecture overview
+
+![](https://i.imgur.com/GQbup7u.png)
```
chpapa: I think the "cloud function" user table is confusing. I don't think the user table in Cloud Functions is related with Skygear at all? It is just a Database for user profile and is optional and entirely controlled by users? In that case we can ignore it in our graph and just make it clear that Cloud Functions user can have their own DB.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+
+* should support to indicate it is restricted to be invoked by authenticated user only.
+* should support to have a table with user_id as foreign key.
+* should support to access cloud funtion user's attribute in context.
+* should support to set a user's role, disabled state.
+* should support to query user.
+
+## Auth related information
+
+Auth gear would focus on auth related information only, they are
```
chpapa: I think this list is not complete? Make it clear it is not complete is better... for example we have password history which I guess we save it in our auth DB too.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+Disabled        bool
+DisabledMessage string
+DisabledExpiry  string
+Roles           []string
+Verified        bool
+```
+
+## How to update user attributes in client SDK
+
+Since client SDK _**won't**_ have API support to update user attributes directly, developer should create their own cloud function for client SDK to update user attributes accordingly.
```
chpapa: I think this statement is misleading. `forgotPassword` change attributes, `login` could change attributes too (it change `lastLoginAt`)

I think what you meant is Client SDK won't have API to change attributes that require permissions checking? Or how to define it?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+- `adminResetPassword`
+- `assignUserRole`
+- `fetchUserRole`
+- `revokeUserRole`
+- `setAdminRole`
+- `setDefaultRole`
+
+## How to update user attributes in clound function
+
+For restricted attributes, cloud function should use corresponding interface in auth gear to update.
```
chpapa: I think this list is not completed too? For example Enable Users, Reset Password, etc..

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+function welcome(req, res) {
+    res.end("welcome " + req.context.user.email);
+}
+
+module.exports = skygear.auth.authed(welcome);
+```
+
+## How support to user referenced table in cloud function
+
+Auth gear should have auth related hooks to cloud function, so cloud function should have enough information to create user referenced table. Following is a possible example:
```
chpapa: This one need to think carefully...

I think there are two types of hooks:

1. Hooks that's enough for users to implement their own auth DB, basically like [the one at auth0](https://auth0.com/docs/connections/database/custom-db/create-db-connection#step-2-create-database-action-scripts)
2. ONE simple hooks that just call and pass all attributes of the user whenever it is updated (basically like the one-way sync concept, except we don't write it to DB, but pass it as a JSON object to a Cloud Functions, something like that)

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,126 @@
...
+## Goal
+
+* Keep auth gear simple, focus on auth related functionalities.
+* Keep client SDK simple, less interaction with auth gear directly.
+* More cloud functions, a developer should create corresponding cloud funtion to fulfill application logic.
+
+## Major changes
+
+* Remove admin API in client SDK.
+* Remove user profile support in auth gear.
```
chpapa: I'm thinking, maybe we should consider our use case and becareful if we want to support some very basic "user metadata" instead.

One example is, say when we support the feature of JWT provider, users may want to add some metadata into the claims of JWT tokens, if we have no way for users to define some metadata, there are no easy way for them to include data in their DB into those claims.

kaojohnny: @chpapa Thanks for your feedback, though I don't think we should consider such complex use case at this moment, in my opinion, even auth gear won't support "user metadata", a developer still can use other JWT provider (not skygear) to meet the requirement.

chpapa: What do you mean the user can still use other JWT provider to meet the requirement? 

For example, if a user has used Skygear Auth to kick start a new app, and later need to use Skygear Auth to integrate with chat, say Layer, which requires First Name, Last Name, Display Name, and Avatar in the JWT claims, how do they do that (without re-implementing everything?)

kaojohnny: For this case, I don't think developer has to re-implement everything.

Let's think `CF` has `user_profile` table in DB intially, a developer could implement CF to satisfy layer authentication flow easily. According to https://docs.layer.com/reference/client_api/authentication.out, cloud function should have enough information (first name, last name, display name) to generate JWT for layer (`/identity` endpoint in layer's doc) to get session token. 

And in this way, skygear can even create a set of cloud functions to support layer authentication, developer can import layer cloud function set as an plugin.

chpapa: Hmmm, to me it basically meant the developer need to reimplement everything (implement their JWT provider) while Skygear have an auth server.

While Skygear Auth as a JWT provider should be on the roadmap.

chpapa: While we might don't need it in v1, we definitely should make sure the design can accomodate those features on the roadmap.

(It reminds me we may want to take some time to re-organize the feature repo issues to remove those irrelevant in NEXT already. cc: @tensiuyan )

chpapa:
Thanks @kaojohnny !

Do you think we can list out all of the hooks that Auth need to have? And integrate that list of hooks with different use case?

I think (in very general term), some use cases are:
* return profile in specific format (like the avater in a hypothtical Auth UIKit)
* sign up with profile use case was well understood in the meeting but not in this design spec
* sign up / login with OAuth such as Google / Facebook Connects, and how we populate the profile via update_profile hooks
* pre-registion / pre-login: e.g. block login or signup by some special criteria
* post-registration / post-login?
* JWT provider

Just a quick braindump, I think there will be more. The above use cases maybe able to consolidate to a small number of hooks. I think can reference other similar services to speed up the design?

Feel free to point out if I missed anything or if there is a better approach to the problem. Thanks!

chpapa:
@kaojohnny here is feedback:

1. By your definition, please consider Auth gear will be a "rich-feature" on the roadmap. I think Auth UIKit and User Management Dashboard are very likely to be needed in short term (3 - 6 months) (and also devices management, 2FA, SAML support etc which is already in needs and suggested many times in v1); Custom DB Connect, JWT provider will probably depend on user feedback after we launched the new version.

2. Got it. For ref, auth0 doc have a "best practices" for that: https://auth0.com/docs/best-practices/user-data-storage-best-practices

3. Can we (like v1) assume hooks could be both async / sync? I think it depends.... for example, pre-registration/pre-login hook is probably blocking, but some post-registration functions may want to run in async.

Looking forward to further update on the design/discussion based on the information above.

BTW, I will try to spend some time to re-organize feature repo issues today to help make it more clear on what features are likely on the roadmap (based on current product knowledge) with @tensiuyan

kaojohnny:
@chpapa I made some updates, as usual, welcome to comments.

Some points:
1. I think one `/after_save_user` attached Cloud Function would fulfill the data consistency requirement, and in the function, a developer would reveal advanced information in `request` object (such as request path, request path, and request body). For the purpose "data consistency", and such Cloud Function should execute in a synchronized way and an asynchronized way is not preferred. 
```
// 1. req.context.user: the user object from auth gear
const user = req.context.user;
// 2. req.path: the original path from the request
const path = req.path;
// 3. req.headers: the original headers from the request
const headers = req.headers();
// 4. req.body: the original payload from the request
const body = req.body.json();
``` 
2. To consider auth gear and Cloud Function as two microservices at equal privilege in skygear, API Gateway should take the responsibility to authenticate/authorize the request, and add current execute user to request context. Both auth gear and Cloud function would able to access the current user in the request context.
3. For support rich-featured service (like User Management, JWT provider), auth gear should have a query user interface.
4. For other attached Cloud Function (`/before_login`, `/after_login`, ...), I think they add unnecessary complexity to auth gear flow, and should consider in custom database connection manner, which means instead of  (`/before_login`, `/after_login`, ...) we should consider have (`/login`, `/signup` cloud function).
5. To provide rich-featured auth gear, such as User Management Dashbord (consider user can login to the dashboard and update his own profile), I think auth gear's user profile should not be a free form (a JSON object), but should have some fixed fields (like avatar, display name, ...). Though this topic may not quite satisfy the topic of this spec, maybe have another issue to discuss?

chpapa:
@kaojohnny in respond to your comment:

1. Can you also consider the use case of UIKit, and also embed language preference / user segment for push notifcation (#258 and #262), to see if your architecture without a user metadata is do-able / what is the user experience?
2. I don't think it is necessary to consider Auth and Cloud Functions as equal (what is the reason actually?) 
3. Yes I think it is needed.
4. I don't really get it, what if the users don't want to implement their own auth DB but want to add a logic to block signup or login? How can they do it with `/login`?
5. It sounds possible and I have also thought about it, if we don't need to consider requirements like those in 1.

kaojohnny:
@chpapa @Steven-Chan @carmenlau 

I add an overview.slde file to give a glance of the design, I would like share uncertainties:

1. Can we just use `before_update_user_sync` and `after_update_user` to replace all other hooks?
2. If we do want every hook for every auth action, should we provide both "SYNC" and "ASYNC" version for both before and after hook, like:
`before_signup_sync`, `before_signup`, `after_signup_sync`, `after_signup`?
3. In current design, hook function won't have a way to alter user attributes, is it enough to use?
4. Since CF may have its own version of user object, how to make sure `currentUser` in SDK have a normalized user data? (use case: CF has some extra user attributes besides user profile).
5. Besides to support query interface in auth gear, for future User Management Dashboard, should auth gear also support update interface? (use case: Admin wants to update user disabled state via User Management Dashboard).

chpapa:
@kaojohnny thanks a lot, I read the slide and have some questions:

* On slide 11, what is the meaning of the json-like user struct? Does it meant our SDK / API will try to "merge" auth data, metaData and profile whenever users request?


> 1. Can we just use `before_update_user_sync` and `after_update_user` to replace all other hooks?

* I'm not sure what it meant, why it can replace all other hooks (such as `after_logout`)?



> 2. If we do want every hook for every auth action, should we provide both "SYNC" and "ASYNC" version for both before and after hook, like:

Maybe, we could only know after we list out the use case? (or play safe to provide all first...)



> 3\. In current design, hook function won't have a way to alter user attributes, is it enough to use?

Again, we could only answer this question as we list out use case, what is your research so far? BTW, what is user attributes? Is it auth data, metadata or profile? (Speaking of which, maybe we need to define the terminology and stick with it consistently to what are they in the document / first slide?)



> 5\. Besides to support query interface in auth gear, for future User Management Dashboard, should auth gear also support update interface? (use case: Admin wants to update user disabled state via User Management Dashboard).

I think it depends, if the User Management Dashboard is part of Auth Gear, maybe we can allow it to call internal functions / directly access the DB.

On the other hand, if it is part of CMS, an interface is better.

I know it is related to business requirement but I honestly have no idea now...

> 4\. Since CF may have its own version of user object, how to make sure `currentUser` in SDK have a normalized user data? (use case: CF has some extra user attributes besides user profile).

Sorry I don't get it.

kaojohnny:
@chpapa 

Let me clear terminologies I used in the doc:
- auth data: auth related state, such as disabled, last login at, ..., etc.
- user metadata: indicated as fixed user properties, such as avatar, first name, last name, display name, ..., etc. 
- user profile: varied user properties, differs from application to application
- user attributes: merge user auth data, user metadata, user profile together.

1. In normal usage, I am thinking `before_update_user_sync` and `after_update_user` can be used to describe all auth actions (login, logout, disabled),  so a developer can use `before_update_user_sync` or `after_update_user` to fulfill requirements in most cases. I don't have many use cases for this, it is just a wild guessing.
2. I will figure out if current design is enough by checking more use cases.
3. Same as the previous one.
4. Let's check later.
5. Let's auth gear saves following user attributes:
```
{
    metaData: {
        avatarUrl: "http://avatar.com/1.png",
        ...
    },
}
``` 
And CF saves different user avatar URL `http://avatar.com/2.png` for some reason. I think when user login, it should get respose
```
{
    metaData: {
        avatarUrl: "http://avatar.com/2.png",
        ...
    },
}
```

chpapa:
> 1. In normal usage, I am thinking `before_update_user_sync` and `after_update_user` can be used to describe all auth actions (login, logout, disabled),  so a developer can use `before_update_user_sync` or `after_update_user` to fulfill requirements in most cases. I don't have many use cases for this, it is just a wild guessing.

Hmmm, I think it is basically asking if we shall multiplex multiple hooks into one? You will have something like an "action" attribute if you put all hooks into one?

So the consideration is if we will have multiple hook point for login/logout/disabled, or one hook point with an "action" attribute of "login/logout/disabled"?

Gut feeling is an explicit hook interface better than a variable in context, but I'm not sure...


> 5\. Let's auth gear saves following user attributes:

Sounds make sense...?

chpapa:
> so I suggest we have a separate spec specifically for web hook

Agree.

BTW, I'm thinking, for hook design, maybe we want to have two specs (could start with a general one first...)

1. A very general, NOT Skygear specific way to invoke functions, such as general HTTP hook, or something like public OpenAPI / gRPC call.
1. A Skygear specific way to invoke functions that provide extra benefits on security, usability and latency. (e.g. use better transport / more efficient cross-language packing data format, or simply gRPC or alike)

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,382 @@
...
+| `logout` | `before_logout_sync(user, orgUser, currentUser)`<br/>`before_logout(user, orgUser, currentUser)`<br/>`after_logout_sync(user, currentUser)`<br/>`after_logout(user, currentUser)` |
+| `password` | `before_change_password_sync(user, orgUser, currentUser)`<br/>`before_change_password(user, orgUser, currentUser)`<br/>`after_change_password_sync(user, currentUser)`<br/>`after_change_password(user, currentUser)` |
+| `pasword` | `before_reset_password_sync(user, orgUser, currentUser)`<br/>`before_reset_password(user, orgUser, currentUser)`<br/>`after_reset_password_sync(user, currentUser)`<br/>`after_reset_password(user, currentUser)` |
+| `verify` | `before_verified_sync(user, orgUser, currentUser)`<br/>`before_verified(user, orgUser, currentUser)`<br/>`after_verified_sync(user, currentUser)`<br/>`after_verified(user, currentUser)` |
+| `update_user` | `before_update_user_sync(user, orgUser, currentUser)`<br/>`before_update_user(user, orgUser, currentUser)`<br/>`after_update_user_sync(user, currentUser)`<br/>`after_update_user(user, currentUser)` |
+
```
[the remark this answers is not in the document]
kaojohnny: @Steven-Chan 

1. In my opinion, "hook" reflects API invoking action rather than data field changes, so I don't think we need "un-verifying" hook (we don't have such API to "un-verifying" user). 
2. Same as the previous one, I think `reset_password` and `change_password` may act similarly,  they are different when they are invoked. Admin may want to trace them separately.

And follow the thought, I think `update_user` is only invoked when user metaData and profile updated, which implies auth gear needs to provide a new API `update_user` to support update user metaData and profile.

If we do want a trigger reflects data field changes (like v1 skygear record hook?), maybe we should define hook like: `change_verified`, `change_password`, `change_disabled`, ..., etc.  Want to know more about your opinions. Thanks.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,395 @@
...
+```javascript
+const skygear = require('skygear');
+
+/* 
+ * user: user object to be saved
+ * orgUser: original user object
+ * context: current exection context
+ */
+function before_XXX_sync(user, orgUser, context) {
```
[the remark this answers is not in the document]
kaojohnny: How about `original_user`?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,395 @@
...
+
+\* this diagram is an overview only, detail flow explain in later section.
+
+- Cloud Functions may have its own DB.
+- It's developer's responsibility to create corresponding Cloud Function to handle admin related auth features (enable/disable user, assign role, ...).
+- Cloud Function would implement hook functions to handle user changes.
+
+## Move auth admin APIs from SDK to Cloud Function
+
+Following function will be removed from SDK, a developer should create its Cloud Function to handle admin tasks.
```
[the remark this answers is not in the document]
kaojohnny: I think in Cloud Function, such admin API should be invoked with the master key.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,395 @@
+# auth gear and Cloud Function interaction overview in skygear next
+
+Base on the new product architecture decision (auth gear + Cloud Function and drop record gear), skygear next should bring up a new experience on how auth gear is used in Cloud Function and Client SDK.
+
+## Glossary
+
+* auth gear: a skygear provided component which utilizes user authentication and authorization process, and user profile handling. It would bring enhanced features in the future, such as: JWT provider, Auth UIKit, user management dashboard.
+* Cloud Function: a developer would create a Cloud Function to fulfill application requirements. A cloud function should be a single purpose that attached to certain events or triggered by requirement.
+* auth data: auth related state, such as disabled, last login at, ..., etc.
+* user metaData: indicated as common user properties, such as avatar, first name, last name, display name, ..., etc.
```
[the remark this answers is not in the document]
kaojohnny: I would like to keep "auth data" and "user metaData" idea separately, so I replace the term "user metaData" with "user common properties" because their behavior are different fundamentally.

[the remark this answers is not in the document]
chpapa: @kaojohnny how are they different? For example, how `disabled` different from `avater` fundamentally, and what are you trying to communicate with the user / other skygear developers when you have two terms/category for them? Asking coz I don't see any differences.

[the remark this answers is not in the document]
kaojohnny: @chpapa  `disabled` is related to auth state, which is only changed by specific API (`/auth/disable/set`). 

For `avatar`, `display_name`, `first_name`, they can be changed in a batch (`update_user`, the API doesn't exist yet), and they are not related to auth state. A developer can change them freely.

And second, auth gear allows a hook to change user common attributes and user profile in before hooks, but not support to change auth data in all hooks. Which means, a developer can't simply write `user.disabled = true;` in CF to disable a user. Such action must go through public auth gear API.

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
chpapa: sounds better

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
chpapa: I think we need a better definition of auth data. Anyway, I will comment on the whole doc again for a better bigger picture to confirm this document.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,402 @@
...
+
+```javascript=
+const skygear = require('skygear');
+
+module.exports = (req, res) => {
+    res.end("welcome " + req.context.user.email);
+}
+```
+
+## Use cases: save user profile in Cloud Function DB after user signup
```
chpapa: I think we should write a short discussion about why this use case is important here. Maybe something like these?

"We decided the recommended use cases, is user of Skygear Auth shall use `after_user_profile_and_metadata_updated` to create and sync a user profile table in their own database if they need it, below is an example of how it should be done"

chpapa: BTW, the hook name is fake... 

Moreover, maybe we should provide / merge the sample of "create" and update? and also show that we will have an API to let user update the metadata and profile in ONE example?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Move auth admin APIs from SDK to Cloud Function
+
+Following function will be removed from SDK, a developer should create its Cloud Function to handle admin tasks.
+
+| Attribute | Interface | Old JS SDK API |
+| -------- | -------- | ------ |
+| disabled | `POST /auth/disable/set`| `adminDisableUser`<br/>`adminEnableUser` |
+| password | `POST /auth/reset_password`| `adminResetPassword` |
+| role | `POST /auth/role/assign`<br/>`POST /auth/role/admin`<br/>`POST /auth/role/default` | `assignUserRole`<br/>`revokeUserRole`<br/>`setDefaultRole` |
```
chpapa: Shall we add the APIs for update user profile and metadata too?

kaojohnny: I don't think so, `update_user_meta_data_and_profile`(temp name) API is not admin API, it will need client SDK support. This section lists API supports **removed** from client SDK (JS, iOS, Android).

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,402 @@
...
+
+// DB operation
+err = userStore.update(user)
+if err != nil {
+    response.Err = skyerr.MakeError(err)
+    h.TxContext.RollbackTx()
+    return response
+}
+
+err = hooks.ExecuteAfterHooks(user)
```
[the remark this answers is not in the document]
kaojohnny: Both `sync` an `async` supports to raise an exception to stop operation, this behavior follows skygear v1's design.

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
kaojohnny: Take `last_login_at` as an example, a developer should consider creating `after_login` hook to handle requirements.

The reason for keeping `update_user_metadata_and_profile`(temp name) not triggering when auth data changed is that I want to avoid multiple hooks trigger for different reasons.

Take login as an example (`last_login_at` changed), following hooks would trigger:
`before_login_sync`
`before_update_user_sync`
`before_login`
`before_update_user`
`after_login_sync`
`after_update_user_sync`
`after_login`
`after_update_user`

A developer could feel confused where to implement application logic.

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
chpapa: Interesting. Maybe need to reconsider? I think async but able to revert a transaction is very weird. If it is in a transaction, that is sync, not async by definition.

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
chpapa: Consider the major use case, if `last_login_at` or `created-at` is in profile DB of user, the user need to distribute code related with his profile DB in many hooks.

That seems even more confusing than if `update_user` combine all user object attribute update.

I don't really think having `update_user` hook instead of `update_user_metadata_and_profile` hook is more confusing.

features/256-auth-gear-interaction-overview-in-next/api.md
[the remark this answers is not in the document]
kaojohnny: @chpapa  Sorry, I am wrong, async won't cause rollback. 

`ExecuteBeforeHooks` or `ExecuteAfterHooks` would execute all hooks (async, sync) one by one. For sync hook, the loop will be halt and wait for the response (with timeout), for the async hook, it just creates a go routine and continues to next hook.

kaojohnny:
Note for another change of user object structure:

old:
```
{
    id: <id>,
    createdAt: <createdAt>,
    updatedAt: <updatedAt>,
    disabled: <disabled>,
    roles: [<role>, <role>, <role>, ...],
    avatarUrl: <avatarUrl>,
    birthday: <birthday>,
    preferredLang: <preferredLang>,
    // more common user attributes
    ...
    profile: {
        // any other free form data
        ...
    }
}
```

new:
```
{
    id: <id>,
    createdAt: <createdAt>,
    updatedAt: <updatedAt>,
    disabled: <disabled>,
    roles: [<role>, <role>, <role>, ...],
    metadata: {
        avatarUrl: <avatarUrl>,
        birthday: <birthday>,
        preferredLang: <preferredLang>,
        // any other user common attributes
        ...
        // any other arbitrary form attributes
        ...
    }
}
```

kaojohnny:
update: auth hooks naming

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -93,7 +95,9 @@ if err != nil {
     return response
 }
 
-err = hooks.ExecuteAfterHooks(user)
+hooks.ExecuteAfterAsyncHooks(user)
+
+err = hooks.ExecuteAfterSyncHooks(user)
```
chpapa: hooks.ExecuteAfterSyncHooks should goes BEFORE Async hooks?

Because if the Sync Hooks is failed, Async hooks may not want to be triggered.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -172,25 +176,38 @@ Followings are hooks of auth actions:
...
-| `verify` | `before_verify_sync(user, context)`<br/>`before_verify(user, context)`<br/>`after_verify_sync(user, context)`<br/>`after_verify(user, context)` |
-| `unverify` | `before_unverify_sync(user, context)`<br/>`before_unverify(user, context)`<br/>`after_unverify_sync(user, context)`<br/>`after_unverify(user, context)` |
-| `update_metadata` | `before_update_metadata_sync(user, original_user, context)`<br/>`before_update_metadata(user, original_user, context)`<br/>`after_update_metadata_sync(user, original_user, context)`<br/>`after_update_metadata(user, original_user, context)` | 
+| `roles` | `before_roles_changed_sync(user, original_user, context)`<br/>`before_roles_changed(user, original_user, context)`<br/>`after_roles_changed_sync(user, original_user, context)`<br/>`after_roles_changed(user, original_user, context)` |
+| `enable` | `before_enable_changed_sync(user, original_user, context)`<br/>`before_enable_changed(user, original_user, context)`<br/>`after_enable_changed_sync(user, original_user, context)`<br/>`after_enable_changed(user, original_user, context)` |
```
chpapa: naming is confusion. Either:

1. Keep `[before|after]_enabled` / `[before|after]_disabled`, or
1. `[before|after]_enable_status_changed`.

I prefer the former.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -172,25 +176,38 @@ Followings are hooks of auth actions:
...
+| `password` | `before_password_changed_sync(user, original_user, context)`<br/>`before_password_changed(user, original_user, context)`<br/>`after_password_changed_sync(user, original_user, context)`<br/>`after_password_changed(user, original_user, context)` |
+| `verify` | `before_verify_changed_sync(user, original_user, context)`<br/>`before_verify_changed(user, original_user, context)`<br/>`after_verify_changed_sync(user, original_user, context)`<br/>`after_verify_changed(user, original_user, context)` |
```
chpapa: naming is confusion. Either:

1. Keep `[before|after]_verified` / `[before|after]_unverified`, or
1. `[before|after]_verification_status_changed`.

I prefer the former.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -172,25 +176,38 @@ Followings are hooks of auth actions:
...
+| `metadata` | `before_metadata_changed_sync(user, original_user, context)`<br/>`before_metadata_changed(user, original_user, context)`<br/>`after_metadata_changed_sync(user, original_user, context)`<br/>`after_metadata_changed(user, original_user, context)` |
```
chpapa: Is `metadata_updated` better than `metadata_changed`?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -172,25 +176,38 @@ Followings are hooks of auth actions:
...
+| `user` | `before_user_changed_sync(user, original_user, context)`<br/>`before_user_changed(user, original_user, context)`<br/>`after_user_changed_sync(user, original_user, context)`<br/>`after_user_changed(user, original_user, context)` |
```
chpapa: `user_updated`? Feel a bit strange, "user changed or user updated" sounds like it is the users have changed to something else, while I know what it meant is "user object updated". We need a better name. Some idea:

- `[before|after]_user_object_updated`
- `user_profile_sync` / `user_profile_async` (this one is not consistent and don't have a "before/after" prefix, the idea is this hook is special and only for syncing profile to DB, which is always "after")

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -172,25 +176,38 @@ Followings are hooks of auth actions:
...
+   2. a user changes password by himself.
+   3. a user forgot password, and reset password via email form.
+
+   They are corresponding to following `context.req.path`:
+
+   1. `/reset_password`
+   2. `/change_passowrd`
+   3. `/forgot_password/reset_password`
+
+2. `user_changed` hooks are useful when a developer isn't sure which hook to implement, or it can be a hook to centralize a business logic that may be crossed over many hooks.
```
chpapa: I don't think this is a good description. A better description should be:

2. `user_profile_sync` / `user_profile_async` are special hooks for a developer to implement an external user profile store and want ONE hook to capture all changes to the user auth data or metadata.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,441 @@
...
+  UNIQUE (user_id)
+);
+```
+
+The structure of user object could be:
+
+```
+{
+    id: <id>,
+    createdAt: <createdAt>,
```
carmenlau: I think here we are showing how the user object looks like only, but not the data transmit? It is important for request and response that we use underscore instead of camelcase for variable names.
For the predefined metadata too, will cloud sdk will do the conversation too? Or we will just keep it simple to use underscore for all platform?

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,441 @@
...
+ */
+function before_XXX_sync(user, context) {
+    console.log(user.metadata.loveCat); // false
+    
+    // alter user metadata
+    user.metadata.loveCat = true;
+    
+    /*
+     * or rasie exception
+     * throw new Error("some error");
```
carmenlau: Instead of js `Error` object, I think we can use `SkygearError` here to provide error code too?

kaojohnny:
@chpapa @carmenlau 

Some updates:
1. Update hook naming, I use `metadata_modified` and `user_object_modified`, I think "modified" is more clear than "changed" or "updated". Feel free to let me if you think it is still ambiguous.
2. I updated the user object to use the underscore style instead of camelCase.
3. Change to use `SkygearError`.

chpapa:
@kaojohnny just wondering have you considered the idea of making the hook for user profile special?

I was thinking it should be special without "after/before" , and even always async maybe better becoz:

1. The only use case for that shall be implementing external profile DB.
1. So there should be no use case for the before hook.
1. If we remove the sync hook too, we can do optimization for that hook in future, such as batch calling (group multiple changes to one request, eventually consistent)

It seems better than keeping before/after and sync/async for the hook.

kaojohnny:
@chpapa Thanks for your reply. 

For `user_profile_sync` and `user_profile_async`, I have some concerns:

1. `before_user_object_modified_sync`, `before_user_object_modified`, ..., are general hooks and also fulfill implementing external DB requirement.
2. besides sync external user profile DB, I think `before_user_object_modified_sync`, `before_user_object_modified`, ..., are also good candidates for implementing common business logic requirements across all hooks, such as proprietary logging system, user validation involves metadata and auth data (before hook it is).
3. for batching API requests to change user attributes, I may not get your idea clearly, in my initial thoughts, it is more like client SDK's job? It saves multiple changing requests, and merge them together and send the merged single request at one time.

chpapa:
> 1. `before_user_object_modified_sync`, `before_user_object_modified`, ..., are general hooks and also fulfill implementing external DB requirement.

Why would people want to use the before sync / before async hook to implement external DB requirements? I don't see any reason the before sync hook should be used instead of after sync / after async hook, more over before async SHOULD NOT be used to implement external DB. It is just confusing.

> * besides sync external user profile DB, I think `before_user_object_modified_sync`, `before_user_object_modified`, ..., are also good candidates for implementing common business logic requirements across all hooks, such as proprietary logging system, user validation involves metadata and auth data (before hook it is).

I think you meant some sort of audit system by "proprietary logging system"? In that case they should still use the after hook only. User validation involves metadata and auth data.... that should be implemented at some sort of before login or signup hook?



> 3\. for batching API requests to change user attributes, I may not get your idea clearly, in my initial thoughts, it is more like client SDK's job? It saves multiple changing requests, and merge them together and send the merged single request at one time.

No, I'm talking about batch processing between Skygear Auth and the 3rd parties user profile DB. That should be done by Skygear Auth when it call `after_user_object_modified`/`user_profile_sync`

kaojohnny:
1. What I am thinking is a developer is free to use those hooks to fulfills the app's requirement, and "implement external profile DB" is one of them. I can't find a particular reason why we don't provide consistent hook design on user object modified event except extra hooks could raise confusion.
2. Suppose a developer has a validation related both to a user's `disabled` and `verified` state at the same time, he has to implement four hooks `before_enable_sync`, `before_disable_sync`, `before_verified_sync` and `before_unverified_sync`. On the contrary, a developer can have a sole validation in `before_user_object_changed_sync`.
3. Not so sure how to optimization batching operation for now, but for external user profile DB, a developer may want to have `sync` hook to guarantee data integrity. Just provide `async` hook may not sufficient.

chpapa:
> 1. What I am thinking is a developer is free to use those hooks to fulfills the app's requirement, and "implement external profile DB" is one of them. I can't find a particular reason why we don't provide consistent hook design on user object modified event except extra hooks could raise confusion.

The reasons, as explained above are as follows:

1. Easier and less confusion, e.g. so the user won't mistakenly use `before_user_object_sync` to do the external profile DB
1. Possibility to optimize the API, such as the batch processing idea above.

> 2\. Suppose a developer has a validation related both to a user's `disabled` and `verified` state at the same time, he has to implement four hooks `before_enable_sync`, `before_disable_sync`, `before_verified_sync` and `before_unverified_sync`. On the contrary, a developer can have a sole validation in `before_user_object_changed_sync`.

validation of what? What exactly is the use case? For example, if it is to validate at login, the user only have to write the code at `before_login_sync`, and access the user object to validate it at once. I don't see why he need to implement four hooks at you suggested.

If it is an external DB profile, with a field that depends on the value of `disabled` and `verified`, he could understand it as the same "sync with external user profile" functions.

> 3\. Not so sure how to optimization batching operation for now, but for external user profile DB, a developer may want to have `sync` hook to guarantee data integrity. Just provide `async` hook may not sufficient.

We also don't need to implement the optimization now, but just make sure we have the room for optimization in future. In fact, we don't need the `sync` or `async` concept to ensure data integrity, the best way to implement it, in my opinion is:

1. `sync_user_profile` hook is always async relative to the operation that trigger the data change (there are no reason why they need to be sync at all)
1. We guarantee eventually consistent only -- if the sync call is failed, it will retry in a decay timeout.

kaojohnny:
@chpapa I had updated user modified hook at 0aa2d71. But I still want to point out: the hooks naming `user_object_sync` and `user_object_async` implies and restricts the usage of the hooks, and if we do need to add "BEFORE" hook in the future for some reason, it may introduce extra confusion and difficulty to explain the reason.

carmenlau:
Would like to share my thoughts on sync or async. I think we will still want to have both sync or async. If consistency is important to the application, they may want to use sync version hook to sync the user object. That can make sure the user can only signup or do the operation when both auth gear and webhook is working. If the webhook is down, retry immediately may not help. 
If the consistency is relatively not important, they can use async to have better performance.

chpapa:
@kaojohnny @carmenlau thoughts and comments:

1. Regarding naming, I thought if the hooks is specifically for user profile sync, we could make the naming directly related to it. `user_object_sync` is more based on our concept of calling it user object, not related with profile. I thought `sync_user_profile` was good but idea are welcomed.

> But I still want to point out: the hooks naming `user_object_sync` and `user_object_async` implies and restricts the usage of the hooks, and if we do need to add "BEFORE" hook in the future for some reason, it may introduce extra confusion and difficulty to explain the reason.

2. I think we should just make sure it is a hook for the profile sync, if there are other user_object DB operation hooks required, that is separated and follow the old sematic of [before|after]_[name]_[async]

> If consistency is important to the application, they may want to use sync version hook to sync the user object. That can make sure the user can only signup or do the operation when both auth gear and webhook is working. If the webhook is down, retry immediately may not help, more sophisticated retry mechanism is needed.

Let me try to explain in more details. The consistency here is a different concept from the DB operation. In other DB related hooks, the sync/async are (as pointed out by @kaojohnny 's spec):

```
> before_sync_hook()
> before_async_hook()
> db_operation()
> after_sync_hook()
> after_async_hook()
```

For the user profile sync use case, we can actually have 2 ways of ensuring consistency for sync:
1. Option A: Do `after_sync_hook()` as suggested by @carmenlau above, which will rollback DB operations if failed to call.
2. Option B: Do a special `after_async_hook()` that guarantee "eventually consistent", which may batch process the data update, and retry. Basically it is like a journaling DB.

And here are the pros and cons of A and B:

Option A:
* Pros: Simple to implement
* Pros: Ensure real time consistency
* Cons: Increase latency to ALL user object related DB operation
* Cons: Decrease the uptime of the Skygear Auth System (because if the 3rd party fail, the user object related operation failed too)

Option B:
* Pros: No latency or uptime effect on the Auth System
* Cons: More complicated
* Cons: Eventually consistent, not real time

I think Option B is much better, so I suggest even if we don't implement the "full version of Option B" now, we should design the API so that Option B is the default. And we make sure it is simple for users to implement the external user profile.

kaojohnny:
@chpapa Thanks for your detail explanation. I try to conclude `sync_user_profile`:
1. a __AFTER__ hook: for external user profile DB, before the hook is useless
2. a __ASYNC__ hook: ASYNC hook makes future optimization possible

According to this, I think hook arguments should be modified as well; it should be `sync_user_profile(user).` Removing `original_user` and `context` is because these two arguments make no sense if the hook executed in batching update.

It seems this hook is so special, do you think we can keep origin `before_user_object_modified_sync(user, original_user, context)`, `before_user_object_modified(user, original_user, context)`, ..., which provides a alternative approach for developers?

chpapa:
> According to this, I think hook arguments should be modified as well; it should be `sync_user_profile(user).` Removing `original_user` and `context` is because these two arguments make no sense if the hook executed in batching update.

Agree.



> It seems this hook is so special, do you think we can keep origin `before_user_object_modified_sync(user, original_user, context)`, `before_user_object_modified(user, original_user, context)`, ..., which provides a alternative approach for developers?

Wouldn't reject the idea but I don't have the right use case in my head yet.

chpapa:
> According to this, I think hook arguments should be modified as well; it should be `sync_user_profile(user).` Removing `original_user` and `context` is because these two arguments make no sense if the hook executed in batching update.

Sorry I just look at the pr, I didn't meant to say it is mutally exclusive. Sometime Option A is needed, but it is just we should recommend users, unless they have a strong reason that need real time sync, always choose Option B for performance and reliability.

On the other hand, I don't have a good idea for naming.... `sync_user_profile` and `sync_user_profile_strong`? @@

kaojohnny:
I can't think of better naming either...`sync_user_profile` and `batch_sync_user_profile`?

chpapa:
> I can't think of better naming either...`sync_user_profile` and `batch_sync_user_profile`?

I think the important factor is for real-time sync or not, batch is just a side benefit for eventually consistent.

Some brainstorm...

`sync_user_profile` / `sync_user_profile_realtime`
`sync_user_profile` / `sync_user_profile_strong`
`user_profile_sync` / `user_profile_async`

kaojohnny:
@chpapa @tensiuyan @Steven-Chan @carmenlau I added some naming candidates here, free to modify the table to add you ideas.

| AFTER ASYNC (default choice) | AFTER SYNC |
| -------- | -------- | 
| `async_user_profile(user)` | `sync_user_profile(user, original_user, context)` |
| `sync_user_profile(user)` | `sync_user_profile_realtime(user, original_user, context)` |
| `sync_user_profile(user)` | `sync_user_profile_strong(user, original_user, context)` |
| `user_profile_async(user)` | `sync_user_profile_realtime(user, original_user, context)` |
| `sync_user(user)` | `sync_user_coinciding(user, original_user, context)` |

carmenlau:
@kaojohnny The functions arguments are different between sync and async. Would like to know is it intended?

kaojohnny:
@carmenlau For future possible optimization, after async hook would be executed in batch, and `original_user` and `context` makes no sense in batch execution (where user object's change may come from a series of requests).

chpapa:
I think they can have the same (user only?) argument.

I like user_sync and blocking_user_sync, seems most clear (and blocking sounds negative to warn the problem)

carmenlau:
After detailed consideration, I think we may want to use `sync_user` and `blocking_sync_user` instead. Since we have convention to put `sync` or `async` at the end of the function name for the other hooks, user may be confused that `user_sync` is an `async` function actually. I updated the table for these 2 functions.

tensiuyan:
Talked with a few colleagues about their opinion on the naming convention,  here's a summary of their feedback:
- the name looks confusing as there are two different `sync` in this set of hooks (e.g. after_user_login_sync vs sync_user) 
- they'd expect that there'd be a hook named something like `user_modified` and `user_updated` to do the auth-db sync
- They prefer a lower level name than a use case based name since they think we shouldn't restrict how developers use a hook. Besides,  when developers want to do things other than the auth-db sync using the same hook (e. g. Log user changes),  the hook name wouldn't have covered this usage if it is called `sync_user`.

These feedback are just for reference. We do not have to rule out the reasons why we wanted to have a use case based name in the first place. Hope it helps.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -0,0 +1,440 @@
...
+
+2. `user_sync` and `blocking_user_sync` are special hooks, they are triggered __AFTER__ an user object is added or there is any changes to an user object, they can be used to implement an external user profile store. It also can use as one hook to capture all changes of the user auth data or metadata.
+
+   Note that `user_sync` as an `AFTER`, `ASYNC` hook, its implementaion could be a simple after async hook like other hooks, or it can be optimized by other mechanisms (batching data update), so the `original_user` and `context` arguments are removed.
+
+   `blocking_user_sync` is an `AFTER`, `SYNC` hook. `blocking_user_sync` is especially useful when consistency is important to the application. Skygear next will rollback DB if hook failed.
+
+3. And to avoid spiral request loop, it is forbidden to send request to auth gear in hooked Cloud Function.
+
+4. `context` should contain following information:
```
chpapa: Shall we add a context that is a monotonically increasing ID (per user, or global)? Something like `context.counter`.

Another option, is we make our `context.req.id` strictly monotonically increasing. It might be useful in multiple situation in future.

The reason I'm suggesting is, how are we going to handle the following case? Let's say there are `after_enabled` and `after_disabled` called concurrently:

| Req 1 | Req 2| Remarks 
| ---- | ---- | ---- |
| enable_op |  | Req 1 called |
| after_enabled_sync() | disable_op | Req 2 called |
|  | after_disabled_sync() | |
|  | [external system update user status to disabled] | For network latency reason, req 2's hook got called first |
| [external system update user status to enabled] | | Oops

In the above situation, the external software have no way to prevent the inconsistency.

It could be avoided for the `sync_user_profile` call, if we order the network call on our end. But if other users need to handle it for rest of "action-based" hook, it seems they need a monotonic increasing ID to do so?

Correct me if I'm wrong.

chpapa:
@tensiuyan @kaojohnny @Steven-Chan I suggest we just merge this issue first, and create feature issue for the `sync_user_profile` and handle it after we launch the first version, we can revisit the issue later. It's been dragged too long and maybe we know better when we see how user uses it.

However, when I'm thinking about how to implement external DB without `sync_user_profile`, I think of a conflict situation for concurrency request, my suggestion is we add `context.counter` monotonic increasing as suggested above, see what you think.

features/256-auth-gear-interaction-overview-in-next/api.md
kaojohnny: Hopefully I understand this question correctly, as I know, this issue only happens when the client SDK didn't wait for previous op finish and execute the next one.

A developer should implement

```
await enable_op();
await disable_op();
```

Instead of 

```
enable_op();
disable_op();
```

So execution order is 

| Req 1 | Req 2| Remarks 
| ---- | ---- | ---- |
| enable_op |  | Req 1 called |
| after_enabled_sync() | | enable hook invoked |
|  | disable_op | Req 2 called |
|  | after_disabled_sync() | disable hook invoked |

chpapa: @kaojohnny no, this issue happens when two requests happens concurrently in race condition. e.g. one users logged in two devices at the same time and call the ops above.

kaojohnny:
Add remark about revisiting sync_user_profile after launch first version at 687c967.

kaojohnny:
Add data inconsistency problem at b372d64.

chpapa:
@kaojohnny thanks Johnny... although in the spec I think it is good to refer that we plan to address it in #287 :) No other problems so far for merging it first...

Issue #268, opened 2019-01-15
Re-Design of Cloud Functions at NEXT


chpapa:
# Description

This is the meta issue for re-design of Cloud Functions at NEXT.

Sub-issues:
* Routing of CF Handlers and Lambda at NEXT #264 

Here are the requirements:
* Need to support deployment of Cloud Functions (ad-hoc, no way to connect), Micro-Services (Long Running, discoverable by others), and Web
* Listing of all deployed endpoints #208 
* Basic access control by users, roles, keys #55 
* Invoke CF easily
* Support triggers by HTTP and cronjob
* Reasonable quick to deploy and run (<15 sec for a minimal functions)
* Logging #79 

These features should be considered in design and could be implemented after first launch:
* Allow custom ssl certs #153 
* Easy for developers to write and run functional / integration test for each endpoint #204 
* Easy for debug (such as hot reload & local testing #170?)
* Easy to configure headers like CORS #169 
* URL rewrite for static hosting #228 
* Look at deployment history or have immutable deployment #124 
* Parameterised Paths in HTTP handler #78 
* Stream / search logs
* Custom Domains Support #246 
* Support server side rendering #264 

Potentially in future, we may want to:
* Improve Chatbot Development Experience #17 
* Able to re-run some old requests, or have a "staging" handler that will log and compare production and staging responses for people to decide if new handler can be deployed or need to rollback.

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #278, opened 2019-01-21
WIP: Add next CF design

Steven-Chan:
connect #268

### DIFF
--- a/features/268-next-cf/api.md
+++ b/features/268-next-cf/api.md
@@ -1,0 +1,627 @@
+# Overview
+
+Skygear Cloud Function (CF) is a platform for skygear user to deploy custom logic.
+
+The followings list what are included in CF.
+
+- OpenFaas
+- CF Gateway
+  - It acts like a proxy of OpenFaas gateway
+  - It transforms Skygear requests to OpenFaas requests
+- CF Controller
+  - It manages functions in OpenFaas
+- CF Config Store
+  - For CF Gateway, it tells how the request should route to OpenFaas Gateway
+  - For CF Controller, it stores functions configurations
+  - Roughly what each function stores: id + path + openfaas path + config
+
+*Note that: except OpenFaas, these components are more like mental model, they may be merged with existing components, like gateway and its database.*
+
+Here is the diagram of all components. Intereactions between components will be shown in below sections.
+
+```
+                                              +-------------------+
+                                              |                   |
+                                              |                   |
+                                              |   Gateway         |
+                                              |                   |
+                                              |                   |
+                                              +-------------------+
+
+
+
+
+
++-------------------+                         +-------------------+             +-------------------+           +-------------------+
+|                   |                         |                   |             |                   |           |                   |
+|                   |                         |                   |             |                   |           |                   |
+|   Auth gear       |                         |   CF Gateway      |             |   CF Config Store |           |   CF Controller   |
+|                   |                         |                   |             |                   |           |                   |
+|                   |                         |                   |             |                   |           |                   |
++-------------------+                         +-------------------+             +-------------------+           +-------------------+
+
+
+                                  +-------------------------------------------------------------------+
+                                  |OpenFaas                                                           |
+                                  |           +-------------------+                                   |
+                                  |           |                   |                                   |
+                                  |           |                   |                                   |
+                                  |           |   OpenFaas        |                                   |
+                                  |           |   Gateway         |                                   |
+                                  |           |                   |                                   |
+                                  |           +-------------------+                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |           +----------------------------------------------+        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |               Function pods                  |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           +----------------------------------------------+        |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  +-------------------------------------------------------------------+
+
+```
+
+## Function deployment
+
+```
+                                              +-------------------+
+                                              |                   |
+                                              |                   |
+                                              |   Gateway         |
+                                              |                   |
+                                              |                   |
+                                              +-------------------+                                              Request from skycli
+                                                                                                                        +
+                                                                                                                        |
+                                                                                                                        |
+                                                                                                                        |
+                                                                                                                        |
++-------------------+                         +-------------------+             +-------------------+           +-------v-----------+
+|                   |                         |                   |             |                   |           |                   |(Reject if not login or
+|                   |                         |                   |             |                   |           |                   |not enough permission)
+|   Auth gear       |                         |   CF Gateway      |             |   CF Config Store <-----------+   CF Controller   |
+|                   |                         |                   |             |                   |Update     |                   |
+|                   |                         |                   |             |                   |           |                   |
++-------------------+                         +-------------------+             +-------------------+           +-------------------+
+                                                                                                                        |Call corresponding api to
+                                                                                                                        |OpenFaas gateway or k8s cluster
+                                  +-------------------------------------------------------------------+                 |
+                                  |OpenFaas                                                           |                 |
+                                  |           +-------------------+                                   |                 |
+                                  |           |                   |                                   |                 |
+                                  |           |                   |                                   |                 |
+                                  |           |   OpenFaas        <-----------------------------------------------------+
+                                  |           |   Gateway         |                                   |
+                                  |           |                   |                                   |
+                                  |           +-------------------+                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |           +----------------------------------------------+        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |               Function pods                  |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           +----------------------------------------------+        |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  +-------------------------------------------------------------------+
+```
+
+## Function invocation
+
+```
+                                                  HTTP Request
+                                                      +
+                                                      |
+                                                      |
+                                                      |     Is it a request to the gear?
+                                              +-------v-----------+
+                                              |                   |
+                                    Yes       |                   |
+         +------------------------------------+   Gateway         |
+         |                                    |                   |
+         |                                    |                   |
+         |                                    +-------------------+
+         |                                            |No
+         |                                            |
+         |                                            |
+         |                                            |
+         |                                            |
++--------v----------+                         +-------v-----------+             +-------------------+           +-------------------+
+|                   |                         |                   |             |                   |           |                   |
+|                   |                         |                   +------------->                   |           |                   |
+|   Auth gear       |                         |   CF Gateway      |Lookup       |   CF Config Store |           |   CF Controller   |
+|                   |                         |                   |routing      |                   |           |                   |
+|                   |                         |                   |             |                   |           |                   |
++-------------------+                         +-------------------+             +-------------------+           +-------------------+
+                                                      |Route to the
+                                                      |corresponding function
+                                  +-------------------------------------------------------------------+
+                                  |OpenFaas           |                                               |
+                                  |           +-------v-----------+                                   |
+                                  |           |                   |                                   |
+                                  |           |                   |                                   |
+                                  |           |   OpenFaas        |                                   |
+                                  |           |   Gateway         |                                   |
+                                  |           |                   |                                   |
+                                  |           +-------------------+                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  |           +----------------------------------------------+        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |               Function pods                  |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           |                                              |        |
+                                  |           +----------------------------------------------+        |
+                                  |                                                                   |
+                                  |                                                                   |
+                                  +-------------------------------------------------------------------+
+```
+
+# Components
+
+## OpenFaas
+
+As the base faas platform of CF, it provides the fundamental work of deploying and managing functions on a cluster.
+Details of OpenFaas would not be included here. Briefly speaking what OpenFaas provides us are:
+- function deployment
+- a watchdog program that wrap function to a http server
+- routing
+- scaling functions down to zero (like hibernater in skygear v1)
+- monitoring (with prometheus)
+
+Extra features of CF we may or may not use in the future:
+- async function invocation
+
+However, one key feature OpenFaas does not provide which we need in skygear is multi-tenancy. Once you can access to the OpenFaas Gateway, you can manage all functions managed by the gateway.
+
+So skygear user would not access the OpenFaas directly to manage their functions.
+
+### Setup
+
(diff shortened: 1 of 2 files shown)


Steven-Chan:
Update: Added section for Security Concern

features/268-next-cf/api.md
```diff
@@ -0,0 +1,298 @@
...
+- `status`: `function created`, `deploying`, `deploy failed`, `running`, `stopped`
+- `deployed_at`: time when the function is successfully deployed
+- `failed_message`: message or data of failed deployment from the k8s or faas platform
+- `plan`: pricing plan associated with the function, indicating how much resource would be allocated and the allowed idle time
+
+### Immutable deployment
+
+When function of the same name get deployed again, a new record of a version of the function would be created and the old function would not be deleted immediately. There would be a number of version of the same function deployed on OpenFaas.
+
+By default, the latest version of the function would be exposed to the endpoint as configured by the skygear user. When there are needs to access to older version, the request only need to includes the version hash to tell the CF Gateway to invoke the specific version. (Where and how to put the version hash can be further discussed, e.g. path or header)
```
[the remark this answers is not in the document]
chpapa: @jasonho215 thx for comment! Would you mind to describe more your use case or need? That would help Steven to design if the function versioning or app versioning is enough or not.

Steven-Chan:
Meeting note:

- What to include in a version? (function, configuration, additional service provided by skygear (redis)) Need to consider different use case.
- How to host a website with our cf? If use different functions to handle different pages, the routing would be handled by cf, how it works together with function versioning?
- Add detail for routing, for different use case, e.g. mixing website and api server.
- Add user story for deploying functions. How to use skycli to deploy all functions, specific functions or delete a function (by overwriting the whole config)?

features/268-next-cf/api.md
```diff
@@ -0,0 +1,309 @@
...
+- `function1` is the name of function, function deployed with the same name will considered as the same function with different version
+- `type` is the type of the function
+  - `http-handler`
+    - a single function that accepts http requests from an exact path
+    - anytime may be scaled down to zero
+  - `http-service`
+    - an http server that accepts http requests from the path
+    - may be long running
+  - `function`:
+    - a single function (in a program) that read input from stdin and write output to stdout
```
kaojohnny: May I know more about what is the stdin and stdout of a Cloud Function? When will I use it?

Steven-Chan: `http-handler` and `http-service` are targeted to expose http api, however, we may need to support other types of trigger, e.g. cron job, event trigger.

stdin and stdout function is natively supported by Openfaas, so I think it's good to choose it as the base type for other trigger.

Otherwise if we want a long running daemon to serve these type of functions, we may consider using other ipc.

features/268-next-cf/api.md
```diff
@@ -0,0 +1,309 @@
...
+    - anytime may be scaled down to zero
+- `env` indicates the environment that the code would be built with, e.g.
+  - `docker` is the base environment that the Dockerfile would be deployed directly, or would be wrapped by another minimal Dockerfile
+  - environment of other languages (e.g. `node`, `python`, `golang`) would accept source files from the developers and build with a Dockerfile provided by us
+- `src` may specify where to find Dockerfile or source code files
+- `secrets` list what secrets would be passed to the functions, developer need to set secrets with skycli
+- `permission` describes how the request can access the function
+  - `key_required`, if true, need API key or master key
+  - `user_required`, if true, need access token of a verified and not disabled user
```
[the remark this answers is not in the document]
Steven-Chan: The `user_required` behaviour is copied from skygear v1, this makes developers have less thing to understand, though sacrificing some flexibility. I think its good to discuss if we want to allow more flexibility for developers.

> Some app may not have the idea of "verified" user, especially when they are signup with username/password

They should be able to disable the feature if they don't need it, so all users should be verified once signup.

Steven-Chan:
Update: Added a separate doc for versioning

features/268-next-cf/api.md
```diff
@@ -0,0 +1,305 @@
...
+- Gateway receive http request
+  - if that path start with one that belongs to a gear, it forwards the request to the gear
+  - otherwise it forwards the request to CF Gateway
+- CF Gateway match the path against the function routes stored in CF Config Store
+  - if a function is matched, it rewrite the path to that function and forward the request to the function
+  - otherwise it return function not found error
+
+When Gateway receive a function not found error from CF Gateway, it should transform to a generic route not found error. If the Gateway is deployed without CF Gateway configured, it should immediately return route not found error if path does not match any gear.
+
+Since CF Gateway sits behined the Gateway, there would be a naming restriction of CF routes that it cannot collide the with gear path. So currently CF routes cannot have the prefix `/auth/`.
```
kaojohnny: I think this "naming restriction" should be careful, which it implies:

1. user can't have cloud function path `/auth/` to overwrite auth gear's path (a user doesn't want to use auth gear but want to have `/auth/`), or extend auth gear's capabilities by adding CF.
2. we should reserve gear names (chat, cms, ...?) for future possible usage.

Steven-Chan:
Update: Added section for serving content and example configuration

features/268-next-cf/api.md
Steven-Chan: Yeah.

I think we may make some rules so that cf path would not conflict with the gear one, e.g. cf path cannot start with `_` and gear path would all start with `_`.

Steven-Chan:
Update: Updated SPA error response and added details for route matching

Steven-Chan:
Update:
- added some diagrams
- static assets and functions use different domain
- allow tagged version to alias a specific version

Steven-Chan:
Research:
- Can major CDN provider mix static file caching and api endpoint (no cahcing) on single domain, e.g. by respecting different HTTP cache control header

Update:
- Explain the three examples

chpapa:
> * The error handling is universal to one distribution, so both assets and api would return the same
>   thing for specific 4xx or 5xx response according to the `Error Pages` setting. <- **This seems to be a problem**

Argh, this is tricky.

Any other CDN we can use? lol

Steven-Chan:
Update: Add notes to the configuration example.

Steven-Chan:
@chpapa For CDN, I've had a short discussion with @carmenlau. We suggest a more generic solution to it.

We set the origin of the CDN to our gateway, which is responsible for both api and static assets routing. For static assets, it would proxy pass to the storage (just like v1), and allow the CDN to cache the response. For api endpoint, the server would disallow caching in CDN.

This requires the CDN to respond to cache control http header.

features/268-next-cf/api.md
```diff
@@ -0,0 +1,625 @@
...
+static:
+  - src: index.html
+    path: /
+  - src: asset
+    path: /static
+```
+
+- `https://myapp.api.skygear.io/function/ABC` -> `function-server`, forwarded path `/ABC`
+- `https://myapp.api.skygear.io/functionABC` -> `functionABC`
+- `https://myapp.api.skygear.io/functionABCD` -> `function-server`, forwarded path `ABCD`
```
kaojohnny: `https://myapp.api.skygear.io/functionABCD` -> `function-server`, forwarded path `ABCD`, this match is a little weird to me.  May I know the reason why it is a prefix match for `function-service`? Which implies `https://myapp.api.skygear.io/functionABCD` and `https://myapp.api.skygear.io/function/ABCD` got the same route.

features/268-next-cf/api.md
```diff
@@ -0,0 +1,625 @@
...
+We have two options:
+- Gears need to verify requests from gateway, maybe by having a shared secret
+- Applying network policy to restrict traffic between gear pods and other pods, https://kubernetes.io/docs/concepts/services-networking/network-policies/#isolated-and-non-isolated-pods
+
+# Features
+
+## Private functions / Micro-service discoverability
+
+- If a pod needs to be discovered by external network, services need to be setup
+- CF Gateway may block function request from outside if a function is marked as private
```
kaojohnny: I think auth hooks should be default private cloud function. Right?

features/268-next-cf/api.md
```diff
@@ -0,0 +1,625 @@
...
+    src: api
+    secrets:
+      - DATABASE_URL
+
+static:
+  - src: build/index.html
+    path: /
+  - src: build/asset
+    path: /static
+```
```
kaojohnny: Because of some files (like favicon.ico, manifest.json) are stored in root folder by default (for example I use `react-create-app` and use `npm run build` to build production app), may I know does it allow to have duplicate `path`? like

```
static:
  - src: build/index.html
    path: /
  - src: build
    path: /
  - src: build/static
    path: /static
``` 

or (`static` folder is upload automatically)

```
  - src: build/index.html
    path: /
  - src: build
    path: /
```

features/268-next-cf/api.md
Steven-Chan: `http-service` is a prefix match instead of exact match because it allows the service to serve different resources by the http request path. Otherwise exact match won't allow the service to do any routing.

For the gateway,`https://myapp.api.skygear.io/functionABCD` and `https://myapp.api.skygear.io/function/ABCD` would route to the same service, but the service, by its routing, may be handled by different handlers.

features/268-next-cf/api.md
Steven-Chan: Yes.

features/268-next-cf/api.md
Steven-Chan:
```
static:
  - src: build/index.html
    path: /
  - src: build
    path: /
  - src: build/static
    path: /static
```

The above config is not a valid one.

And because my proposal does not follow a directory structure of common build tools, I expect the developer needs to update the build tools config, or rearrange the file location after being built.

chpapa: I have to be honest I didn't look/think clearly into it, I was just triggered by the statement "because my proposal does not follow a directory structure of common build tools"

It sounds like something might seriously affect the user's experience?

features/268-next-cf/api.md
```diff
@@ -0,0 +1,627 @@
...
+- get the path of the request after the gateway rewrite
+- set the corresponding http headers and write content to the response body
+
+## Static assets
+
+In skygear v1 cloud, requests with path `/static/*` will go to s3 endpoint with proxy pass. This implies that
+- these requests will be served with the same domain of the api server
+- the content data will always pass through the skygear cluster
+
+Next cloud will serve the content with another domain, e.g. `myapp.skygear.io` for static assets v.s. `myapp.api.skygear.io` for functions and gear api, because
```
chpapa: Just want to ask if we will reconsider this. It seems if we could have the versioning on this and serve static assets via same domains, the benefits are quite great:

1. CORS or some security troubles will be gone.
1. Developers develop SPA/PWA could use Skygear platform as a total solutions for deployment and versioning besides DB.

features/268-next-cf/api.md
chpapa: Any reason for prefix match for cloud functions vs longest match for services? Would it be bad if we simply use longest match for everything?

features/268-next-cf/api.md
```diff
@@ -0,0 +1,627 @@
...
+- `https://myapp.skygear.io/static/abc` -> static assets with path `/abc` in `asset`
+- `https://myapp.skygear.io/` -> static assets `index.html`
+- `https://myapp.skygear.io/any/other/path` -> static assets `index.html`
+
+# Example Configuration
+
+## Mixing functions with micro-service
+
+```yaml
+cf:
```
chpapa: call it something other than `cf`? CF implies Cloud Functions...

features/268-next-cf/api.md
```diff
@@ -0,0 +1,627 @@
...
+
+We have two options:
+- Gears need to verify requests from gateway, maybe by having a shared secret
+- Applying network policy to restrict traffic between gear pods and other pods, https://kubernetes.io/docs/concepts/services-networking/network-policies/#isolated-and-non-isolated-pods
+
+# Features
+
+## Private functions / Micro-service discoverability
+
+- If a pod needs to be discovered by external network, services need to be setup
```
chpapa: What about if a user run micro-services A which provides a gRPC endpoint for another micro-service B? How B discover A and call it themselves?

features/268-next-cf/api.md
Steven-Chan: I'm a bit confused with prefix match vs longest match, maybe they differ in handling `/`? Maybe its also what @kaojohnny was asking that i didn't get it.

```
- Path of `http-handler` and http `function` are exact match
- Path of `http-service` are matched with prefix
```

What I want to express in `matched with prefix` should be `matched with the longest prefix`, simple string match.

Steven-Chan: > Would it be bad if we simply use longest match for everything?

My original thought is that both `http-handler` and http `function` are simple function, it's a limitation that they cannot do further routing inside the function.

Just chat with @carmenlau, she thinks it may be good to allow function also have a more flexible matching than exact match, because some 3rd party web hooks may put data in the path, e.g. user id or event name.

I have a pretty strong stand of keeping http `function` to be exact match because the function itself should not have any knowledge of the original request. But for `http-handler` since the function can receive original http request, so I think it's ok to make it have the same matching mechanism with `http-service`.

features/268-next-cf/api.md
Steven-Chan: I suddenly realise we can add a little config so that it can fit common build tools and fulfil my original proposal.

```
static:
  - src: build/index.html
    path: /
  - src: build
    exclude:
       - index.html <- this will exclude the file build/index.html
    path: /static
```

With the above configuration, everything can be put in the `build` directory, while the server won't serve the path `https://myapp.skygeario.com/static/index.html`.

(Though, actually, it is not a really big deal to also serve the path `https://myapp.skygeario.com/static/index.html`)

features/268-next-cf/api.md
```diff
@@ -0,0 +1,627 @@
...
+
+CF Gateway receives request from Gateway, and
+- forward the request to corresponding function in OpenFaas
+- for configurations of a function would be carried out in the CF Gateway, e.g.
+  - access control by keys, roles, etc, the CF Gateway connect to auth gear or the db directly to do authz
+  - secrets injection (or use k8s secrets? can be discussed)
+
+### Routing to OpenFaas
+
+Function path must be uniquely defined for OpenFaas. CF Gateway will do the path rewrite, the format would be like `/{app_name}+{version_hash}+{function_name}`, e.g. `/demo-app+abcde+hello-world` when an app called `demo-app` register a function with path `/hellow-world`.
```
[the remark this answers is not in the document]
Steven-Chan: It depends on which layer is ready to be exposed publicly. 

This part is missed in the spec yet. My thought is to put authn and authz in the function pod, so that OpenFaas api gatway should be safe to be publicly accessible.

features/268-next-cf/api.md
Steven-Chan: Yes, I think so, since our original concern was about the CDN, and we figured out there should be a feasible way to put CDN in front of skygear without special requirement on domain setting.

features/268-next-cf/api.md
Steven-Chan: We may provide a setting for developer to discover other service by name, something like `link` in docker-compose.

chpapa: Thanks. Can you write it in the spec? @Steven-Chan

features/268-next-cf/api.md
chpapa: Looking forward to the update.

features/268-next-cf/api.md
chpapa: @Steven-Chan Great. Another idea is what if we have longest prefix match on src too? In that case we don't need users to specify exclude?

(Or yes, maybe it is not a big deal to also serve the path of `/static/index.html`)

features/268-next-cf/api.md
chpapa: agree.

Steven-Chan:
Updated:
- route matching for http-handler
- static asset config, versioning, and use same domain with functions
- service to service discovery

PR #279, opened 2019-01-23
Next cli design

carmenlau:
Only the basic commands here, working on the detail
Open for comments and requirements updates


tensiuyan:
Thanks Carmen! Cross-checked with the user journey doc, below are the commands found in the doc but not on the list yet:

```
skycli call js-test
skycli call js-test data=abc
skycli call --info=stack js-test data=abc
```

```
skycli services enable mongodb --size 50GB
```

```
skycli serve --reload --skip-dep js-test
```

Since I'm not aware of the complexity and the implications of those commands, I'm simply listing them out here just in case we've overlooked something.

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,285 @@
...
+
+- skycli login (alisa of `skycli auth login`)
+
+- skycli logout (alisa of `skycli auth logout`)
+
+- skycli auth
+    - [skycli auth login](#skycli-auth-login)
+    - [skycli auth logout](#skycli-auth-logout)
+
+- skycli config
```
chpapa: I'm aware of these set of config command looks similar (and work similarly) to kubectl, which could be an advantage, however, as Skygear would works kind of "repo" based, would it be better if we make skycli adopt config (which app, which cluster), depends on the working directory's `.skyclirc` file? It also has the advantage of developers can check-in their .skyclirc file, and as long as the user is logged in, running `skycli deploy` can always deploy to the correct repo.

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,285 @@
...
+    - [skycli user create](#skycli-user-create)
+    - [skycli user list](#skycli-user-list)
+
+- skycli app
+    - [skycli app list](#skycli-app-list)
+    - [skycli app create](#skycli-app-create)
+    - [skycli app add-user [USER_EMAIL]](#skycli-app-add-user)
+    - [skycli app view-config](#skycli-app-view-config)
+    - [skycli app update-config -f [TENANT_CONFIG_YAML_FILE]](#skycli-app-update-config)
+    - [skycli app add-domain [CUSTOM_DOMAIN]](#skycli-app-add-domain)
```
chpapa: Have we decided to do or not do the "immutable deployment" like now v2? If yes we are actually mapping a custom domain to the right deployment URL with the right SSL cert only here.

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,285 @@
...
+    - [skycli app update-domain [CUSTOM_DOMAIN]](#skycli-app-update-domain)
+    - [skycli app list-domain](#skycli-app-add-domain)
+
+- skycli cf
```
chpapa: Does `skycli cf` meant it is only for Cloud Functions? Or also micro-services?

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,285 @@
...
+- skycli user (Admin only, cluster user management)
+    - [skycli user create](#skycli-user-create)
+    - [skycli user list](#skycli-user-list)
+
+- skycli app
+    - [skycli app list](#skycli-app-list)
+    - [skycli app create](#skycli-app-create)
+    - [skycli app add-user [USER_EMAIL]](#skycli-app-add-user)
+    - [skycli app view-config](#skycli-app-view-config)
+    - [skycli app update-config -f [TENANT_CONFIG_YAML_FILE]](#skycli-app-update-config)
```
chpapa: similar question to the above, would we have per repo config?

features/270-next-skycli/api.md
carmenlau: I added a section to tell where is the information store.

https://github.com/SkygearIO/features/blob/24ed5a631ebda8e8ed760c58525c1c79a55d5de5/features/270-next-skycli/api.md#config-files-structure

I prefer to store account and cluster info in user directory. And store app in repo directory, `set-app` and `set-cluster-server` will update different files.

features/270-next-skycli/api.md
carmenlau: Should be both cloud functions and micro-services, the type is defined in the cloud function config file. Maybe we need better name instead of `cf` here, let me think

features/270-next-skycli/api.md
carmenlau: We have not decided do or not to do yet. If we want to support it we may need one more commands to link the domain to the version. see https://github.com/SkygearIO/features/blob/e69039226ede835a3ad5f1ee1de97484bad51422/features/270-next-skycli/commands.md#skycli-domain-verify

For implementation, I can think of 2 ways to do it.
approach 1: Every versions need to have its own url like now v2, in the k8s cluster it will proxy pass the request to versioned url in load balancer layer.
approach 2: In the current CF design, CF gateway will do the path rewrite. If CF store has the information about custom domain to version, we can do the path rewrite there.

features/270-next-skycli/api.md
carmenlau: I don't want to store the config file in repo as the config consists of auth gear sensitive config (e.g. master key, token secret... ). So for the update config flow, user should download the config, edit and re-upload the config. Only the cluster will store the state.

Example
```
# 1. If you don't have a copy of the tenant config file, get it from the view command.
$ skycli app view-config > ./tenant-config.yaml

# 2. Update the tenant-config.yaml file

# 3. Apply the change to the app
$ skycli app update-config -f ./tenant-config.yaml

```

carmenlau:
@tensiuyan 

- I added the `invoke` command for calling cf in skycli
- Added the service commands

For the `serve` command, I think it may look differently based on the implementation. So I would suggest to design it when we design hot reload & local testing.

kaojohnny:
I find there are two different ways to update config, 

First one is:
```
$ skycli app view-config > ./tenant-config.yaml
$ skycli app update-config -f ./tenant-config.yaml
```

Another is:
```
$ skycli config set-app [APP_NAME]
$ skycli config set-cluster-server [CLUSTER_SERVER_URL]
```

Maybe we want keep one style? like:
```
$ skycli view-config > ./config.ini
$ skycli update-config -f ./config.ini
```

carmenlau:
Updates:
- Updated adding custom domain flow with domain verification. Also have suggestion on how custom domain work with immutable deployment.
- Add some new commands for auth feature configuration, like welcome email or sso.

Please see the welcome email and sso sections. I didn't complete all the feature configuration commands. Would like to hear your thought before completing them.

For those operations like setting email template, it is different to have very good ux in cli. In the spec, I tried to use both interactive cli and cli params. Interactive cli aims to make the on-boarding easier. cli with params is for ci (I had been asked for any cli commands that can set those email template in v1, developers need it for ci). There are still some ux problem in this proposal, developer can't update the configure from their original state. They need to provide all the data everytime.

Another possible solution that i can think of, when user use `skycli welcome-email config`. We will open the browser, return to cli after user edit the config in browser. But if we go for this ways, we will have many pages. Maybe we should do the portal directly. 

I suggest to go for the pure cli solution first and aim for the better ux in portal. Would like to see if there is any other idea?

carmenlau:
@kaojohnny For your comment about update config, the cases are different.

`skycli config set-app` and `skycli config set-cluster-server` are setting the cli context. Which app is the skycli editing and use which cluster.

`skycli app update-config` is updating the app tenant config.

Maybe the name makes the commands confusing. Let me think if there is another better name 🤔

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,168 @@
...
+- skycli cf
+    - [skycli cf list](commands.md#skycli-cf-list)
+    - [skycli cf deploy [FUNCTION_NAME]](commands.md#skycli-cf-deploy)
+    - [skycli cf logs [FUNCTION_NAME]](commands.md#skycli-cf-logs)
+    - [skycli cf invoke [FUNCTION_NAME] --payload [PAYLOAD_JSON]](commands.md#skycli-cf-invoke)
+
+- skycli secret
+    - [skycli secret list](commands.md#skycli-secret-list)
+    - [skycli secret create [SECRET_NAME] [SECRET_VALUE]](commands.md#skycli-secret-create)
+    - [skycli secret update [SECRET_NAME] [SECRET_VALUE]](commands.md#skycli-secret-update)
```
hidden-dimensions: To support key rotation, the stupid way is setup 2 secrets: `SECRET1` and `SECRET2` and up to the admin to replace either one, however I see most platform do not show the last update time, so it is very easy to replace the wrong one. We may do better on this or even handle the rotation is more human way.

carmenlau: Key rotation is a common use case that we should consider! We are discussing app versioning in another thread, which provide a more human way to handle rotation. (https://github.com/SkygearIO/features/pull/278)
For skycli, I think we should show the last update time to avoid human error too.
Thanks! @hidden-dimensions

features/270-next-skycli/commands.md
```diff
@@ -0,0 +1,517 @@
...
+Added domain api.example.com successfully!
+
+1. Add the TXT record below to your DNS provider to verify your own example.com
+
+    Type    Host            Value
+    TXT     example.com     skygear-cloud-verification=5636486ffc5a4dfebf4a13f480bd9a95
+
+2. Add `CNAME` record to make sure your domain is pointing to `cf.skygearapis.io`.
+
+After updating DNS records, run `now domain verify api.example.com` to verify domain.
```
kaojohnny: `now domain verify api.example.com` should be `skicli verify api.example.com`.

carmenlau: Updated, thanks @kaojohnny

carmenlau:
I updated the PR to remove the features configuration commands (e.g. `skycli welcome-email`). Since we don't have clear idea which approach is better before we collect user feedback from the first version, maybe we can go for the `update-tenant-config` command with more user friendly config (see 63c600ccc242d1b706fc2d98ba48a040400fa962) first. 

I also removed CF related commands from this PR, since I think we may still want to change them after we update the CF spec. We can discuss and confirm the CF related commands in another PR.

I am going to merge this PR if it looks good. Should there be any questions please let me know!

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Commands
+
+- skycli auth
+    - [skycli auth login](commands.md#skycli-auth-login)
+    - [skycli auth logout](commands.md#skycli-auth-logout)
+
+- skycli config
+    - [skycli config view](commands.md#skycli-config-view)
+    - [skycli config set-app [APP_NAME]](commands.md#skycli-config-set-app)
```
chpapa: If account and cluster information are in user's directory, and per app information are in per app's directory, isn't "change app" will be done by implicit action of moving into an app directory, and there should be no way for users to "change an app" without moving into the app's directory?

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+## Config files structure
+
+### ~/.skycli/skyclirc
+
+- cluster controller endpoint
+- user credentials 
+
+### ./.skyclirc
```
chpapa: will it try to go to the root git repository to find the .skyclirc? Or it is a bad practices? I don't know...

features/270-next-skycli/api.md
```diff
@@ -0,0 +1,139 @@
...
+                         |              |    |           |
+                         | gateway db   |    |    k8s    |
+                         |              |    |           |
+                         +--------------+    +-----------+
+
+```
+
+## Config files structure
+
+### ~/.skycli/skyclirc
```
chpapa: shall we consider support the XDG base directory specification at the beginning? https://wiki.archlinux.org/index.php/XDG_Base_Directory

carmenlau:
@kaojohnny @Steven-Chan @chpapa 
I merged this PR and opened PR #288 for the config update. Please check and comments are welcome :)

PR #281, opened 2019-01-29
Basic webhooks design

kaojohnny:
connects #285


features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+## Verify Request
+
+`X-Skygear-Webhhok-Signature` is generated the HMAC hex digest value from the `secret` of a hook.
+
+## TBD
+
+- How `X-Skygear-Webhhok-Signature` works?
+- Should webhook provide retry mechanism?
```
chpapa: Sure. The most standard way is to respect 503 Services Unavailable or 429 Too Many Requests, if the `Retry-After` header was set.  Similarly, we need to consider if we shall respect 301, 302, 307, 308 for redirection,

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,156 @@
...
+- Should have a limit number of hooks of a event?
+- Should support wildcard event (`*`, any time any event)?
+- Timeout of `SYNC` hooks.
```
chpapa: A default value or configurable?

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,156 @@
...
+## Verify Request
+
+`X-Skygear-Webhhok-Signature` is generated the HMAC hex digest value from the `secret` of a hook.
+
+## TBD
+
+- How `X-Skygear-Webhhok-Signature` works?
+- Should webhook provide retry mechanism?
+- Should have a limit number of hooks of a event?
+- Should support wildcard event (`*`, any time any event)?
```
chpapa: Any valid use case? gut feeling is no....

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,156 @@
...
+http://localhost:3000/<gear name>/hooks/<hook id>/delete 
+EOF
+{
+     "result": "OK"
+}
+```
+
+## Verify Request
+
+`X-Skygear-Webhhok-Signature` is generated the HMAC hex digest value from the `secret` of a hook.
```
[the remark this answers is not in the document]
kaojohnny: Thanks, I updated the spec to use `sha256` to generate HMAC.

features/101-design-webhooks/design.md
kaojohnny: I think 5s default value should be enough for webhooks.

features/101-design-webhooks/design.md
kaojohnny: I update the webhook config to accept a list of events, so we can drop wildcard support.

features/101-design-webhooks/design.md
kaojohnny: 1. For `Retry-After`, I have concerns if it returns a very long delay, it may cause problems on `SYNC` webhooks.
2. For 3xx redirect response, I have concerns if infinite loop redirection happens, so I would like to consider it is a failed response.

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,188 @@
...
+A `SYNC` hook is considered failed if it can't response within 5 seconds.
+
+## Retry mechanism
+
+If a request got
+
+- 503 Services Unavailable
+- 429 Too Many Requests
+ 
+webhook will retry after 5 seconds, at most total 3 times.
```
carmenlau: Is the retry for async only? Or both sync and async?

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,188 @@
...
+    "event": "event_name",
+    "data": {
+        "key_name_1": "event_data_key_value",
+        "key_name_2": "event_data_key_value",
+        ...
+    }
+}
+```
+
+## Responding to a webhook
```
carmenlau: Do we have the response payload example? For success and error use case. 

For example, if I want to reject the request due to validation errors. How can I pass the validation error message to the gear?

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,188 @@
...
+
+| argument | description | require |
+| -------- | -------- | -------- | 
+| `events` | A list of gear events, such as: ["after_signup", "before_login"]. | ✓ |
+| `url` | The url of the webhook. | ✓ |
+| `async` | default is `true`. | |
+| `secret` | default is empty, if provided, it will be used as the key to generate `X-Skygear-Webhhok-Signature` digest. | |
+
+
+Each gear should provide following REST interfaces:
```
carmenlau: Does it mean that we will store the hooks information in the gear db? My original thought is storing it in the tenant config (gateway db), since now the gear db contains data not the setting. Cannot think of any pros and cons, maybe good too :)

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+## Verify Request
+
+When `secret` of a webhook is not empty, `X-Skygear-Webhhok-Signature` will be added to request headers, it is the HMAC hex digest of the POST request payload. The digest is generated using the sha256 hash function and the hook `secret` as the key.
+
+## Timeout of a `SYNC` hook
+
+A `SYNC` hook is considered failed if it can't response within 5 seconds.
```
carmenlau: 5 sec maybe harsh for some cases, do we need to make it configurable?

kaojohnny: Thanks, I update the webhook config format to support timeout configuration (though I think the timeout value should have a limit, I set it to 10)

features/101-design-webhooks/design.md
kaojohnny: I update the spec to include success/error response payload.

features/101-design-webhooks/design.md
kaojohnny: I think for both sync and async should support retry.

Steven-Chan:
So each gear provide extra API for hook registration? Then where to store these information?

Feel like those gear configuration API should belong to a controller that work with skycli, instead of gears themselves.

kaojohnny:
@Steven-Chan 

1. I think a REST API is useful for a developer if they don't want to use skycli, and it is easier to demonstrate the indention of webhook registration. 
2. I think skycli should use this interface to register Cloud Functions as well.
3. The reason for the interface belongs to gear is because There is some information known by a gear only, such as event name, async/sync support.

features/101-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: Not a must, could check if SHA256 is still good enough these days lol

features/101-design-webhooks/design.md
chpapa: 1. Sorry didn't make it clear, I think SYNC request would have to ignore `Retry-After`
1. For redirection, we can simply like a browser to have a counter of max redirect. A practical use case is a redirection to a new version, and people can track the old hooks via logs.

But I agree both features are not needed in first launch (but on the spec we should have it and open issues later)

features/101-design-webhooks/design.md
chpapa: I would prefer for sync does not support retry.... that seems a bit contradict with the definition of "Sync" and the latency requirement

chpapa:
> @Steven-Chan
> 
> 1. I think a REST API is useful for a developer if they don't want to use skycli, and it is easier to demonstrate the indention of webhook registration.
> 2. I think skycli should use this interface to register Cloud Functions as well.
> 3. The reason for the interface belongs to gear is because There is some information known by a gear only, such as event name, async/sync support.

skycli must go through the controller, gear must know the hooks config. That meant controller must save  

Sounds like the controller need to be responsible for the config file and configure to each gear...

kaojohnny:
Add a section about skycli and versioning.

kaojohnny:
Todo:
- remove deploy mechanism, it should belong to skycli spec
- should add async timeout setting
- only `SYNC` supports retry
- retry should respect `Retry-After` header with some limitation

kaojohnny:
update: remove deploy mechanism and add async timeout setting.

features/101-design-webhooks/design.md
```diff
@@ -0,0 +1,129 @@
...
+                "path": "request_path",
+                "body": <original_request_body>,
+                "id": "request_id"
+            }
+        }
+    }
+}
+```
+
+Note that, `context` is above format is still under discussion.
```
chpapa: should it be:

{
  "event": xxx,
  "data": {
    "context": { } 
     ... 
  }
}

or

{
  "event": ...,
  "data": { ... },
  "context": { ... }
}

kaojohnny: In https://github.com/SkygearIO/features/pull/261, we have hooked cloud function signature as: `before_password_changed_sync(user, original_user, context)`. 

Here I would like to keep the same signature for web hook, so the request payload is defined like:
```
{
    "event": "before_password_changed",
    "data": {
        "user": <user object>,
        "original_user": <user object>,
        "context": <context object>
    }
}
```
`data` attribute is one-one mapping to hook arguments. A Developer can easily understand the pattern and use web hook like using hooked cloud function.

features/285-design-webhooks/design.md
```diff
@@ -0,0 +1,129 @@
...
+- `SYNC` - gear's operation will be blocked until a response is received. Gear should respect `SYNC` hook's response, if it indicates failed, operation should abort and return a formatted error response to the client.
+- `ASYNC` - gear's operation won't be blocked and the response from the webhook is ignored.
+
+## Request Headers
+
+Following headers will be added to webhook POST request:
+
+- `X-Skygear-Webhook-Req-ID`: an ID of the request.
+- `X-Skygear-Webhook-Async`: if is a `async` hook, then the value is `TRUE` otherwise is `FALSE`. 
+- `X-Skygear-Webhhok-Signature`: a signature string of the webhook event, a develop should can verify the request by skygear's SDK or manually.
```
Steven-Chan: `X-Skygear-Webhhok-Signature` -> `X-Skygear-Webhook-Signature`

features/285-design-webhooks/design.md
```diff
@@ -0,0 +1,129 @@
...
+| `secret` | default is empty, if provided, it will be used as the key to generate `X-Skygear-Webhhok-Signature` digest. | |
+| `timeout` | - default value of sync hook is 5 seconds, it allows to be configured up to 10 seconds.<br/>- default value of async hook is 60 seconds and up to 300 seconds.  | |
+
+## versioning
+
+Like skygear cloud functions, webhooks should be consider as a "moving part". So no mattter a user deploy webhooks by `skycli` or gear REST interface, a new app version should be created as well. Note that, a new app version and a new unit version will be generated when deploy by REST interface every time.
+
+## Verify Request
+
+When `secret` of a webhook is not empty, `X-Skygear-Webhhok-Signature` will be added to request headers, it is the HMAC hex digest of the POST request payload. The digest is generated using the sha256 hash function and the hook `secret` as the key.
```
Steven-Chan: `X-Skygear-Webhhok-Signature` -> `X-Skygear-Webhook-Signature`

features/285-design-webhooks/design.md
```diff
@@ -0,0 +1,129 @@
...
+
+Note that, `context` is above format is still under discussion.
+
+## Responding to a webhook
+
+To acknowledge receipt of a webhook, webhook endpoint should return a 2xx HTTP status code. All response codes outside this range, including 3xx codes, will indicate to skygear the request is failed.
+
+For some events, it may allow to webhook to modify changes (e.x. auth gear's before family hooks) when response code is 2xx. The behavior is defined by each gear, so will not be covered in this spec.
+
+Take auth gear as an example, its before family hooks accept that a hook returns a user object to modify current user object.
```
Steven-Chan: Do we support accepting empty body if the hook does not need to modify the object? Or we would simply suggest the developer return the original object.

features/285-design-webhooks/design.md
```diff
@@ -0,0 +1,129 @@
...
+            "user": <current user object>,
+            "req": {
+                "path": "request_path",
+                "body": <original_request_body>,
+                "id": "request_id"
+            }
+        }
+    }
+}
+```
```
Steven-Chan: I prefer `context` to be outside the `data` dictionary because

- conceptually the boundary between `data` and `context` seems clear
- `context` is a common thing to most hooks
- `data` will not be able to use the key `context`

features/285-design-webhooks/design.md
kaojohnny: I think missing fields means unmodified, so empty body is accepted. Another thing is gear should validate return object, if failed, gear should return an error to SDK. Do you think it is okay?

features/285-design-webhooks/design.md
kaojohnny: Thanks updated.

features/285-design-webhooks/design.md
kaojohnny: Understand, I had moved `context` to the root level.

features/285-design-webhooks/design.md
Steven-Chan: Got it.

Issue #282, opened 2019-01-30
Re-organize Github Repo to focus on NEXT


chpapa:
# Description

The principle are:
* After 2.0 released, there shall be no confusion on github / npm / other package managers which Skygear vesion to use
* We will continue development of 2.0 on NEXT branch of each repo, but it will replace the master branch once first alpha version of 2.0 is released.
* Current 1.0 master will have a 1.0 branch, all current users / development shall start using the 1.0 branch and avoid master branch.
* Existing Repo for 1.0 which will not appear in 2.0 (Such as `forgot_password` which is merged inside auth gear) shall be marked as archived in Github.

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #283, opened 2019-01-30
Add next repo reorganisation plan

carmenlau:
connect #282

### DIFF
--- a/features/282-next-repo-re-organization/plan.md
+++ b/features/282-next-repo-re-organization/plan.md
@@ -31,6 +31,7 @@ The top level is the git repository.
     - Change `master` to `v1-record-id`, change `next` to `master`
     - Remove obsolete v1 code (Remove all v1 code, push, pubsub, record... etc. Except auth and gateway)
     - Update ci
+        - When we move `next` to `master`, update quay.io `skygear-server` image deployment script to use `canary` tag in `v1-branch`
         - Only run `next` test
         - Remove upload binary when creating github release
         - Update `docker-hub` and `quay` docker image deployment script to deploy `skygear-auth`, `skygear-gateway` and `skygear-migrate` images.
--- a/features/282-next-repo-re-organization/plan.md
+++ b/features/282-next-repo-re-organization/plan.md
@@ -4,10 +4,10 @@
 
 The top level is the git repository.
 
-- skygear-server
+- skygear
     - gateway
     - auth gear
-    - CF gateway and related
+    - cloud functions and microservices
 - skycli
 - skygear-SDK-JS
 - skygear-SDK-iOS
@@ -60,4 +60,3 @@ The top level is the git repository.
     - (When release) Deploy v1 guides and api reference to another domain and deploy v2.0 to docs.skygear.io
 
 - Archive all v1 repositories
-- Decide new name for skygear-server?
--- a/features/282-next-repo-re-organization/plan.md
+++ b/features/282-next-repo-re-organization/plan.md
@@ -1,0 +1,63 @@
+# Re-organize Github Repo to focus on NEXT
+
+## Repository layout
+
+The top level is the git repository.
+
+- skygear-server
+    - gateway
+    - auth gear
+    - CF gateway and related
+- skycli
+- skygear-SDK-JS
+- skygear-SDK-iOS
+- skygear-SDK-Android
+- skygear-controller
+    - skycli and portal backend
+    - manage cluster users
+    - manage app tenant config
+    - CF controller (deploy cf to openfaas)
+- skygear-next-cloud
+    - Documentation, instructions and scripts to create cloud version cluster
+- skygear-deploy
+    - Documentation, instructions and scripts to create EE version cluster
+    - Documentation, instructions and scripts to run CE version (maybe another repo if they are too different)
+- skygear-doc
+- guides
+
+## Todo Checklist
+
+- skygear-server
+    - Change `master` to `v1-record-id`, change `next` to `master`
+    - Remove obsolete v1 code (Remove all v1 code, push, pubsub, record... etc. Except auth and gateway)
+    - Update ci
+        - Only run `next` test
+        - Remove upload binary when creating github release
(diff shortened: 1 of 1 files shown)


features/282-next-repo-re-organization/plan.md
```diff
@@ -0,0 +1,63 @@
+# Re-organize Github Repo to focus on NEXT
+
+## Repository layout
+
+The top level is the git repository.
+
+- skygear-server
+    - gateway
+    - auth gear
+    - CF gateway and related
```
chpapa: change it to cloud functions (or the name of the module?)

chpapa: maybe just called it platform? (so it is Cloud Functions and Microservices Platform, Skygear Platform)

features/282-next-repo-re-organization/plan.md
```diff
@@ -0,0 +1,63 @@
...
+- skycli
+- skygear-SDK-JS
+- skygear-SDK-iOS
+- skygear-SDK-Android
+- skygear-controller
```
chpapa: second thought, we can't use controller without platform anyway right? So why not just merge it to `skygear-server` too?

features/282-next-repo-re-organization/plan.md
```diff
@@ -0,0 +1,63 @@
...
+    - Documentation, instructions and scripts to run CE version (maybe another repo if they are too different)
+- skygear-doc
+- guides
+
+## Todo Checklist
+
+- skygear-server
+    - Change `master` to `v1-record-id`, change `next` to `master`
+    - Remove obsolete v1 code (Remove all v1 code, push, pubsub, record... etc. Except auth and gateway)
+    - Update ci
```
chpapa: and old CI for v1? In case it is still needed...

features/282-next-repo-re-organization/plan.md
```diff
@@ -0,0 +1,63 @@
...
+    - Change `master` to `v1-branch`
+    - Remove all code and start next code in `master`
+
+- skygear-doc and guides
+    - Change `master` to `v1-branch`
+    - Remove obsolete content and start new content in `master`. Use docs-staging.skygear.io to before release.
+    - (When release) Deploy v1 guides and api reference to another domain and deploy v2.0 to docs.skygear.io
+
+- Archive all v1 repositories
+- Decide new name for skygear-server?
```
chpapa: `skygear` is preferred as it is one for all.

features/282-next-repo-re-organization/plan.md
carmenlau: The controller will only work with the db and k8s, it will not interact with the platform directly and it is quite standalone. But you are right, controller needs platform. It needs the gateway db schema and tenant config structure. We can merge it to `skygear`, but see if we want to make it public. It was private component in v1.

chpapa: Hmmm you are right, it is a bit tricky as for portal backend, I think it should be private.

But for skycli backend (and cf controller), they need to be open source for the community edition to actually function right?

Can we move portal related code to another repo, and keep controller a open source repo or inside the main repo?

carmenlau: I thought the controller should be the endpoint of both portal and skycli (at least for EE and cloud version), most of their function overlap (signup, login, custom domain...etc.). cli and portal are just the ui.

For CE, the controller maybe another program that we can open source? The features are quite different, no user signup, no custom domain management...etc. Maybe we will not use k8s too, not decided yet. I think we can design this later when we do CE version.

carmenlau:
Updated, let me know should there be any questions. If it looks good we can merge it and follow the plan to update the repos. Thanks!

features/282-next-repo-re-organization/plan.md
```diff
@@ -0,0 +1,63 @@
...
+- skycli
+- skygear-SDK-JS
+- skygear-SDK-iOS
+- skygear-SDK-Android
+- skygear-controller
+    - skycli and portal backend
+    - manage cluster users
+    - manage app tenant config
+    - CF controller (deploy cf to openfaas)
+- skygear-next-cloud
```
chpapa: Maybe don't call it `skygear-next-cloud`? Since NEXT is not the official name.

carmenlau: We will keep this name, since this is a private repo. Original `skygear-cloud` are the v1 setup, we will update the repo description to avoid confusion.

Steven-Chan:
I find I'm still a bit confused with what would be open-sourced and what would be close-sourced, and does it matter with what features would be provided in CE, EE and cloud version.

The main concern is skygear-controller, which I think would provide different features for each version. If the controller for CE version can be close-sourced, the current plan would be OK. Otherwise, there would be one more repo needed for the open-sourced version of the controller.

Therefore, I suggest we have a document about the CE, EE and cloud version and make a list of features for what to include in each version.

chpapa:
> The main concern is skygear-controller, which I think would provide different features for each version. If the controller for CE version can be close-sourced, the current plan would be OK. Otherwise, there would be one more repo needed for the open-sourced version of the controller.

CE version is open source. EE and Cloud are not. So we probably need two repo for controller of CE vs EE+Cloud

> Therefore, I suggest we have a document about the CE, EE and cloud version and make a list of features for what to include in each version.

Not sure we have any differences in terms of features for now, except EE versions should works in private Cloud / VMs, CE versions should work on a single machine. Otherwise all versions should aims to have identical features (yes, even CE versions could have multi-tenant! lol, optional of coz)

Steven-Chan:
> Not sure we have any differences in terms of features for now, except EE versions should works in private Cloud / VMs, CE versions should work on a single machine.

I remember we had a discussion on features of the CE and EE version and how they are hosted a few weeks ago, and I'm not sure we did make the decision about it or not. So I think it's better to write it down to ensure everyone in the team is aligned.

chpapa:
Ic, but how they are hosted should be out of scope from this PR. Let's handle it on EE/CE feature issue.

PR #286, opened 2019-02-22
Next cli cloud function commands

carmenlau:
require #279

No new update for the commands yet, just separate the cf commands in other PR

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -196,20 +196,6 @@ $ skycli app view-tenant-config
 Update tenant config interactively or by providing a file. For the config format,
 please see [user facing config](user-facing-config.md)
 
-### skycli app deploy
-
-`skycli app deploy [FUNCTION_NAME]`
-
-Deploy cloud functions by reading CF config file. `[FUNCTION_NAME]` is the
-function name in config file. If `[FUNCTION_NAME]` is missing, the commands will
-update everything based on the config. (e.g. cloud code, static asset... etc)
-
-#### Flags
-
-- `--cloud-code=[FUNCTION_KEY]` Deploy specific cloud code item.
-- `--all-cloud-code` Deploy all cloud code items (e.g. static asset is not included).
-- `--static` Deploy static asset.
-
 #### Example
 
 **Update in editor**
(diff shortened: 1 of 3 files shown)


carmenlau:
@Steven-Chan @kaojohnny @chpapa @tensiuyan 
Update the spec with cloud code commands. Finally we put `deploy` under `app` instead of `cloudcode` because it will do more, not only deploying cloud code. It will deploy cloud code, update auth hook and deploy static asset in future. See if it looks good?

chpapa:
@carmenlau I'm thinking what would our command looks like for microservices and cloud functions, are they identical so we group them in cloudcode? or `functions` and `services` should have different command group? (Oops and just realize the `microservices` and `services` will conflict...)

carmenlau:
@chpapa `cloudcode` is group of `function`, `http-handler` and `http-service`. In the cf spec, we have `skygear.yaml` like this, we changed `cf` to `cloudCode` since the word `cf` is not accurate for services and handlers.

```
cloudCode:
  functionABC:
    type: function
    path: /functionABC
    # other config
  function-server:
    type: http-service
    path: /function
    # other config
  api-server:
    type: http-service
    path: /api
  marketing-campaign:
    type: http-handler
    path: /campaign2019
  ....


static:
  - src: index.html
    path: /
  - src: asset
    path: /static
```

`skycli` should be consistent with `skygear.yaml`, so `cloudcode` should represent both `function`, `http-handler` and `http-service` in skycli too. `skycli app deploy --all-cloud-code` will deploy all items under `cloudCode` in `skygear.yaml`. `skycli cloudcode list` will also list all deployed cloud code items.

Or will we consider new idea to separate them in the deployment flow? instead of group them into cloud code, if so, we may want to update the `skygear.yaml` too

Thoughts are welcomed @louischan-oursky @Steven-Chan

chpapa:
@carmenlau 

Thanks for the clear explanation.

I'm actually thinking, from user's perspective, would it be more natural if we mix `cloudcode` config with `static` -- just a question, I don't have any comments and haven't think about it deeply.

So let's imagine we rename `cloudcode` to say, `deployments`, and use the same command to deploy cloud functions, microservices, all static files.... which developer experience feel better?

carmenlau:
Maybe something like this, it looks good! I will update the CF spec first to fill in some detail and update this PR later.

```
deployments:
  functionABC:
    type: function
    path: /functionABC
    # other config
  function-server:
    type: http-service
    path: /function
    # other config
  api-server:
    type: http-service
    path: /api
    # other config
  marketing-campaign:
    type: http-handler
    path: /campaign2019
  frontend:
    type: static
    src: index.html
    path: /
  frontend-asset:
    type: static
    src: asset
    path: /static
```

carmenlau:
Updates:

- Updated cf spec to merge static asset and cloud code, rename `cloudcode` to `deployments`.
- Updated cli spec to move `cloudcode` commands under `app`, 4 new commands for user code deployment.
    - skycli app deploy
    - skycli app view-deploy
    - skycli app logs [DEPLOYMENT_ITEM_NAME]
    - skycli app invoke-function [FUNCTION_NAME] --payload [PAYLOAD_JSON]

comments are welcomed :)

chpapa:
@carmenlau 望落好似好美好

carmenlau:
@chpapa Then let's go ahead XD

@louischan-oursky @Steven-Chan Can you have a quick look when have time? I am going to merge it:)

features/270-next-skycli/commands.md
```diff
@@ -246,6 +232,85 @@ $ skycli app view-tenant-config > ./tenant-config.yaml
...
+    type: static
+    src: index.html
+    path: /
+  static-asset:
+    type: static
+    src: asset
+    path: /static
+```
+
+1. Running `skycli app deploy`, all deployment items will be deployed.
```
louischan-oursky: The synopsis

```sh
skycli app deploy < --all | ITEM_NAME >
```

is safer since user may expect `skycli app deploy` to print help.

carmenlau: Sure, updated!

Issue #270, opened 2019-01-15
skycli - Command Line interface


chpapa:
# Description

TBD

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
@tensiuyan We finished the development of Skycli already?! XD

PR #288, opened 2019-02-25
Skycli config for next

carmenlau:
connect #270

### DIFF
--- a/features/270-next-skycli/api.md
+++ b/features/270-next-skycli/api.md
@@ -53,7 +53,7 @@ skycli [group] [verb/action] [parameters]
 
 - skycli app
     - [skycli app create](commands.md#skycli-app-create)
-    - [skycli app init](commands.md#skycli-app-init)
+    - [skycli app scaffold](commands.md#skycli-app-scaffold)
     - [skycli app list](commands.md#skycli-app-list)
     - [skycli app add-user [USER_EMAIL]](commands.md#skycli-app-add-user)
     - [skycli app view-tenant-config](commands.md#skycli-app-view-config)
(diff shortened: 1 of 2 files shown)


carmenlau:
Update config to address comment in PR #279 

> If account and cluster information are in user's directory, and per app information are in per app's directory, isn't "change app" will be done by implicit action of moving into an app directory, and there should be no way for users to "change an app" without moving into the app's directory?

Yes, user will need to move into the app's directory to change app. I updated the app config file from `./.skyclirc` to `skygear.yaml`, should make it looks more explicit like v1 skycli.

> will it try to go to the root git repository to find the .skyclirc? Or it is a bad practices? I don't know...

It is common, like eslint has `.eslintrc` and babel has `.babelrc` in the app's directory. But I still want to change it to `./skygear.yaml` because

1. more explicit and clear
2. Don't want skycli to have too many config, seems looks great to combine it with the cf config. Makes `skycli deploy` more clear.

> shall we consider support the XDG base directory specification at the beginning?

Sounds great, updated the spec for that.

features/270-next-skycli/commands.md
```diff
@@ -42,7 +42,7 @@ app=myapp
 
 `skycli config set-app [APP_NAME]`
 
-Change skycli current app context to `APP_NAME`
+Change skycli current app context to `APP_NAME` by creating or updating `./skygear.yaml`.
```
chpapa: Creating? When it creates how it knows which "parent" directory is the right one?

carmenlau: ummm... Actually, there are 2 cases that user need to set app. 
1. Using general app commands (add user, update tenant config...etc.) without CF
2. Deploy CF

We don't need to worry about case 2 as user must need to write the config to deploy CF. And the problem is case 1. 

In order to avoid confusion and let user know skycli use app's directory config file to config app, is it better to remove this command? But user will need to write the minimal config file for case 1.  Thought?

features/270-next-skycli/api.md
```diff
@@ -113,13 +113,36 @@ access skycli instead of using user account.
 
 ## Config files structure
 
-### ~/.skycli/skyclirc
+### $XDG_CONFIG_HOME/skycli/config or ~/.skycli/config
```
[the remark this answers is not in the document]
carmenlau: I took git config as reference when designing this. Yes, you are right, better to stick with the spec. Updated, thanks for checking!

features/270-next-skycli/commands.md
chpapa: What about have a command for user to init a new app directory? Like create app etc?

For general app commands, they must do it in the app directory or else we will tell user there are no app in the current directory?

Just imagining the use cases, not sure if it is the easiest to use / understand

carmenlau: > For general app commands, they must do it in the app directory or else we will tell user there are no app in the current directory?

Yes, I assume validation is needed. If user call commands without login or config file, we should let user know.

> What about have a command for user to init a new app directory? Like create app etc?

If this is `init` command, i think it should help init the project directory too. I updated the PR to remove `config set-app`, update `app create` and add `app init` commands.

carmenlau:
Had a short discussion with @Steven-Chan, the `app init` command is a bit misleading that he thought the command will have changes to the app. Actually it only creates local project directory with the select template.

I updated the spec as follow, hope it will be better:
- `skycli app create` will create app only, it will not create `skygear.yaml`. It will has hints to user, they should use `skycli app scaffold` to generate the project directory.
- Rename `skycli app init` to `skycli app scaffold` which support both interactive mode or provide arguments by flags.
- Add global flag `--app=APP_NAME`, for all commands that need app as context can use this flag. The flag has higher priority than `skygear.yaml` file.
- Add 2 use cases in `api.md`, for user who use skycli with or without CF

Would like to hear your thoughts. @chpapa @kaojohnny @tensiuyan

features/270-next-skycli/api.md
```diff
@@ -113,27 +113,144 @@ access skycli instead of using user account.
...
+    Confirm? (Y/n)
+
+    Fetching the list of your apps...
+    ? Select an app to associate with the directory: (Use arrow keys)
+    > myapp
+      myapp2
+      myapp3
+    (Move up and down to reveal more choices)
+
+    Fetching examples...
```
chpapa: Is it multiple choice? Each app can have multiple js cloud functions, handler etc... right? So if it only choose one that seems a bit misleading to new user?

features/270-next-skycli/api.md
```diff
@@ -113,27 +113,144 @@ access skycli instead of using user account.
...
+
+    ? What is your app name? myapp
+    Creating app...
+    Your API endpoint: https://myapp.api.skygear.io
+    Your Client API Key: DSOJSDOSDKOJASNLSLC
+    Your Master API Key: FJOADJOFAJOFJOASDJK
+    Created app successfully!
+
+    To setup app project directory, run:
+        skycli app scaffold
```
chpapa: Ask user if they want to run app scaffold here? I guess 90% of the cases will require that? So we run `skycli app scaffold --app=[myapp]` in the current folder.

features/270-next-skycli/api.md
chpapa: Or just have another command for people to fetch example...?

carmenlau: My original thought for this command, we will have a git repo that store the example project code, each examples should be a complete project that is created based on different use case. Some sample can be very simple like `helloworld-js-function`, we can develop more complicated samples in future like `pyramid-py-service`... etc. The command will fetch the master branch of repo to get the sample list, download the selected sample and change the `skygear.yaml` app only. Since every sample is a complete project, so user can only choose once and we are not going the merge the projects.

The folder structure of the sample repo maybe something like this, one folder per sample.
```
- helloworld-js-function/
- pyramid-py-service/
- ... more sample
- samples.json
```

If we think this is a good idea, maybe we can change the name of the first version list...? To avoid misleading.
```
- helloworld-js-function
- basic-js-handler
- expressjs-service
- auth-gear-hooks
```

If we need to support select multiple, the story will be different. The list may not be define by use case. And we may want to have another commands to let user add handler or function after the project created.

carmenlau:
After short discussion with @chpapa, I update the spec again.

- Update `app create` command to ask if user want to setup the project folder now.
- Update the example project list to one `js-example`, the example will include basic usage of function, http handler, micro-service and 1 auth hook example. So user will know they can use them at the same time.

Welcome for comments:) @kaojohnny @Steven-Chan @tensiuyan

features/270-next-skycli/commands.md
```diff
@@ -1,3 +1,18 @@
+## skycli overview
+
+### Global flags
+
+`gcloud [GROUP] [--app=APP_NAME] [ACTION]`
```
kaojohnny: `gcloud` seems a typo.

features/270-next-skycli/api.md
```diff
@@ -113,27 +113,137 @@ access skycli instead of using user account.
...
+
+    # User can choose whether to setup the project folder now.
+    ? Do you want to setup the project folder now or you can do it later by
+    `skycli app scaffold` command? (Y/n) n
+
+    # Add user to app
+    $ skycli app add-user dev@example.com --app=myapp
```
kaojohnny: May I know the expected behavior/purpose of `--app=myapp`? If a developer has multiple apps on the cluster, and each app has different CFs, Will the setting `--app=myapp_2` update local skygear.yml dynamically to the config on the cluster? Or just replace the `app: myapp_2` in the skygear.yml.

If it is just a simple text replace in skygear.yml, may I know the use cases?

Thanks

features/270-next-skycli/api.md
```diff
@@ -113,27 +113,137 @@ access skycli instead of using user account.
 
 ## Config files structure
 
-### ~/.skycli/skyclirc
+### ${XDG_CONFIG_HOME:-$HOME/.config}/skycli/config
```
kaojohnny: `XDG` stands for?

chpapa: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html @kaojohnny

features/270-next-skycli/api.md
chpapa: I guess @carmenlau meant to say it just temporarily (within the running session) replace the configuration at skygear.yml.

features/270-next-skycli/api.md
kaojohnny: lol KDE thing. Thanks.

features/270-next-skycli/commands.md
carmenlau: Stupid typo 🙈Fixed, thanks!

features/270-next-skycli/api.md
carmenlau: yes, it is temporarily within the running session. The use case is for fast switching between different apps. E.g. if the admin need add user to 2 apps. He can simply use `skycli app add-user dev@example.com --app=myapp` and `skycli app add-user dev@example.com --app=myapp2`. And no need to update the `skygear.yaml`

kaojohnny: Thanks. :)

carmenlau:
I am going to merge this if no new comments. And we are going to do the create app command base on the spec. 🎉

tensiuyan:
Hi @carmenlau! I think this is ready to be merged. Thanks!

PR #289, opened 2019-03-06
Add cf request authentication and authorisation

Steven-Chan:


features/268-next-cf/api.md
```diff
@@ -325,6 +325,37 @@ cf:
...
+- If `permission` of the function to be invoked is specified, gateway may reject the request if the user does not have enough privilage
+
+## Protecting the user function
+
+In openfaas, every function would expose a cluster ip, thus accessible by other programs on the cluster.
+
+Since authentication is done in the gateway, the request forwarded from that point requires protection, otherwise other functions may invoke another function with a fake user.
+
+There are two way to protect the function:
+- sign (and encrypt) the user data, and any data that permission is needed to access
```
chpapa: so effectively each tenant or functions have a private key in somewhere like tenant config, sign and encrypt the parameters, and each microservices/cloud functions SDK need to decrypt the parameters with the public key pinned on the k8s side?

If that is the case, I suggest we might want to (okay need some serious thought on the security of it):
1. when user run `skycli deploy`, a key pair will be generated for each function/ microservices securely.
1. the private key will be sent to the controller via a secure channel to store with the secure hash
1. the public key will be bundled in the docker image of the function/services, and upload altogether to the cluster.

alternatively, is it good enough if we generate a secure hash for the real path of each cloud functions / microservices? Can we make sure there are no way for users to bruteforces a list of all real's path to call those functions/services?

PR #291, opened 2019-03-12
Update user auth in skycli

carmenlau:
We are going to use auth gear as the auth part of cluster controller, so we can gain advantage from auth gear features when we do the portal. So I update the cli spec as follow.

1. Change `user add` to `user signup` and accept public signup. 
1. Update `set-cluster-server` command and require api key.
1. Remove login with admin-key, if admin need to use skycli as admin. They should provide master key in the api-key flags in `set-cluster-server` command.

For point 1, if we do need to make signup private in future. We can use the signup hook to block the public signup.

The will deploy a single tenant auth gear with the controller to handle user signup request.

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -53,11 +53,15 @@ admin.
 
 ## skycli user
 
-### skycli user create
+### skycli user signup
 
-`skycli user create`
+`skycli user signup`
 
-Cluster admin only. Create cluster user, need to provide email and password interactively.
+Signup cluster user, command will ask for password interactively.
+
+#### Flags
+
+- `--email=[EMAIL]` Provide user email.
 
 ### skycli user list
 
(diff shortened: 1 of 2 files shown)


chpapa:
> For point 1, if we do need to make signup private in future. We can use the signup hook to block the public signup.

Love this :)

features/270-next-skycli/commands.md
```diff
@@ -2,15 +2,14 @@
 
 ### skycli auth login
 
-`skycli auth login [--email=EMAIL] [--admin-key=ADMIN_KEY]`
+`skycli auth login [--email=EMAIL]`
```
chpapa: For CI environment, let's say someone wants to use Travis to run `skycli deploy` for their applications whenever Git repo updated, how would authentication works??

tensiuyan:
Hi @carmenlau I think this is ready to be merged. Thanks!

PR #292, opened 2019-03-15
Add section for comparing openfaas and fission

Steven-Chan:


features/268-next-cf/openfaas-vs-fission.md
```diff
@@ -0,0 +1,154 @@
...
+- Very attractive pool manager
+  - excels at cold start time, consistently within a few hundreds ms
+  - very suitable for our multitenant design, consume very little resources when idle
+- Still support deploying function to separate pod
+  - which is similar to OpenFaas / v1
+
+Cons:
+- Comlicated architecture
+  - hard to trace unexpected behaviour
+- Components except the router are not scalable
```
chpapa: How bad it is?

Steven-Chan: The best case for fission handling incoming requests, all requests hit the memory cache in router, the requests would be then directly forwarded to the function pod. In this case, no request would travel the unscalable components.

Worst case is all requests require those unscalable components involved, they will be obviously unable to handle requests as fast as the scaled up router, thus become the bottleneck.

I think the problem would be insignificant in non-cloud version. For cloud version, not sure when will it reach the limit, but I think having multiple fission backend would be a simple solution to this.

Steven-Chan: Another concern is that for functions run with pool manager, they are by design unscalable, because fission does not implement scaling function pods specialised from the pool.

From the test, a single pod function can still handle a few tens requests per second in real life use case, which is not really bad but definitely good enough for bigger business.

One solution can be switching those high load functions to new deploy, but then cold start would be sacrificed.

Another solution is to make pods from pool manager scalable, and one of the main contributors from fission has opened an issue about it https://github.com/fission/fission/issues/756, though I think we will have to wait for a long time till this feature implemented, or we do it ourselves.

chpapa: I am not quiet sure about what is "functions run with pool manager" ?

Fission seems developed by quiet a small group only?

Steven-Chan: Pool manager is a terminology of fission and it is the main feature of fission to achieve fast cold start.

The concept is having a number of pre-warmed generic pod that are able to dynamically load custom code of a predefined function signature. If a generic pod is assigned with one function, the pod would serve the same function before the pod become idle and get killed.

Steven-Chan: > Fission seems developed by quiet a small group only

Yes.

Just took a look at openfaas repos contributors, seems only 1 main developer on the main repo and 3 on faas-netes (the openfaas-k8s integration).

life1347: Hi there, I just noticed this thread while surfing the issues. To solve the problem, a user can choose to use newdeploy instead.

The difference between newdeploy and poolmanager is that newdeploy creates a kubernetes deployment dedicated for a function and so no need for executor to get involved as long as minscale setting is not zero. When requests go to the function, the kubernetes HPA scales in/out the replicas to meet HPA setting.

Here is the doc for the executor: https://docs.fission.io/docs/concepts/components/core/executor/

I would love to provide more information if you need any help from me. :)

Issue #294, opened 2019-03-18
User should allow to login without specify using username or email (`loginID`)


kaojohnny:
# Description

Currently when a user login to auth gear, he has to specify what `loginID` he is using. For example: `skygear.auth.login({username: "example"}, <PW>);` or `skygear.auth.login({email: "example@example.com", <PW>});`. The functionality should extend to allow to user login without specifying what `loginID` he is using, auth gear should automatically check existing principals to find corresponding user object.

PR #295, opened 2019-03-18
next: Login without specifying login id key

kaojohnny:
connects #294

### DIFF
--- a/features/294-login-without-specifying-login-id-key/api.md
+++ b/features/294-login-without-specifying-login-id-key/api.md
@@ -1,8 +1,8 @@
-For supporting login without specifying `loginID`, have two proposals:
+For supporting login without specifying `loginID`, we have two proposals:
 
 ## Proposal 1 - auth gear find corresponding user object automatically
 
-Keeping most of current implementation, auth gear finds possible user object automatically. Consider following use case:
+We can keep most of the current implementation; auth gear finds possible user objects automatically. Following is the use case:
 
 ```javascript
 skygear.auth.signup({
@@ -10,7 +10,7 @@ skygear.auth.signup({
   username: 'example'
 }, 'password');
 
-// new:
+// find corresponding user by 'example@example.com'
 skygear.auth.login('example@example.com', 'password')
 .then((user) => {
   console.log('login successfully');
@@ -18,82 +18,149 @@ skygear.auth.login('example@example.com', 'password')
 ```
 
 
-### Possible Problem
+### Problem
 
-If user signup with incorrect `loginIDMetadataKeys`, it may cause user identify miss assign. Consider following scenario:
+`loginID` could be ambiguous, consider the following scenario:
 
 ```javascript
 
 // signup user A
-skygear.auth.signup({
-  email: 'example@example.com', 
-  username: 'example'
-}, 'password');
+skygear.auth.signup({email: 'example@example.com'}, 'password');
 
 
 // signup user B
 // B's username is A's email
-skygear.auth.signup({
-  username: 'example@example.com', 
-  email: 'example'
-}, 'password');
+skygear.auth.signup({username: 'example@example.com'}, 'password');
 
+skygear.auth.login('example@example.com', 'password');
 // doesn't know which user to match
 // user A: {email: 'example@example.com'}
+// or
 // user B: {username: 'example@example.com'}
-skygear.auth.login('example@example.com', 'password');
 ```
 
 ### Code change
 
-Just brute force and find all possible principals. 
+Should be simple. Just brute-force find all possible principals. 
 
-When auth gear is got a login request without specifying `loginIDMetadataKey`, auth gear will try to find all possible principals based on current `loginIDMetadataKeys` setting. 
+When auth gear is got a login request without specifying `loginIDMetadataKey,` auth gear will try to find all possible principals based on current `loginIDMetadataKeys` setting. 
 
-For example, if a user login with string `"example"`, and `loginIDMetadataKeys=[["username"], ["email"], ["phone"]]`, auth principal store finds all principals which has `{username: "example"}`, `{email: "example"}` and `{phone: "example"}` in `authdata` field.
+For example, if a user login with string `"example"`, and `loginIDMetadataKeys=[["username"], ["email"], ["phone"]]`, auth principal store finds all principals which `authdata` field equals to `{username: "example"}`, `{email: "example"}` and `{phone: "example"}`.
 
-And then login handler will check input password with stored password, if **only one** of the principals matched, then the user login successfully. If there are more than one principals matched, auth gear will return `Ambiguous loginID` error to the client. 
+And then login handler will check input password with principal stored password, if there is **only one** principal matched, then the user login successfully. If there are more than one principals paired, auth gear will return `AmbiguousLoginID` error to the client. 
 
-The major drawback of this solution is the API UX is not good, basically just throw the problem back to the app developer. But the situation of two users uses same password with interchanged `loginID` should be a rare case.
+Should also avoid ambiguous signup. Assume `loginIDMetadataKeys=[["username"], ["email"], ["phone"]]`, when `{username: 'example'}` is signing up, `{email: 'exmaple'}`, `{phone: 'example'}` should be rejected with error `AmbiguousLoginID`.
 
 
 ## Proposal 2 - meaningless login ID
 
-Allow a user to signup with arbitrary string, then user can use the string to do authentication. Consider following scenario:
+Allow a user to signup with arbitrary string with an arbitrary key as `loginIDMetadataKey`, and then the user can use the string to do authentication. 
+
+`<loginIDMetadataKey>: <loginID>`
+
+### Sample use case: signup with email
 
 ```javascript
-skygear.auth.signup('example@example.com', 'password');
-skygear.auth.login('example@example.com', 'password');
-
-// or signup with a list of loginID
-skygear.auth.signup([
-  'example',
-  'example@example.com',
-], 'password');
-// then user login with the arbitrary to login
-skygear.auth.login('example@example.com', 'password');
-skygear.auth.login('example', 'password');
+// user can signup with arbitrary string
+// `loginIDMetadataKey=email` is inserted implicitly
+skygear.auth.signupWithEmail('example@example.com', 'password');
+skygear.auth.login(
+  'example@example.com', 'password'
+).then((user) => {
+  console.log(user);
+  /*
+  {
+      user_id: <id>,
+      loginID_source: 'email'
+      metadata: {
+        email: 'example@example.com',
+      }
+  }
+  */
+});
 ```
+
+### Sample use case: signup with username
+
+```javascript
+// `loginIDMetadataKey=username` is inserted implicitly
+skygear.auth.signupWithUsername('example', 'password');
+skygear.auth.login(
+  'example', 'password'
+).then((user) => {
+  console.log(user);
+  /*
+  {
+      user_id: <id>,
+      loginID_source: 'username'
+      metadata: {
+        // loginIDs
+        username: 'example',
+      }
+  }
+  */
+});
+```
+
+### Sample use case: signup with custom loginID
+
+```javascript
+// signup with a dict of loginIDs
+skygear.auth.signup({
+  role: 'example',
+  phone: '8912345',
+}, 'password');
+skygear.auth.login(
+  'example', 'password'
+).then((user) => {
+  console.log(user);
+  /*
+  {
+      user_id: <id>,
+      loginID_source: 'role'
+      metadata: {
+        // loginIDs
+        role: 'example',
+        phone: '8912345'
+      }
+  }
+  */
+});
+skygear.auth.login(
+  '8912345', 'password'
+).then((user) => {
+  console.log(user);
+  /*
+  {
+      user_id: <id>,
+      loginID_source: 'phone'
+      metadata: {
+        // loginIDs
+        role: 'example',
+        phone: '8912345'
+      }
+  }
+  */
+});
+```
+
+As shown in above, the user object will list all `loginID`s in `metadata`.
 
 ### Code change
 
-1. [skygear-server] Remove `loginIDMetadataKeys`
-2. [skygear-server] Change `_auth_provider_password.authData` to `text` type.
+1. [skygear-server] Remove `loginIDMetadataKeys` setting.
+2. [skygear-server] Change `_auth_provider_password.authData` to `text` type and add `_auth_provider_password.source` field.
 3. [skygear-JS-SDK] Update SDK APIs
-   - `login(authData: Object, password: String)`
-   - `loginWithEmail(email: String, password: String)`
-   - `loginWithUsername(username: String, password: String)`
-   - `signup(authData: Object, password: String, data: Object)`
-   - `signupWithEmail(email: String, password: String, data: Object)`
-   - `signupWithUsername(username: String, password: String, data: Object)`
-   - `requestVerification(recordKey: String)`  
-   to
-   - `login(loginID: String, password: String)`
-   - `signup(loginID: String, password: String, data: Object)`
-   - `signupWithLoginIDs(loginIDs: Array: password: String)`
-   - `requestVerificationByEmail(email: String)`
-   - `requestVerificationByPhone(phone: String)`
+
+   | Old | New |
+   | -------- | -------- |
+   | `login(authData: Object, password: String)` | `login(loginID: String, password: String)` |
+   | `loginWithEmail(email: String, password: String)` | Remove |
+   | `loginWithUsername(username: String, password: String)` | Remove |
+   | `requestVerification(recordKey: String)` | `requestVerificationByEmail(email: String)`<br/>`requestVerificationByPhone(phone: String)` |
+   |  | `createPasscode(loginIDSource: String, password: String, skip2FA: Boolean)`<br/>`updatePasscode(loginIDSource: String, password: String, skip2FA: Boolean`<br/>`deletePasscode(loginIDSource: String)` |
 4. [skygear-server] Update login handler to handle the new form of the `loginID`.
+5. Implement `createPasscode`, `updatePasscode` and `deletePasscode` to allow user to add `loginID` to an existing user.
 
 ### SSO auto-link
 
@@ -101,4 +168,49 @@ For SSO auto-link user feature, have following possible solutions:
 
 1. assume `loginID` is email.
 2. provide an extra auth-hook, which request to return `userID` by email.
-3. force developer to add `email` in user's `metadata` (or by some configured method to indicate the attribute `key` in the `metadata`).
+3. Check if there is email information in the metadata:
+   1. if there is a `sso_matching_emails`, all emails in the array of value will be matched against the SSO login.
+   2. if ther is a `email`, that email will be matched against the SSO login.
+
+Solution 3 may have uniqueness issue, i.e. two user could have identical email or overlapping email in `sso_matching_emails`.
+
+### Login with a dict
+
+In V1, with proper setting, user can login with a dictionary, for example:
+
+```javascript
+loginIDMetadataKeys=[["username", "email"]]
+
+skygear.auth.signup({
+  username: 'example',
+  email: 'example@example.com',
+}, 'password');
+
+// auth gear saves: {username: 'example', email: 'example@example.com'} as the authData
+
+skygear.auth.login({
+  username: 'example',
+  email: 'example@example.com',
+}, 'password');
+```
+
+For supporting this proposal, user may modify as:
+
+```javascript
+skygear.auth.signup({
+  username_email: 'example_example@example.com',
+}, 'password');
+
+skygear.auth.login('example_example@example.com', 'password');
+console.log(skygear.auth.currentUser);
+/*
+{
+    user_id: <id>,
+    loginID_source: 'username_email'
+    metadata: {
+      // loginIDs
+      username_email: 'example_example@example.com',
+    }
+}
+*/
+```
--- a/features/294-login-without-specifying-login-id-key/api.md
+++ b/features/294-login-without-specifying-login-id-key/api.md
@@ -49,11 +49,11 @@ signupWithLoginIDs({
 For app wants to use an object as `loginID`, we suggest it encodes the object to a JSON string.
 
 ```
-LOGIN_IDS_KEY_WHITELIST = ["email", "role_phone"]
+LOGIN_IDS_KEY_WHITELIST = ["email", "role-phone"]
 
 signupWithLoginIDs({
   email: 'example@example.com',
-  role_phone: '["admin","phone"]'
+  'role-phone': '[\"admin\", \"12345\"]'
 }, 'password') ==> OK
 ```
 
@@ -62,9 +62,9 @@ which creates two `loginID`s:
 | loginID key:String | loginID value:String (`UNIQUE`) |
 | --- | --- |
 | email | example@example.com |
-| role_phone | ["admin","phone"] |
+| role-phone | [\\"admin\\", \\"12345\\"] |
 
-It's developer's responsibility to ensure the encoded string uniqueness ('["admin","phone"]' and '["phone","admin"]' are considered as two different `loginID`). 
+It's developer's responsibility to ensure the encoded string uniqueness ('[\\"admin\\", \\"12345\\"]' and '[\\"phone\\", \\"12345\\"]' are considered as two different `loginID`). 
 
 If `LOGIN_IDS_KEY_WHITELIST` contains some string (not empty), a user can't signup with a `loginID` key which is not in the list.
 
@@ -73,8 +73,8 @@ LOGIN_IDS_KEY_WHITELIST = ["email", "username"]
 
 signup({
   email: 'example@example.com',
-  role_phone: '[\"admin\", \"1234567\"]'
-}, 'password') ==> error: Unknow loginID Key ("role_phone")
+  'role-phone': '[\"admin\", \"12345\"]'
+}, 'password') ==> error: Unknown loginID Key ("role-phone")
 ```
 
 The response user object is:
@@ -85,10 +85,7 @@ Passcode-ID: null,
 {
     user_id: <id>,
     login_ids: {
-      "role-phone": {
-        role: 'example',
-        phone: '8912345',
-      },
+      'role-phone': '[\"admin\", \"12345\"]',
       username: 'example',
       email: 'example@example.com'
     },
@@ -116,10 +113,9 @@ About `Passcode-ID`, refer to [#293](https://github.com/SkygearIO/features/issue
 4. [skygear-server] Update signup and login handler to handle the updated `loginID` concept.
 5. [skygear-server] Include `LoginID-Key` in response header.
 
-Note that `signupWithLoginIDs` and `signup` may have different implementions for different SDK, such as different functions or function overloading on Java.
+Note that `signupWithLoginIDs` and `signup` may have different implementations for different SDK, such as different functions or function overloading on Java.
 
 ## Compare to [meaningless loginID](appendix.md#proposal-2---meaningless-login-id)
-
 
 ### Compound Keys
 
@@ -128,18 +124,18 @@ This proposal allows string loginID value only, for compound keys loginID, it ne
 ```javascript
 // loginID value must be a string
 signupWithLoginIDs({
-  username: 'exmaple',
+  username: 'example',
   email: 'example@example.com',
-  "role-phone": '["example", "123456"]'
+  'role-phone': '[\"example\", \"12345\"]'
 }, 'password'); 
 
 // meaningless loginID
 signup({
-  username: 'exmaple',
+  username: 'example',
   email: 'example@example.com',
-  "role-phone": {
-    role: "example",
-    phone: "123456"
+  'role-phone': {
+    role: 'example',
+    phone: '123456'
   }
 }, 'password')
 ```
--- a/features/294-login-without-specifying-login-id-key/api.md
+++ b/features/294-login-without-specifying-login-id-key/api.md
@@ -215,13 +215,17 @@ Solution 3 may have uniqueness issue, i.e. two user could have identical email o
 
 ## Proposal 3
 
-Adjust current `LOGIN_ID_METADATA_KEYS` definition, rename it `LOGIN_IDS_KEY_WHITELIST` and allow one level string `loginID` only.
+Adjust current `LOGIN_ID_METADATA_KEYS` definition, rename it `LOGIN_IDS_KEY_WHITELIST` and allow one level string `loginID` only. `LOGIN_IDS_KEY_WHITELIST` default to empty, which allows to use any string as the `loginIDSouce`.
 
 ```
+LOGIN_IDS_KEY_WHITELIST = []
+==> allows any string as loginIDSouce
+
 LOGIN_IDS_KEY_WHITELIST = ["username", "email"]
+==> allows "username" and "email" as loginIDSouce
 ```
 
-And `loginID` must be following form:
+And `loginIDData` must be following form:
 
 ```
 {
@@ -230,7 +234,7 @@ And `loginID` must be following form:
 }
 ```
 
-For example, for `LOGIN_IDS_KEY_WHITELIST = ["username", "email"]`,
+For example, for `LOGIN_IDS_KEY_WHITELIST = []`,
 
 ```
 signup({
@@ -246,7 +250,7 @@ which creates two `loginID`s:
 | username | example |
 | email | example@example.com |
 
-User can login by `username` or `email`, but not `username` and `email`.
+User can login by `username` **OR** `email`, but not `username` **AND** `email`.
 
 ```
 login('example', 'password'); ==> OK
@@ -284,9 +288,9 @@ which creates two `loginID`s:
 | email | example@example.com |
 | role_phone | ["admin","phone"] |
 
-It's developer's responsibility to ensure the encoded string uniqueness ('["admin","phone"]' and '["phone","admin"]' are two different `loginID`). 
+It's developer's responsibility to ensure the encoded string uniqueness ('["admin","phone"]' and '["phone","admin"]' are considered as two different `loginID`). 
 
-And a user can't signup a `login_id_source` not in the `LOGIN_IDS_KEY_WHITELIST` as well.
+If `LOGIN_IDS_KEY_WHITELIST ` contains some string (not empty), a user can't signup with a `loginIDSouce` not in the list.
 
 ```
 LOGIN_IDS_KEY_WHITELIST = ["email", "username"]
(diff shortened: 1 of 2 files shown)


features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,104 @@
...
+   - `requestVerificationByPhone(phone: String)`
+4. [skygear-server] Update login handler to handle the new form of the `loginID`.
+
+### SSO auto-link
+
+For SSO auto-link user feature, have following possible solutions:
+
+1. assume `loginID` is email.
+2. provide an extra auth-hook, which request to return `userID` by email.
+3. force developer to add `email` in user's `metadata` (or by some configured method to indicate the attribute `key` in the `metadata`).
```
chpapa: I think the best way (no matter it is Proposal 1 or 2) could be:

1. if `sso_matching_emails` does not exists in metadata, use `email` to match.
2. if `sso_matching_emails` exists, all emails in the array of value will be matched against the SSO login.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+### Possible Problem
+
+If user signup with incorrect `loginIDMetadataKeys`, it may cause user identify miss assign. Consider following scenario:
+
+```javascript
+
+// signup user A
+skygear.auth.signup({
```
chpapa: I think the example given here is incorrect. In this example:

User A should be only able to sign in with `{email: 'example@example.com', username: 'example', 'password'}`, while User B only can sign in with  `{email: 'example', username: 'example@example.com', 'password'}`

There are no ambiguous. 

I think you want to give an example like:

```
skygear.auth.signup({email: 'example@example.com'}, 'password');
skygear.auth.signup({username: 'example@example.com'}, 'password');
```

In the above case `skygear.auth.login('example@example.com', 'password');` will have ambigous result.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,104 @@
...
+
+Just brute force and find all possible principals. 
+
+When auth gear is got a login request without specifying `loginIDMetadataKey`, auth gear will try to find all possible principals based on current `loginIDMetadataKeys` setting. 
+
+For example, if a user login with string `"example"`, and `loginIDMetadataKeys=[["username"], ["email"], ["phone"]]`, auth principal store finds all principals which has `{username: "example"}`, `{email: "example"}` and `{phone: "example"}` in `authdata` field.
+
+And then login handler will check input password with stored password, if **only one** of the principals matched, then the user login successfully. If there are more than one principals matched, auth gear will return `Ambiguous loginID` error to the client. 
+
+The major drawback of this solution is the API UX is not good, basically just throw the problem back to the app developer. But the situation of two users uses same password with interchanged `loginID` should be a rare case.
```
chpapa: Random idea: why can't we still check uniqueness for LoginIDs with proposal 1?

Assume the system have `LOGIN_ID_METADATA_KEYS=['email', 'username', ['role', username']]`

Let's say we have a table of loginIDs, and run the following APIs:

```
skygear.auth.signupWithEmail("example@example.com", 'pw');
skygear.auth.signupWithUsername("example@example.com", 'pw'); // UNIQUE LOGINIDS ERROR
skygear.auth.signup({role: 'doctor', username: 'example@example.com'}, 'pw'); // THIS IS OK
```

I think this actually make sense?

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+## Proposal 2 - meaningless login ID
```
chpapa: Actually, I like this idea more as I originally suggested something like this lol.

The beauty of it is:
1. We can remove a configuration variable: `LOGIN_ID_METADATA_KEYS`
1. We can simply have the idea of {loginID + password} pair to even support different passwords.

On the API, I suggest we should have a `loginID-type` to indicate the loginID name, so it could be identified in request/sessions.

```
skygear.auth.signup(loginID: String, password: String, loginID-type: String=null)
skygear.auth.login(loginID: String, password: String, loginID-type: String=null)
```

Beside sthat:
* We can still have `signupWithUsername/email`, or login, where `loginID-type` is set to `Username` and `Email` respectively
* We can still have `loginID` as Object instead of String, where we have a standard way to serialize it, but maybe developers don't need it?
* In user object, we need to have a dict of LoginIDs of the user / current users

kaojohnny: In the latest update, I do some adjustments based on your comments, please refer the file for the detail:

Summaries:
> `signup(loginID: String, password: String, loginID-type: String=null, data: Object=null)` 

is quite verbose (include data filed), and because we have to present the `loginID` in `metadata`, so the `loginID-type` can't be default to `null`. Thus I keep using `signup(loginIDs: dict, password: String, data: Object)`

> skygear.auth.login(loginID: String, password: String, loginID-type: String=null)

I think login no need specifying `loginID-type` here.

> In user object, we need to have a dict of LoginIDs of the user / current users

In [253-auth-sdk-callback-in-next](https://github.com/SkygearIO/features/blob/94f5dd8d22d0107c3e48591ab0cc5448b4b0eaf1/features/253-auth-sdk-callback-in-next/api.md), we have `loginID`s in `metadata` already.

> We can still have loginID as Object instead of String, where we have a standard way to serialize it, but maybe developers don't need it?

I added a section to show to show my preferred way to support loginID as Object.

features/294-login-without-specifying-login-id-key/api.md
kaojohnny: I consider this yesterday too, but I am a little worried the case is not easy to handle in the client. For example, when user signup, the user input a string as `username`, and the system responses him the error message: `username is already taken as email already`. Of course we can omit the problem though.

features/294-login-without-specifying-login-id-key/api.md
kaojohnny: By current default setting `LOGIN_ID_METADATA_KEYS=[['username'], ['email']]`, `signup({email: 'example@example.com', username: 'example', 'password'})` and `signup({email: 'example', username: 'example@example.com', 'password'})` will become 4 principals:

`{email: 'example@example.com}`
`{username: 'example'}`
`{email: 'example'}`
`{username: 'example@example.com'}`

And a user could use one of them to login. So the example given here is ambiguous too. For demonstration purpose, I had updated it to a simpler version. Thanks.

features/294-login-without-specifying-login-id-key/api.md
kaojohnny: I add more detail of the solution 3, but I consider there could be a uniqueness problem, multiple users may have the same email in `metadata`.

chpapa:
I revised Proposal 2 a bit to merge some idea with metadata key, not sure if it is too complicated. Thoughts?

Major change from @kaojohnny: 
1. Introduce the concept of `loginID_source`, so each pair of LoginID with Password have a name.
1. Have a `loginIDs_keys` in User Object, which is a dict of string, object pair for all logins ID in metadata
1. If user sign up with a dict of LoginIDs, we serialize it for the user; I'm not quite sure about this, maybe remove this feature in first release first and consider support in future.
1. Added `createLoginID` and `createLoginIDfromMetadata`

## Proposal 2

### Sample use case: signup with email

```javascript
// user can signup with arbitrary string
// `loginID-name=email` is inserted implicitly
skygear.auth.signupWithEmail('example@example.com', 'password');
skygear.auth.loginWithEmail(
  'example@example.com', 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'username',
       passcodeID: 'default'
      }
      loginIDs_keys: {
       email: 'email'
      }
      metadata: {
        email: 'example@example.com',
      }
  }
  */
});
```

### Sample use case: signup with username

```javascript
// `loginID-name=username` is inserted implicitly
skygear.auth.signupWithUsername('example', 'password');
skygear.auth.loginWithUsername(
  'example', 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'username',
       passcodeID: 'default'
      }
      loginIDs_keys: {
       username: 'username'
      }

      metadata: {
        // loginIDs
        username: 'example',
      }
  }
  */
});
```

### Sample use case: signup with loginID consist of multiple metadata

```javascript
// signup with a dict of loginIDs
skygear.auth.signup(
'role-phone', 
{
  role: 'example',
  phone: '8912345',
}, 'password');
skygear.auth.login('example', 'password'); // ERROR: user not found
skygear.auth.login(
  {role: 'example', phone: '8912345'}, 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'username',
       passcodeID: 'default'
      }
      loginIDs_key: {
          'role-phone': ['role', 'phone']
      }
      metadata: {
        // loginIDs
        role: 'example',
        phone: '8912345'
      }
  }
  */
});
```

## Update SDK APIs

   | Old | New |
   | -------- | -------- |
   | `signUp(authData: Object, password: String)` | `signUp(loginID-name: String, loginID: Object, password: String)` <br> `signUp(loginID-name: String, loginID: String, password: String)` |
   | `signUpWithEmail(email: String, password: String)` | =`signUp('email', email, password)` |
   | `signUpWithUsername(username: String, password: String)` | =`signUp('username', username, password)` |
   | `login(authData: Object, password: String)` | `login(loginID: Object, password: String)` <br> `login(loginID: String, password: String)` <br> `login(loginID-name: String, loginID: String, password: String)`|
   | `loginWithEmail(email: String, password: String)` | =`login('email', email, password)` |
   | `loginWithUsername(username: String, password: String)` | =`login('username', username, password)`  |
   | `requestVerification(recordKey: String)` | `requestVerificationByEmail(email: String)`<br/>`requestVerificationByPhone(phone: String)` |
   |  | `createLoginID(loginID-name: String, loginID: String)` <br> `createLoginID(loginID-name: String, loginID: Object)` <br> `createLoginIDfromMetadata(loginID-name: String, metadata_keys: Array)` <br> `deleteLoginID(loginID-name: String)` |
   |  | `createPasscode(passcode-name: String, password: String, skip2FA: Boolean)`<br/>`updatePasscode(passcode-name: String, password: String, skip2FA: Boolean`<br/>`deletePasscode(passcode-name: String)` |

features/294-login-without-specifying-login-id-key/api.md
chpapa: I think for SSO, it would be developer's responsibility if they make sure emails in the `sso_matching_emails` are verified, maybe we can provide uniqueness check

kaojohnny:
@chpapa, Thanks, add some of my thoughts here:

1.  I think it is very complicated. There are many different ideas in a single user object, not easy to understand at first sight. Especially: `loginIDs_keys: {email: 'email'}` (it confued me for a while).
2. Don't think it is common usage to log in by an Object, I think "OR" case should be more common, I mean user prefer `username` **OR** `email` to login, not provide `username` **AND** `email` at the same time to log in.
3. In the sample code you update, for the case a user wants to log in by `email` **or** `username`, it has two steps to do it:
```
skygear.auth.signupWithUsername('example', 'password');
skygear.auth.createLoginID('email', 'example@example.com');
```
rather than
```
skygear.auth.signup({username: 'example', email: 'example@example.com'});
```

carmenlau:
1. For Ben's suggestion, I am not sure what does the `login_source` means. Is it the current login source? So it should be 
` login_source: {
       loginID: 'username',
       passcodeID: 'default'
      }`
` login_source: {
       loginID: 'email',
       passcodeID: 'default'
      }`
` login_source: {
       loginID: 'role-phone',
       passcodeID: 'default'
      }
` instead?

2. For `requestVerification` api change, would like to know more. In v1 and also the current next, we use tenant config to setup which fields can be verified. e.g. `email` field use smtp, `phone` field use nexmo... etc. The fields key is configurable. After config the backend, user can use the api to request verification by providing the field key in client side. In next, the field key should be the meta data key. I am not sure what data should I provide in the new api, the field key or field value?

kaojohnny:
> 2. For requestVerification api change, would like to know more. In v1 and also the current next, we use tenant config to setup which fields can be verified. e.g. email field use smtp, phone field use nexmo... etc. The fields key is configurable. After config the backend, user can use the api to request verification by providing the field key in client side. In next, the field key should be the meta data key. I am not sure what data should I provide in the new api, the field key or field value?

I the spec, I propose to use following two API to verify user: `requestVerificationByEmail(email: String)` and `requestVerificationByPhone(phone: String)`. These two APIs implies a verified email or phone is not necessarily in the user's metadata. This design enables to allow a user to have two accounts with the same verified email address if the app doesn't use email as `LoginID`. Does this make sense to you?

chpapa:
> 1. For Ben's suggestion, I am not sure what does the `login_source` means. Is it the current login source? So it should be
>    ` login_source: { loginID: 'username', passcodeID: 'default' }`
>    ` login_source: { loginID: 'email', passcodeID: 'default' }`
>    `login_source: { loginID: 'role-phone', passcodeID: 'default' }` instead?

@carmenlau yes, typo, updated.

> 2\. Don't think it is common usage to log in by an Object, I think "OR" case should be more common, I mean user prefer `username` **OR** `email` to login, not provide `username` **AND** `email` at the same time to log in.

But we still need to handle it, and I think it is better handled in the way I proposed. Because I think to ask users to "concat" the loginIDs, while the rest of loginIDs each have matching metadata, is a bad design.

For example, if a user has 2 ways to login, `email` and `role` + `email`, it will result in 3 metadata fields: `email`, `role` and `email_role`. While updating `email` metadata will update the loginID automatically, usres need to manually make sure the `email_role` are updated as well.




> 3\. In the sample code you update, for the case a user wants to log in by `email` **or** `username`, it has two steps to do it:

We can solve it by simply introducing a `signupByLoginIDs` API.

chpapa:
## Proposal 3: Completely separate LoginIDs with metadata

I'm putting my original proposal here for reference with the same format with Proposal 1 for consideration.

### Sample use case: signup with email

```javascript
// user can signup with arbitrary string
skygear.auth.signupWithEmail('example@example.com', 'password');
skygear.auth.loginWithEmail(
  'example@example.com', 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'email',
       passcodeID: 'default'
      }
      loginIDs: {
       'email': 'example@example.com'
      }
      metadata: {
      }
  }
  */
});
```

### Sample use case: signup with username

```javascript
skygear.auth.signupWithUsername('example', 'password');
skygear.auth.loginWithUsername(
  'example', 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'username',
       passcodeID: 'default'
      }
      loginIDs: {
       'username': 'example'
      }
      metadata: {
      }
  }
  */
});
```

### Sample use case: signup with loginID consist of multiple metadata

```javascript
// signup with a dict of loginIDs
skygear.auth.signup('role-phone',
{
  role: 'example',
  phone: '8912345',
}, 'password');
skygear.auth.login('example', 'password'); // ERROR: user not found
skygear.auth.login(
  {role: 'example', phone: '8912345'}, 'password'
).then((user) => {
  console.log(user);
  /*
  {
      user_id: <id>,
      login_source: {
       loginID: 'role-phone',
       passcodeID: 'default'
      }
      loginIDs: {
          'role-phone': {'role': example, 'phone': 8912345'} 
      }
      metadata: {
      }
  }
  */
});
```

## Update SDK APIs

   | Old | New |
   | -------- | -------- |
   | `signUp(authData: Object, password: String)` | `signUp(loginID-key: String, loginID: Object, password: String)` <br> `signUp(loginID-key: String, loginID: String, password: String)` |
   | `signUpWithEmail(email: String, password: String)` | =`signUp('email', email, password)` |
   | `signUpWithUsername(username: String, password: String)` | =`signUp('username', username, password)` |
   | `login(authData: Object, password: String)` | `login(loginID: Object, password: String)` <br> `login(loginID: String, password: String)` <br> `login(loginID-key: String, loginID: String, password: String)` <br> login(loginID-key: String, loginID: Object, password: String)`|
   | `loginWithEmail(email: String, password: String)` | =`login('email', email, password)` |
   | `loginWithUsername(username: String, password: String)` | =`login('username', username, password)`  |
   | `requestVerification(recordKey: String)` | `requestVerificationByEmail(email: String)`<br/>`requestVerificationByPhone(phone: String)` |
   |  | `updateLoginIDs(loginIDs: Object)` |
   |  | `createPasscode(passcode-name: String, password: String, skip2FA: Boolean)`<br/>`updatePasscode(passcode-name: String, password: String, skip2FA: Boolean`<br/>`deletePasscode(passcode-name: String)` |

chpapa:
@carmenlau @kaojohnny please bear in mind that all proposals we have so far (logically...) will need users to handle uniqueness of loginIDs in some strange situation. For example, when a user use the same value for username and emails, that is duplicated.

I think technically we can teach users that, if they want "true unique email address", they can do it like:

```
loginIDs: {
    'uniqueEmail': {'email': 'example@example.com'}
}
```

But that is super tricky...

chpapa:
@carmenlau @kaojohnny and I think all 3 proposals difference are very subtle, all 3 have the same underlying data structure. So the only differences are more on how we want the interface looks like.

I like the clarity of Proposal 3 personally.

kaojohnny:
## Proposal 4: Signup with loginID_source

1. Would like to propose signup with `loginID_source`, it makes create multiple principles at the same time.
2. passcodeID allows to be `null` when it is not specified. ("default" is a little confused to me)

The signup API is `signup(loginIDs: dict, password: string, data: dict, passcodeID: string=null)`

```javascript
skygear.auth.signup({
  username: 'example',
  email: 'example@example.com',
  'role-phone': {
    role: 'example',
    phone: '8912345',
  }
}, 'password');
```

creates three principals:

| loginID_source | loginID | 
| ------------- |-------------| 
| username | example |
| email | example@example.com |
| role-phone | {role: 'example', phone: '8912345'} |

so user can login by "example", "example@example.com" or `{role: 'example', phone: '8912345'}`

the user object is

```javascript
{
    user_id: <id>,
    login_source: {
       loginID: 'role-phone',
       passcodeID: null
    },
    login_IDs: {
        'username': 'example',
        'email': 'example@example.com',
        'role-phone': {'role': example, 'phone': '8912345'} 
    },
    metadata: {
    }
}
```

chpapa:
@kaojohnny so I think your proposal 4 is basically just mentioning the API of `signUpWithLoginIDs(loginIDs: Object, password: String)` in proposal 2/3 in details?

As mentioned above, I think the difference for 1/2/3 is basically how we represent. Proposal 1 use env var, 2 use loginIDs with ref to metadata, 3 make loginIDs a separate dict.

chpapa:
@carmenlau @kaojohnny @Steven-Chan 

To help us make the decision, let me type the summary below

1. The major design difference of Proposal 1/2/3 are on the interface:
    1. LoginIDs are defined by envvar `LOGIN_ID_METADATA_KEYS`, point to metadata
    2. LoginIDs are defined by `loginID_keys`, point to metadata
    3. LoginIDs are defined by the `loginIDs` dict
1. All 3 would have the same "uniqueness" problem, it was just which favor of uncomfortable we want on each Proposal:
    1. Since LoginIDs are defined implicitly, some strange implicit uniqueness problem will arise. E.g. two users have the same `email` and `username` will show an error for ambiguous loginIDs, which is a surprise
    2. LoginIDs are defined as a dictionary of index to metadata, which might still surprise users for uniqueness problem for new comer, but easier to figure out at the dictionary.
    3. LoginIDs are defined as a dictionary of values, less surprise.
1. Other strange points are (may not be exclusive):
    1. ?
    2. The `loginID_keys` are difficult to understand at first sight
    3. Putting `email` or some data in `loginIDs` feel strange.

kaojohnny:
I prefer the proposal of removing `AUTH_ID_METADATA_KEYS` (proposal 2 or 3), for these two proposals, addition to current user object structure mentioned above, I want to bring back my original design in https://github.com/SkygearIO/features/pull/254 which putting `email`, `username`, ... , in the root level of a user object, I still think it is easy to use and easy to understand, I mean `user.username` is quite natural to every developer and framework, especially when compare to the cost of teaching developer to understand what is `login_IDs` .

```
HTTP/1.1 200 OK
Content-Type: application/json
Login-Source: role-phone
Passcode-ID: null

{
  "user_id": "395ed6c2-ef73-40e4-b6f4-199e0b03b20b",
  "email": "example@example.com",
  "username": "example",
  "role-phone": {
    "role": "admin",
    "phone": "1234567"
  },
  "metadata": {
    "age": 52,
    "name": "oursky"
  }
}
```

And with supplement APIs:
`updateMetadata(metadata: dict)`
`updateLoginID(loginIDSource: String, loginIDValue: String|Object)`
`addLoginID(loginIDSource: String, loginIDValue: String|Object)`
`removeLoginID(loginIDSource: String)`
`createPasscode(passcode-name: String, password: String, skip2FA: Boolean)`
`updatePasscode(passcode-name: String, password: String, skip2FA: Boolean)`
`deletePasscode(passcode-name: String)`

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -105,62 +118,82 @@ skygear.auth.login(
...
 ).then((user) => {
   console.log(user);
   /*
+  LoginID-Source: username,
   {
       user_id: <id>,
-      loginID_source: 'phone'
-      metadata: {
-        // loginIDs
+      "role-phone": {
```
chpapa: I think it is cleared to put all loginIDs under `loginIDs: {}` instead of flatten them on user object.

chpapa:
@kaojohnny However, I don't think we can use `user.username` to replace `user.loginIDs.username`. The reasons are:

1. I think it is actually clearer to say `user.loginIDs.username`? Especially when you consider it could be `user.loginIDs.role-phone` vs `user.role-phone`. 
1. If we flatten all values of loginIDs to user object, does that meant people can also login with `user_id`, or even `metadata` which is also a keys in the user object? Or we keep a map of all keys in user object that could be used as loginID elsewhere (which defeats the purpose of proposal 3)

kaojohnny:
@chpapa We have a quick discussion today. I will update the spec to meet today's consensus later. Mainly for keeping things simple, we will want to keep `LOGIN_ID_METADATA_KEYS` to validate user's input, but we won't support compound keys. That means we will support `LOGIN_ID_METADATA_KEYS=[["username"], ["email"]]` but not `LOGIN_ID_METADATA_KEYS=[["username", "email"]]`, user can not login by an dictionary. And second, we consider moving `loginIDs` out of `metadata` is clear to developers.

chpapa:
> user can not login by an dictionary

I think this is false? For example, if `loginIDs { "role-phone": {role: "doctor", phone: "1234567"} }` and `LOGIN_ID_METADATA_KEYS=["role-phone"]`, the user can still login by dictionary.

> we will want to keep `LOGIN_ID_METADATA_KEYS` to validate user's input

If we make it mandatory, we need to make sure our `updateLoginIDs` API validate with the environment variable too, and have strong documentation/error code to hints the error for usability.

However I would advise making it optional (so only if user enable this config, the validation will be done on server side).

kaojohnny:
@carmenlau @Steven-Chan @chpapa I add proposal 3 https://github.com/SkygearIO/features/blob/9e09b4ddf302e5e97e9746b118dde1ae07d8cee8/features/294-login-without-specifying-login-id-key/api.md#proposal-3 according to yesterday's discussion, when you have time, please help review and give comments. Thanks.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,370 @@
...
+   2. if ther is a `email`, that email will be matched against the SSO login.
+
+Solution 3 may have uniqueness issue, i.e. two user could have identical email or overlapping email in `sso_matching_emails`.
+
+## Proposal 3
+
+Adjust `LOGIN_ID_METADATA_KEYS` definition, allows one level string `loginID` only.
+
+```
+LOGIN_SOURCES = ["username", "email"]
```
chpapa: Suggested to change it to `LOGIN_IDS_KEYS_WHITELIST`

Also, make it default as empty, and empty meant any loginIDs key could be used.

This is to avoid new users, say they want to try Skygear, and use `callsign` as the LoginID, but turns out cannot login and confused.

I think whitelist is a good balance for server side checking if it is needed.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,370 @@
...
+For example, for `LOGIN_SOURCES = ["username", "email"]`,
+
+```
+signup({
+  username: 'example',
+  email: 'example@example.com'
+}, 'password')
+```
+
+which creates two `lingID`s:
```
chpapa: :/lingID/loginID

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,370 @@
...
+```
+LOGIN_SOURCES = ["email", "role_phone"]
+
+signup({
+  email: 'example@example.com',
+  role_phone: '["admin","phone"]'
+}, 'password') ==> OK
+```
+
+which creates two `lingID`s:
```
chpapa: typo

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,370 @@
...
+
+| login_id_source:String | login_id:String (`UNIQUE`) |
+| --- | --- |
+| email | example@example.com |
+| role_phone | ["admin","phone"] |
+
+It's developer's responsibility to ensure the encoded string uniqueness ('["admin","phone"]' and '["phone","admin"]' are two different `loginID`). 
+
+And a user can't signup a `login_id_source` not in the `LOGIN_SOURCES` as well.
```
chpapa: `LOGIN_IDS_KEY_WHITELIST`, and only if it is not empty.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,370 @@
...
+  "role-phone": {
+    role: 'example',
+    phone: '8912345',
+  }
+}, 'password'); ===> proposal 2: OK, proposal 3: error: Incorrect loginID format
+```
+
+### Compound Keys
+
+Proposal 3 allows String loginID only, for compound keys loginID, it needs to use a work-around.
```
chpapa: Please create a feature issue to support compound LoginIDs after this proposal is merged.

kaojohnny: May I understand more about the feature issue? In this spec, we propose developer uses JSON string to store compound keys, in the new feature issue created after this proposal merged, does it expect we may want another mechanism to support compound keys?

chpapa: I think the features should treat compound LoginIDs as first class citizien, maybe we support JSON or other format at the API endpoint, and have the appropriate interface on each client SDK.

The end goal should be, people should feel comfortable to use LoginIDs to store the data, so they don't need to "sync" the data in loginIDs and other places (such as metadata)

And also maybe if we consider this feature seriously, we want to handle the case of "login by Role + Username or Role + Email" too.

So we drop this feature at this version, but write down the notes for everything we learnt in this discussion for the new feature, so we don't need to go through it again in future :)

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,374 @@
...
+2. provide an extra auth-hook, which request to return `userID` by email.
+3. Check if there is email information in the metadata:
+   1. if there is a `sso_matching_emails`, all emails in the array of value will be matched against the SSO login.
+   2. if ther is a `email`, that email will be matched against the SSO login.
+
+Solution 3 may have uniqueness issue, i.e. two user could have identical email or overlapping email in `sso_matching_emails`.
+
+## Proposal 3
+
+Adjust current `LOGIN_ID_METADATA_KEYS` definition, rename it `LOGIN_IDS_KEY_WHITELIST` and allow one level string `loginID` only. `LOGIN_IDS_KEY_WHITELIST` default to empty, which allows to use any string as the `loginIDSouce`.
```
chpapa: Consistency problem: is it loginIDsource? Or loginsource? Or... I saw different names..

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,374 @@
...
+
+| login_id_source:String | login_id:String (`UNIQUE`) |
+| --- | --- |
+| email | example@example.com |
+| role_phone | ["admin","phone"] |
+
+It's developer's responsibility to ensure the encoded string uniqueness ('["admin","phone"]' and '["phone","admin"]' are considered as two different `loginID`). 
+
+If `LOGIN_IDS_KEY_WHITELIST ` contains some string (not empty), a user can't signup with a `loginIDSouce` not in the list.
+
```
chpapa: Typo

kaojohnny:
@chpapa Thanks for your comments, I made some modifications, when you have time, please help review. Thanks.

1. Remove the term `loginIDSouce`, and call it `loginIDKey`.
2. Move proposal 1, and proposal 2 to another markdown file.
3. signup defined as `signup(loginIDKey: String, loginIDValue: String, password: String, data: dict)`.
4. Add `signupWithLoginIDs(loginIDs: dict, password: String, data: dict)`.

features/294-login-without-specifying-login-id-key/api.md
kaojohnny: I can't find the typo. 

If you mean `loginIDSouce` is the typo in the quoted section, `loginSouce` is typo actually. Anyway, I remove the term and call it `loginIDKey` in the API.

chpapa: :s/souce/source/g

kaojohnny: Sorry for this.

chpapa:
Thanks.

> 4\. Add `signupWithLoginIDs(loginIDs: dict, password: String, data: dict)`.
BTW, for this one, I think the development team can consider for different SDKs, we may choose to use different functions, or overload the same functions (e.g. Java) for multiple loginIDs login.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -0,0 +1,158 @@
+For supporting login without specifying `loginID`, we have two proposals:
```
chpapa: Only one proposal in the final doc?

kaojohnny: Thanks. Removed.

kaojohnny:
If there is no other comment, I will merge the PR later today. Thanks.

carmenlau:
From the user response, so the `LOGIN_IDS_KEY_WHITELIST` should have implicit meaning that `role_phone` means using 2 fields to login?

If yes,
1. we should confirm we use `role_phone` or `role-phone`, we have both of them in this spec.
1. Maybe it is better to use same form for api input and response object? IMO, using string as api input may not be a good idea since different json serializer may give different result (have space and no space) that will affect the login result.
We use string as login input, it is for "login without key" use case? (Let me know if i am wrong) If yes I would suggest using array in both api input and response.

Assume id keys have implicit meaning, `-` means multiple compound keys.
```js
LOGIN_IDS_KEY_WHITELIST = ["email", "username", "role-phone"]

signupWithLoginIDs({
  email: 'example@example.com',
  role_phone: ["admin", "phone"]
}, 'password') // ==> OK

login('example', 'password'); // ==> OK
login('example@example', 'password'); // ==> OK
login(["admin", "phone"], 'password'); // ==> OK
login({
  username: 'example'
}, 'password') // ==> OK
login({
  username: 'example',
  email: 'example@example.com'
}, 'password') // ==> won't support this.

// response
{
    user_id: <id>,
    login_ids: {
      "role-phone": ["admin", "phone"],
      username: 'example',
      email: 'example@example.com'
    },
    metadata: {
      age: 18
    }
}

// user access role
user["login_ids"]["role-phone"][0]
```

For login without key case, auth gear will compare the login ids if the same length only.
For example, if the app has `LOGIN_IDS_KEY_WHITELIST = ["email", "username", "role-phone", ,"role-userid"]`

```js
login('example', 'password'); // will search ids {"email": "example"} or {"username": "example"}
login(["admin", "1234"], 'password'); // will search ids {"role-phone": ["admin", "1234"]} or {"role-userid": ["admin", "1234"]}
```

We will still store loginID value as string in db, auth gear will do the serialize and deserialize

carmenlau:
So we will not use the *meaningless loginID* approach, right? I am confused about the compound keys section here. https://github.com/SkygearIO/features/blob/ada4a8e87ac2ba42a83c32821aeb39012ce06a41/features/294-login-without-specifying-login-id-key/api.md#compound-keys

Sorry for the late questions 🙇🏻‍♀️

kaojohnny:
@carmenlau Thanks for your comments.

Followings are original ideas, hope to clarify anything I don't state clearly:

1. `LOGIN_IDS_KEY_WHITELIST` is empty by default, so a user can use any string as the loginID key, which implies `role-phone` is not necessarily means compound keys.
2. In skygear-server, it saves loginID in the table:
```
CREATE TABLE _auth_provider_password (
  principal_id text PRIMARY KEY REFERENCES _auth_principal(id),
  login_id_key text,
  login_id_value text,
  password text NOT NULL,
  CONSTRAINT _auth_provider_password_login_id_value_key UNIQUE (login_id_value)
);
```
3. It's developer's responsibility to create unique loginID string value, JSON string is recommended one. So `signup("role-phone", "[\"admin\", \"12345\"]")` and `signup("role-phone", "[\"12345\", \"admin\"]")` are considered to creating two users. A user may also use other string form, such as `signup("role-phone", "{"role":"admin","phone":"12345"}")` is fine as well.
4. So `login(loginIDValue: String, password: String)` is the only API signuture, user doesn't need loginIDKey in login operation.
5. I fixed the user object as:
```
LoginID-Key: username,
Passcode-ID: null,
{
    user_id: <id>,
    login_ids: {
      "role-phone": '[\"admin\", \"1234567\"]',
      username: 'example',
      email: 'example@example.com'
    },
    metadata: {
      age: 18
    }
}
```
---
May I know your concern is about SDK should help serialize compound keys as loginID value? And help deserialize corresponding loginID value as well.

chpapa:
> 2\. So `login(loginIDValue: String, password: String)` is the only API signuture, user doesn't need loginIDKey in login operation.

@kaojohnny I think we still need a `login(loginIDKey: String, loginIDValue: String, password: String)` in case someone want to specify the login ID...?

carmenlau:
Thanks for clarifying @kaojohnny. I think too mush as I saw object in the user response, so we will only have string login id in this stage.

From the conversation in this thread https://github.com/SkygearIO/features/pull/295#discussion_r268044458, we will create another feature issue for compound loginIDs support, right? If this is the case, I think it is better not to suggest user to serialize the object to JSON string in this stage. Since we only do string comparison in auth gear, the serialized json maybe different in different SDK and have unexpected result. And we should cover the compound loginIDs case in the new feature issue, not here. Thought? @kaojohnny @chpapa 

The other parts look good to me, thanks Johnny!

kaojohnny:
@chpapa @carmenlau I updated the spec for
1. Not suggest using JSON for compound keys.
2. Add API for login with specific loginIDKey.

chpapa:
Specification wise, can we suggest that maybe another simpler approach for compound key LoginID, is we can support string as login ID only, instead we add a few called something like `domain:String`, to suppose the case of "Role" + "LoginID"? It seems more straight forward and easy. Compound key could be too complicated and don't have much use case beyond a simple "Domain" + "LoginID" model.

PR #299, opened 2019-04-09
Update cloud function spec with secret management and remove tenant config from versioning

carmenlau:
- Add secret management detail
    - Make secret immutable and define the detail behaviour
- Remove tenant config from versioning
    - Discussed with @Steven-Chan, both of us think since we only have single auth gear version, so it makes sense that app only have single version tenant config.

### DIFF
--- a/features/268-next-cf/api.md
+++ b/features/268-next-cf/api.md
@@ -614,3 +614,53 @@ Currently, micro-service allow developers to do routing in the app, while others
 - Custom Domains Support #246 (https://github.com/SkygearIO/features/issues/246)
 
 Handled by the k8s ingress controller. Skycli may provide command for skygear user to add these features.
+
+## Secret management
+
+With immutable deployment, secrets are also immutable. For managing secrets,
+following operations are allowed.
+
+- skycli secret create
+- skycli secret remove
+- skycli secret rename
+
+To avoid human error, deployment will be failed if config file is using secrets
+that are not exist. Removing or renaming a secret will not affect existing deployment.
+If skygear user want to update a secret (e.g. key rotation), he will need to rename the
+existing secret to another name. Create a secret with the same name and re-deploy the functions.
+
+### Example: Key rotation
+
+```sh
+$ skycli secret rename MY_SECRET_KEY MY_SECRET_KEY_OLD
+$ skycli secret create MY_SECRET_KEY [KEY_VALUE]
+$ skycli cf deploy
+
+# Test the function and works great
+# Remove the old key
+$ skycli secret remove MY_SECRET_KEY_OLD
+```
+
+### Implementation Details
+
+**Secret Model**
+
+- id
+- name
+- k8s_secret_name
+- created_at
+- app_id
+- deleted
+
+Controller will only store the secret name to k8s secret name mapping. Controller
+will not keep a copy of the secret value, the values will be stored in k8s only.
+`skycli secret remove` will only soft delete the mapping record, the k8s secret
+will still be there, in case some old deployment are using them. After a secret
+is removed, it cannot be used in the new deployment. Rename a secret will soft
+delete the mapping record, and re-create a new secret with new name.
+
+When deploying a cloud function, skygear user can specify the secrets for a
+function. Controller will check the mapping and create fission function with
+k8s secrets. Fission support reading secrets value from files. We will update
+the fission environment runtime to convert those values to environment variable,
+so skygear user can access those secret form env.
--- a/features/268-next-cf/versioning.md
+++ b/features/268-next-cf/versioning.md
@@ -37,7 +37,6 @@ So the moving parts of an app for skygear user would be,
 - Tenant configuration
 - CF functions configuration
 - Static assets
-- Secrets
 
 ### Moving unit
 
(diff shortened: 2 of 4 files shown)


carmenlau:
@kaojohnny @chpapa @Steven-Chan I update the spec to add some detail about secret management of CF. Please take a look when have time and let me know for any comments. I will start developing first. thanks!:)

features/268-next-cf/api.md
```diff
@@ -614,3 +614,53 @@ Currently, micro-service allow developers to do routing in the app, while others
...
+
+To avoid human error, deployment will be failed if config file is using secrets
+that are not exist. Removing or renaming a secret will not affect existing deployment.
+If skygear user want to update a secret (e.g. key rotation), he will need to rename the
+existing secret to another name. Create a secret with the same name and re-deploy the functions.
+
+### Example: Key rotation
+
+```sh
+$ skycli secret rename MY_SECRET_KEY MY_SECRET_KEY_OLD
```
chpapa: Will we provide a function to rotate key in future? A recommended way to handle it as a separate feature issue?

carmenlau:
> What is our current secrets management model like? Is that each tenant can only access their own secrets?

Yes, secrets are under an app (tenant). Skygear user needs to add the secret to the app first by using`skycli secret create ...` and then they can use it in their cf config like this
```
cf:
  function1:
    type: http-handler
    path: /hello-world
    env: node
    src: js/hellow-world
    secrets:
      - DATABASE_URL
```

> We can keep it simple now but may need to jot down the attack vectors (e.g. user could possibly break something and access the k8s secrets store?) And create feature issues related with it.

In the current architecture, user will not access the k8s directly. Instead all the operations will be done through the skygear controller, so the attack vectors will may.
- If the k8s credentials leak and user can access the k8s cluster directly, all secrets may be exposed
- (From k8s) If a node is compromised, the only secrets that could potentially be exposed should be the secrets belonging to containers scheduled onto it
- (From k8s) If the master is compromised, all secrets in the cluster may be exposed

Since we are using managed k8s, the master, node and secret store security should be handled by provider. So different providers will be different. For google, *By default, GKE encrypts customer content stored at rest, including Secrets*. 

k8s secrets: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/secrets.md#constraints-and-assumptions
google gke encrypting secrets: https://cloud.google.com/kubernetes-engine/docs/how-to/encrypting-secrets

> Will we provide a function to rotate key in future? A recommended way to handle it as a separate feature issue?

Do you mean a single command for key rotation? Or you think this flow is not struct forward? Sure, we can create separate feature issue if needed.:)

carmenlau:
Thanks for comments:) @chpapa 

> What is our current secrets management model like? Is that each tenant can only access their own secrets?

Yes, secrets are under an app (tenant). Skygear user needs to add the secret to the app first by using`skycli secret create ...` and then they can use it in their cf config like this
```
cf:
  function1:
    type: http-handler
    path: /hello-world
    env: node
    src: js/hellow-world
    secrets:
      - DATABASE_URL
```

> We can keep it simple now but may need to jot down the attack vectors (e.g. user could possibly break something and access the k8s secrets store?) And create feature issues related with it.

In the current architecture, user will not access the k8s directly. Instead all the operations will be done through the skygear controller, so the attack vectors will may.
- If the k8s credentials leak and user can access the k8s cluster directly, all secrets may be exposed
- (From k8s) If a node is compromised, the only secrets that could potentially be exposed should be the secrets belonging to containers scheduled onto it
- (From k8s) If the master is compromised, all secrets in the cluster may be exposed

Since we are using managed k8s, the master, node and secret store security should be handled by provider. So different providers will be different. For google, *By default, GKE encrypts customer content stored at rest, including Secrets*. 

k8s secrets: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/secrets.md#constraints-and-assumptions
google gke encrypting secrets: https://cloud.google.com/kubernetes-engine/docs/how-to/encrypting-secrets

> Will we provide a function to rotate key in future? A recommended way to handle it as a separate feature issue?

Do you mean a single command for key rotation? Or you think this flow is not struct forward? Sure, we can create separate feature issue if needed.:)

chpapa:
> In the current architecture, user will not access the k8s directly. Instead all the operations will be done through the skygear controller, so the attack vectors will may.

Yea, I just want to make sure we don't have stupid things like, etcd store all secrets, and it is open access, so anyone who can "guess/scan" the etcd node IP address could get all secrets.

 

> Do you mean a single command for key rotation? Or you think this flow is not struct forward? Sure, we can create separate feature issue if needed.:)

Yea I think it is not very straight forward, but could be a future issue.

carmenlau:
@chpapa I created separate feature issue for key rotation. fyi https://github.com/SkygearIO/features/issues/301

Let me know if there is any further comments? I am going to merge it if it looks good:)

Any comments? @Steven-Chan @kaojohnny 

Thanks!! 🙏🏻

features/268-next-cf/api.md
```diff
@@ -614,3 +614,53 @@ Currently, micro-service allow developers to do routing in the app, while others
...
+
+With immutable deployment, secrets are also immutable. For managing secrets,
+following operations are allowed.
+
+- skycli secret create
+- skycli secret remove
+- skycli secret rename
+
+To avoid human error, deployment will be failed if config file is using secrets
+that are not exist. Removing or renaming a secret will not affect existing deployment.
```
kaojohnny: > Removing or renaming a secret will not affect existing deployment.

Want to confirming about rename secret.

1. After rename secret, all existing deployment keeps no change, a developer may need to update each `secrets` section of a `cf` in `skygear.yaml` and run deploy again to use the secret by the updated name.
2. When a developer run `skycli secret list`, he cannot see secrets with the previous name.

carmenlau:
I am going to merge this feature issue and the secret feature is implemented. Let me know or open new feature issue if there is any problem.

PR #303, opened 2019-04-15
AuthInfo in cloud code

louischan-oursky:
Rendered version is here https://github.com/louischan-oursky/features/blob/auth-info-in-cloud-code/features/302-auth-info-in-code-code/auth-info-in-cloud-code.md

### DIFF
--- a/features/302-auth-info-in-code-code/auth-info-in-cloud-code.md
+++ b/features/302-auth-info-in-code-code/auth-info-in-cloud-code.md
@@ -1,0 +1,91 @@
+# AuthInfo in cloud code
+
+## Overview
+
+The gateway must resolve the HTTP header `x-skygear-access-token` into `x-skygear-auth-info`
+when the request has been determined to be routed to cloud code.
+The cloud code runtime must decode `x-skygear-auth-info` and
+provide it as a value to the underlying application.
+How it is provided should follow language and framework convention.
+The cloud code SDK must provide ways to decode it.
+
+## x-skygear-auth-info
+
+This is a HTTP Header containing a [AuthInfo](#AuthInfo) encoded with [base64url](#base64url)
+
+Example
+
+```
+x-skygear-auth-info: eyJfaWQiOiAiODdkZmFhY2YtYTg3Mi00NDRhLTk0OGEtMTQ5N2M2YmIyYTAzIiwgImRpc2FibGVkIjogZmFsc2UsICJ2ZXJpZmllZCI6IGZhbHNlfQ==
+```
+
+Note that this representation should be treated as implementation details and subject to changes.
+In the future, signature may be added so that this header can be verified.
+
+If `x-skygear-access-token` is absent, then `x-skygear-auth-info` is absent as well.
+Therefore, the provided value must use [Option Type](https://en.wikipedia.org/wiki/Option_type).
+
+## AuthInfo
+
+This is a JSON Object of the following fields.
+
+- `_id`: string
+- `disabled`: boolean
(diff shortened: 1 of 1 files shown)


features/302-auth-info-in-code-code/auth-info-in-cloud-code.md
```diff
@@ -0,0 +1,91 @@
...
+In the future, signature may be added so that this header can be verified.
+
+If `x-skygear-access-token` is absent, then `x-skygear-auth-info` is absent as well.
+Therefore, the provided value must use [Option Type](https://en.wikipedia.org/wiki/Option_type).
+
+## AuthInfo
+
+This is a JSON Object of the following fields.
+
+- `_id`: string
```
carmenlau: Do we still using `_id`? Better to sync with the user response of auth gear. 

In this spec, we are using `user_id`
https://github.com/SkygearIO/features/pull/254/files

features/302-auth-info-in-code-code/auth-info-in-cloud-code.md
```diff
@@ -0,0 +1,91 @@
...
+}
+```
+
+## base64url
+
+This encoding based on https://tools.ietf.org/html/rfc4648#section-5 without any linebreaks, whitespaces.
+
+## How to provide the [AuthInfo](#AuthInfo) in the cloud code runtime
+
+This is language and framework specific. Suppose cloud code in written in a Express-like framework.
```
carmenlau: So in the first cloud function runtime (nodejs), where will we decode the authinfo? in the environment run or sdk? I think we may want to support both? Do it in environment, so sdk will not be required for simple cloud function. SDK one will be for microservices?

louischan-oursky: I agree that we should support both. What are microservices in the context of skygear next?

chpapa:
I have a more fundamental question: Why use a single HTTP header with BASE64URL encode JSON object? Instead of simply have multiple HTTP headers with prefix? Say:
`X-SKYGEAR-AUTH-USERID`
`X-SKYGEAR-AUTH-VERIFIED`
`X-SKYGEAR-AUTH-DISABLED`
?

The PROs and CONs are....? I have two PROs for using HTTP Header directly:

1. Easier to debug and readable
2. For Microservices, the users don't need to use our SDK to "decode" the auth info or read the manual to understand it is a JSON base64url encoded.

louischan-oursky:
The original thought is to provide a full `_core_user` row serialized as a JSON Object. But I have changed my mind to include the 3 essential fields due to the reason I have stated in the spec. So, yes multiple HTTP headers without any encoding is more human-friendly.

I am not sure about how do we verify the request in a later phase. Here is how github supports request verification https://developer.github.com/webhooks/securing/#validating-payloads-from-github . If we follow that approach we do not need to sign each header individually so no need to encode each header.

chpapa:
> I am not sure about how do we verify the request in a later phase. Here is how github supports request verification https://developer.github.com/webhooks/securing/#validating-payloads-from-github . If we follow that approach we do not need to sign each header individually so no need to encode each header.

We want to sign and verify the auth header, not the payload. One advantage of using ONE HTTP HEADER WITH BASE64 ENCODED JSON is the signing could be the full string of the header's value.

If we split it to multiple header, the signing will be a bit tricky. (Such as signing the concat of userID + verified + disabled).

Well so basically both way is a bit annoying....

louischan-oursky:
I have splitted the header into multiple headers and document the proposed headers signing approach.

PR #306, opened 2019-05-02
Add flow charts for cloud code deployment

Steven-Chan:


carmenlau:
The charts look good to me, if @louischan-oursky doesn't have questions feel free to merge

PR #309, opened 2019-05-06
Support deploy all items in skygear.yaml

carmenlau:
Add detail for using `skycli app deploy` to deploy all functions in skygear.yaml

### DIFF
--- a/features/268-next-cf/api.md
+++ b/features/268-next-cf/api.md
@@ -698,6 +698,7 @@ Every deployment (one or multiple items) will create a new deployment.
 - `app_id`: skygear app id
 - `version`: deployment version, uuid or hash
 - `tag`: deployment tag, the live version will have tag with `latest`
+- `status`: deployment status (pending, running, deploy failed, stopping, stopped, stop failed)
 
 **`deployment_cloud_code`**
 
@@ -720,6 +721,47 @@ Denormalized routing table for gateway routing.
 - `tag`: deployment tag, the live version will have tag with `latest`
 
 **`static_asset` (TBD)**
+
+
+#### Deployment and Cloud Code Status
+
+Deployment and cloud code status values: pending, running, deploy failed, stopping, stopped, stop failed
+
+Status of deployment flow
+
+```
+          +
+          | Start deploy
++---------v-----------+              +-----------------------------------------+
+| Deployment: pending |  same failed |Deployment: deploy failed                |
+| Cloud code: pending +-------------->Cloud code: running/pending/deploy failed|
++---------+-----------+              +-----------------------------------------+
+          |
+          | all success
+          |
++---------v-----------+
+| Deployment: running |
+| Cloud code: running |
++---------------------+
+```
+
+Status of cleanup flow
+
+```
+         + Start cleanup
+         |
++--------v-----------+             +----------------------------------------+
+|Deployment: stopping| same failed |Deployment: stop failed                 |
+|Cloud code: stopping+------------->Cloud code: stopping/stopped/stop failed|
++--------+-----------+             +----------------------------------------+
+         |
+         | all success
+         |
++--------v-----------+
+|Deployment: stopped |
+|Cloud code: stopped |
++--------------------+
+```
 
 #### Deployment flow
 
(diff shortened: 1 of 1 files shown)


features/268-next-cf/api.md
```diff
@@ -664,3 +664,97 @@ function. Controller will check the mapping and create fission function with
...
+User can create or update particular item by `skycli app deploy [ITEM_NAME]`.
+e.g. User had deployed function 1 to 4. If user use
+`skycli app deploy function5` to deploy function 5, the new app version will
+have function 1 to 5 after deployment.
+
+### Implementation Details
+
+#### Models
+
+**`app_version`**
```
louischan-oursky: maybe `app_deployment` is a better name?

features/268-next-cf/api.md
```diff
@@ -664,3 +664,97 @@ function. Controller will check the mapping and create fission function with
...
+
+Every deployment (one or multiple items) will create a new app_version.
+
+- `id`
+- `created_at`: deployment time
+- `app_id`: skygear app id
+- `version`: deployment version, uuid or hash
+- `tag`: deployment tag, the live version will have tag with `latest`
```
louischan-oursky: We need a status column here because the below flow references it. It would be great if we can write down the possible values.

We may need think about the relationship between app_version.status and cloud_code.status.

features/268-next-cf/api.md
carmenlau: how about just call it `deployment`? XD I thought calling it `deployment` before.

features/268-next-cf/api.md
carmenlau: Yes, i missed it, let me add it back!

features/268-next-cf/api.md
louischan-oursky: Good!

carmenlau:
Updated! Thanks :D

carmenlau:
@chpapa @Steven-Chan Please kindly have a look when have time. Some background here, in the first version, we will not have multiple deployment version. But this model design should be compatible if we want to have multiple version support in future.

I will start work on the deployment model first.

features/268-next-cf/api.md
```diff
@@ -666,3 +666,139 @@ function. Controller will check the mapping and create fission function with
...
+- `cloud_code_id`
+
+**`cloud_code_route`**
+
+Denormalized routing table for gateway routing.
+
+- `id`
+- `created_at`
+- `version`: deployment version, will be the same as `version` of `deployment`
+- `path`: user defined path
```
chpapa: stupid question maybe: so it is the path to match which HTTP path should match to which function? Is it always prefix matching or...?

If it is prefix matching maybe better call it `http_path_prefix` something like that is more clear?

It sounds very performance sensitve to me too (or how that works?)

features/268-next-cf/api.md
```diff
@@ -666,3 +666,139 @@ function. Controller will check the mapping and create fission function with
...
+- `app_id`: skygear app id
+- `backend_url`: cloud code backend url
+- `tag`: deployment tag, the live version will have tag with `latest`
+
+**`static_asset` (TBD)**
+
+
+#### Deployment and Cloud Code Status
+
+Deployment and cloud code status values: pending, running, deploy failed, stopping, stopped, stop failed
```
chpapa: just curious, so our clodecode backend (in this case Fission) will have a webhook / pubsub whatever to call our controller to update the status?

features/268-next-cf/api.md
carmenlau: > it is the path to match which HTTP path should match to which function?

yes

>  Is it always prefix matching or...? 

Should be prefix matching for handler, services and static asset. Exact match for function. For the detail in route matching, we can see [route matching](https://github.com/SkygearIO/features/blob/master/features/268-next-cf/api.md#route-matching)

> If it is prefix matching maybe better call it http_path_prefix something like that is more clear?

Here is the db model of controller backend, but we also use `path` in `skygear.yaml`. Maybe we want to change it?

 It sounds very performance sensitve to me too (or how that works?)

> The routing is done by gateway, now we did it by using db sql query. We may want to have task to improve gateway performance in future, like how to reduce the query. Maybe using cache etc..

features/268-next-cf/api.md
carmenlau: We are now polling the status by using fission client (calling the fission controller server) during the deployment task.

PR #311, opened 2019-05-17
Microservice

louischan-oursky:
Rendered version here
https://github.com/louischan-oursky/features/blob/305-microservice/features/305-microservice/microservice.md

It is my vision of supporting microservice. It simply allows user to provide their Dockerfile and build an image capable of serving HTTP at a given port.

### DIFF
--- a/features/305-microservice/microservice.md
+++ b/features/305-microservice/microservice.md
@@ -1,0 +1,85 @@
+# Microservice
+
+## Overview
+
+The document specifies the details of `type: http-service` of a deployment item.
+
+## Configuration of the deployment item
+
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
+    path: /api/
+    port: 8888
+    context: ./backend/service
+    dockerfile: Dockerfile
+```
+
+When the user run `skycli app deploy my-awesome-service`, the following steps are taken:
+
+1. Assert `./backend/service` to be a directory.
+1. If `dockerfile` is specified, it is relative to `context`. Otherwise, the default value is `context/Dockerfile`.
+1. Assert the Dockerfile to be a file.
+1. Look for `<context>/.dockerignore`. If it is not present, the entire context is archived. Otherwise, use the same mechanism as `.dockerignore` to ignore files.
+1. Upload the archive and save as an artifact.
+1. Follow the existing deployment flow.
+
+When the built image is run, it must expose an HTTP service listening at `port`.
+
+## Docker Image Registry
+
+A Docker registry is deployed to store all microservice images.
+It is served with plain HTTP and requires basic authentication.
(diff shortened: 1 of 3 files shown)


tensiuyan:
@chpapa @carmenlau could you please review this today or tomorrow as we need to confirm the spec soon in order to proceed.  Many thanks!

carmenlau:
The spec looks good to me so far. But we may want to cover the secrets and environment variable config in the spec too. We found we used the word `env` in cf to specific the function running environment, but we may also want to use `env` as environment variable setting in microservice. We may want to rename `env` in cf probably? I think we can have another PR for that and merge this first.

chpapa:
Agree with @carmenlau , not much problem for me on the spec.

features/305-microservice/microservice.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Configuration of the deployment item
+
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
+    path: /api/
+    port: 8888
+    context: ./backend/service
```
chpapa: Is the word `context` commonly used for docker? It seems a bit untrival to me. Would other name such as `docker-context` or `docker-build-path` more obvious?

features/305-microservice/microservice.md
```diff
@@ -0,0 +1,85 @@
...
+## Overview
+
+The document specifies the details of `type: http-service` of a deployment item.
+
+## Configuration of the deployment item
+
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
```
chpapa: how would the config looks like if users didn't pass a docker file, but instead, point to a path with node.js looks like? Is the current configuration compatible with that? Or we need a new `type` (prefer no)?

So other config such as `context` need to be considered.

features/305-microservice/microservice.md
louischan-oursky: Yes. See https://docs.docker.com/compose/compose-file/#context

features/305-microservice/microservice.md
louischan-oursky: The type `http-service` is docker only. In the future we can have other types like `nodejs-service` which itself is just an abstraction over `http-service`. That is we can provide the Dockerfile for them given that the project follows some assumed conventions.

The Dockerfile may look like

```dockerfile
FROM node:10-alpine
WORKDIR /usr/src/app
COPY package.json package-lock.json .
RUN npm ci
COPY . .
RUN npm run prepare
EXPOSE 8080
ENTRYPOINT ["npm", "start"]
```

So the developer is expected to define `prepare` and `start` in their package.json. `prepare` is supposed to build their code (because they may be using TypeScript). `start`, as the name suggests, start the server.

PR #312, opened 2019-05-21
Simplify the deploy command to deploy all items in config file only

carmenlau:


chpapa:
Did a quick hallway test with other developers to see if they get --sync's meaning or not (I don't at the first glance)

louischan-oursky:
The intention is to avoid a potential misunderstanding that `--all` is equivalent to enumerating all the name of the deployment items in the config file.

carmenlau:
Did a quick survey, most of them answer `skycli app deploy --sync` correctly (4 out of 5). But if we have concerns, maybe we just keep it simple like this:

1. `skycli app deploy` will deploy all items in `skygear.yml`, all the items that are not in `skygear.yaml` will be removed.

1. `skycli app deploy item1` will update item1 only. 

Actually in the ideal case, `skycli app deploy` should have abilities to compare the code and config with the existing app state. Only deploy the changed part to shorten the deployment time. I think  `skycli app deploy item1`  is useful only before we have this optimization.

Thought?

chpapa:
> Actually in the ideal case, `skycli app deploy` should have abilities to compare the code and config with the existing app state. Only deploy the changed part to shorten the deployment time. I think `skycli app deploy item1` is useful only before we have this optimization.

Maybe let's just keep `skycli app deploy`

Is it easy for the command to prompt the users a confirmation before it delete some deployment? (Or very difficult since it will only be known down the deployment pipeline?)

carmenlau:
@chpapa Yes, maybe let us keep it simple first. Updated the spec.

See if it looks good to you too @louischan-oursky 

Thanks:)

carmenlau:
I will merge this PR and do the simple version first, let me know if there is question

PR #314, opened 2019-05-21
Alternative to auth hooks

louischan-oursky:


features/313-alternative-to-auth-hooks/proposal.md
```diff
@@ -0,0 +1,108 @@
...
+- `before_signup_sync`
+- `before_signup_async`
+- `after_signup_sync`
+- `after_signup_async`
+
+## Usage of auth hooks
+
+Suppose we are writing an application that needs to integrate with an existing user system.
+Let's name that external user system X.
+The app has its own user system which is clearly the Auth gear. During signup, the user of
```
chpapa: "which is clearly the Auth gear"?

louischan-oursky: I mean the application is using skygear next so its auth things (signup, login, etc) are provided by the auth gear.

louischan-oursky:
> It said "alternative to auth hooks" and suggested to remove all before_sync hooks, but auth hooks have a lot of different use cases.

Agree that hooks have other use cases. So I think we still can keep `after_async` hooks.

> On the other hand, this document only covers sign up and does not cover other major use cases (not even login), so I'm not sure/confident about what are the implications.

> If we can have a map of existing auth hooks with possible use cases, vs the new functions to provide, that may help us to have more confidence in removing before hook.

My intention is to give the developer full control of implementing their customized auth flow using the API provided the auth gear.

The possible use cases of hooks I can come up with now are

- Syncing user data (`after_async`)
- Audit trail (`after_async`)
- Send welcome email after signup (`after_async`) (It can also be done without hook if the developer has control)
- **Participate in the auth operation and possibly alter it (`before_sync` and `after_sync`)**

The kind of hook use cases I want to remove is the last bolded one.

> I feel like there are a lot of API to provide and changes for Skygear to support this.

Actually the API is the same set of HTTP API currently provided the auth gear. What we need is to make it stable so the developer can depend on it.

> For example, how would social login works with the API approach if it needs to integrate with external identify provider? What about 2FA login? I don't think these are considered.

Good question! I am sorry that when I wrote this document I haven't thought about them. I am going to review the current set of API of the auth gear and try to cover all possible use cases in the document.

> What is the timeline for this? Is it urgent?

The only implemented hook is `signup`. All other hooks are not implemented yet. If we decide to reject this proposal, then we need to implement all those hooks before we launch.

chpapa:
@louischan-oursky thanks. I think for now my biggest worries is if the new approach would make it super complicated to integrate custom with built-in auth operations like social login, 2FA, passwordless login (such as magic email link), password policy, welcome email, etc.

Another set of worries is if all `before` hook could be removed, e.g. `before` hook for forgot password operation.

chpapa:
@louischan-oursky do you plan to work on it any time soon again? Should we close it first?

louischan-oursky:
I am going to close it. Instead, we should formally specify which hooks are available and what data is passed to the hook handler since it is part of the public API. The [current specification]( https://github.com/SkygearIO/features/blob/master/features/256-auth-gear-interaction-overview-in-next/api.md#execution-flow-of-hooked-cloud-function) does not do this.

Some kind of hooks are not very useful, e.g. `before_signup`. It is async and it cannot interrupt the signup process. The actual signup process could be interrupted by another sync hook. If the developer want to alter the process, they should use before-sync hook. If they want to do something after the process succeeded, they should use after-async hook.

Issue #323, opened 2019-06-18
Revisit loginIDs configuration for email verification, forgot password, welcome email handling


carmenlau:
# Description

Need to update the spec to cover following use cases:
- Update loginIDs configuration with type to achieve
     - Support email related features
     - Support loginIDs normalization in future (email, phone, case insensitive id...etc)
- Consider user have more than 2 emails
      - If user have more than 1 emails as login ids, send welcome email to all emails
      - forgot password api require email, so reset password email should be sent to that email. all loginids password should be reset?
      - Which fields require verification should be configurable under verification config. Double check current verification setting, we should allow user to verify more than 1 email fields
- See any other use cases we should cover

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
Two thoughts.

> If user have more than 1 emails as login ids, send welcome email to all emails
I think our API should make it flexible, they can send a welcome email to just one emails, specific email address, or all? I think for welcome email, most case just 1 email address is fine.

Other issues are:
* Should we have the concept of primary login ID / primary email address?

louischan-oursky:
@kiootic Please also consider the above Ben's comment in your specification.

chpapa:
> @kiootic Please also consider the above Ben's comment in your specification.

Thanks. I think it is about the PR #325? Seems already kind of considered.

(Oh but the PR didn't link back to this issue...)

louischan-oursky:
Here are some goals we need to achieve

---

1. Support implicit login ID key
As in "Login with your email, phone number or username"

1. Support login ID in a realm
Combine email and a realm to form a password principal
email: louischan@oursky.com
realm: teacher
email: louischan@oursky.com
realm: student

1. Support multiple login IDs under the same login ID key natively
So that we do not need to rely on naming convention like (`email`, `email2`, `email3`)
Note that we need to adjust the implementation of verification.

1. Support login ID normalization
    - email normalization by applying a RFC5322 addr-spec parser, ASCII lowercase the domain, ASCII lowercase the local-part
    - phone normalization by validating the phone number is in E.164 format with libphonenumber
    - unicode case folding

1. Support tagging primary email address

1. Support configuring welcome email
    - Do not send at all
[...]

Issue #296, opened 2019-03-26
Compound LoginIDs as first class citizen


kaojohnny:
# Description

In issue https://github.com/SkygearIO/features/issues/294, we suggest user to use JSON when need login/signup by compound loginIDs.

```
signupWithLoginIDs({
  email: 'example@example.com',
  role_phone: '[\"admin\",\"phone\"]' // JSON string
}, 'password')
```

In this feature spec, compound LoginIDs should be treated as the first class citizen. 

```
signupWithLoginIDs({
  email: 'example@example.com',
  role_phone: {
    role: 'admin',
    phone: '1234567'
  }
}, 'password')
```

We could support JSON or other formation at the API endpoint, and have the appropriate interface on each client SDK.

The end goal should be, people should feel comfortable to use LoginIDs to store the data, so they don't need to "sync" the data in loginIDs and other places (such as metadata)

And also maybe if we consider this feature seriously, we want to handle the case of "login by Role + Username or Role + Email" too.

tensiuyan:
from Ben:

Specification wise, can we suggest that maybe another simpler approach for compound key LoginID, is we can support string as login ID only, instead we add a few called something like domain:String, to suppose the case of "Role" + "LoginID"? It seems more straight forward and easy. Compound key could be too complicated and don't have much use case beyond a simple "Domain" + "LoginID" model.

chpapa:
Override and fixed by #325

PR #325, opened 2019-06-19
Login ID design

kiootic:
ref #323 

[Rendered Version](https://github.com/kiootic/skygear-features/blob/323-login-id/features/323-login-id/design.md)

### DIFF
--- a/features/294-login-without-specifying-login-id-key/api.md
+++ b/features/294-login-without-specifying-login-id-key/api.md
@@ -1,3 +1,5 @@
+**This document is deprecated, refer to #323 instead**
+
 For supporting login without specifying `loginID`, we have following proposal:
 
 Adjust current `LOGIN_ID_METADATA_KEYS` definition, rename it to `LOGIN_IDS_KEY_WHITELIST`. It is an empty list by default, which allows using any string as the `loginID` key, and `loginID` value must be a string.
--- a/features/323-login-id/api.md
+++ b/features/323-login-id/api.md
@@ -1,0 +1,207 @@
+# Design Summary
+
+## SDK API
+Following snippets use TypeScript:
+```diff
+- async signup(loginIDs: { [key: string]: string }, password: string, data?: object): Promise<User>;
+- async signupWithUsername(username: string, password: string, data?: object): Promise<User>;
+- async signupWithEmail(email: string, password: string, data?: object): Promise<User>;
++ async signup(
++   loginIDs: { [key: string]: string }[] | { [key: string]: string },
++   password: string, data?: object,
++   realm?: string
++ ): Promise<User>;
++ async signupWithUsername(username: string, password: string, data?: object, realm?: string): Promise<User>;
++ async signupWithEmail(email: string, password: string, data?: object, realm?: string): Promise<User>;
+
+- async login(loginID: { [key: string]: string }, password: string): Promise<User>;
+- async loginWithUsername(username: string, password: string): Promise<User>;
+- async loginWithEmail(email: string, password: string): Promise<User>;
++ async login(loginID: { [key: string]: string } | string, password: string, realm?: string): Promise<User>;
++ async loginWithUsername(username: string, password: string, realm?: string): Promise<User>;
++ async loginWithEmail(email: string, password: string, realm?: string): Promise<User>;
+
++ async requestEmailVerification(loginID: string): Promise<void>;
++ async createLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
++ async deleteLoginID(loginID: string, realm?: string): Promise<void>;
+
+  async whoami(): Promise<User>;
+  async forgotPassword(email: string): Promise<void>;
+  async verifyUser(code: string): Promise<User>;
+```
+
+## Realms
+- Each login ID can be associated with a realm.
+- By default, single realm `default` is white-listed.
+- Parameters `realm` in APIs are default to `default`.
+- (login ID, realm) is unique within an app.
+
+### Example user object
+```json
+{
+    "user_id": "5bf1e4d2-e1c4-4517-93c7-7dae89261da6",
+    "metadata": {},
+    "last_seen_at": "...",
+    "created_at": "...",
+    "created_by": "...",
+    "updated_at": "...",
+    "updated_by": "...",
+    "verified": true,
+    "verify_info": {
+        "test@example.com": true
+    },
+    "access_token": "..."
+}
+```
+
+## Configuration
+- Field `allowedRealms` (list of strings, optional).
+    - Default to single realm `default`.
+    - A list of white-list realms.
+- Field `loginIDKeys` (map of string to login ID config, optional).
+    - Default value:
+      ```json
+      {
+          "username": { "type": "raw" },
+          "email": { "type": "email" },
+          "phone": { "type": "phone" }
+      }
+      ```
+    - Keys are allowed login ID keys; value are login ID configs:
+        - `type`: type of the login ID, can be standard keys (e.g. `email`, `phone`),
+                  or `raw` for unspecified type.
+        - `minimum`: The inclusive minimum amount of login IDs, default to 0.
+        - `maximum`: The inclusive maximum amount of login IDs, default to 1.
+    - Various auth features (e.g. welcome email/user verification/user portal)
+      may make use of the login ID types.
+
+- Field field `welcomeEmail.destination` (string `first` or `all`):
+    - Default to `first`
+    - `first`: first login ID of supported (email/phone) login ID keys.
+    - `all`: all login ID of supported (email/phone) login ID keys.
+
+## Use Case Examples
+
+Unless specified, following examples assume using default configuration.
+
+### Signup with single login ID
+```typescript
+const user1 = await signupWithEmail("test@example.com", "12345678");
+const user2 = await signup({ phone: "+85299999999" }, "12345678");
+
+// following should fail:
+await signup({ fingerprint: "ZmluZ2VycHJpbnQ=" }, "12345678");
+// "login ID key is not allowed": fingerprint is not in the list of white-listed login ID keys
+```
+
+### Signup with realms
+Assume `teacher`, `student`, and `admin` are whitelisted realms.
+```typescript
+await signupWithEmail("test@example.com", "12345678", undefined, "teacher");
+await createLoginID("email", "test@example.com", "student");
+// then:
+
+// - following should succeed:
+await loginWithEmail("test@example.com", "12345678", "teacher");
+await loginWithEmail("test@example.com", "12345678", "student");
+
+// - following should fail:
+await loginWithEmail("test@example.com", "12345678");
+// "credentials are incorrect": 'default' realm is not white-listed
+await signupWithEmail("test@example.com", "12345678");
+// "realm is not allowed": 'default' realm is not white-listed
+await signupWithEmail("test@example.com", "12345678", undefined, "admin");
+// "user duplicated": cannot create new login ID for existing user (use createLoginID instead)
+```
+
+### Signup with multiple login ID
+Assume `email` and `username` are white-listed login ID keys,
+with exactly one username allowed.
+
+```typescript
+const user = await signup([
+    {"email": "test@oursky.com"},
+    {"email": "test@example.com"},
+    {"username": "test"}
+], "12345678");
+// then:
+
+// - following should succeed:
+await loginWithEmail("test@oursky.com", "12345678");
+await login("test@example.com", "12345678"); // login ID is matched literally
+await login({ "email": "test@example.com" }, "12345678"); // normalization may be applied for email login ID keys
+await loginWithUsername("test", "12345678");
+
+// - following should fail:
+await login({
+    "email": "test@oursky.com",
+    "username": "test"
+}, "12345678");
+// "multiple login ID is not allowed": cannot login with more than 1 login ID
+await signupWithEmail("test@example.com", "12345678");
+// "user duplicated": cannot use login ID of existing user
+await signup([
+    {"email": "test1@example.com"},
+    {"email": "test2@example.com"}
+], "12345678");
+// "login ID 'username' is not valid": username is not present (exactly one required)
+```
+
+### User verification
+Assume `email` and `username` are white-listed login ID keys,
+with maximum 2 emails.
+Assume verification criteria is set as `all`.
+
+```typescript
+let user = await signup([
+    {"email": "test@oursky.com"},
+    {"email": "test@example.com"},
+    {"username": "test"}
+]);
+// user.verified is false
+// user.verify_info is {}
+
+// following should fail:
+await requestEmailVerification("test1@oursky.com");
+// "invalid login ID": the login ID does not exist or does not belong to current user
+
+// following should succeed:
+await requestEmailVerification("test@oursky.com");
+
+// after verification:
+user = await whoami();
+// user.verified is false
+// user.verify_info is { "test@oursky.com": true }
+
+// following should succeed:
+await requestEmailVerification("test@example.com");
+
+// after verification:
+user = await whoami();
+// user.verified is true
+// user.verify_info is { "test@oursky.com": true, "test@example.com": true }
+```
+
+### Welcome email
+
+Assume welcome email destination is set to `first`:
+```typescript
+await signupWithEmail("test1@example.com", "12345678"); // sent to test1@example.com
+await signupWithUsername("test", "12345678") // no welcome email is sent
+await signup({
+    "email": "test2@example.com",
+    "username": "test"
+}, "12345678") // sent to test2@example.com
+await signup([
+    { "email": "test3@oursky.com" },
+    { "email": "test3@example.com" }
+], "12345678") // sent to test3@oursky.com
+```
+
+Assume welcome email destination is set to `all`:
+```typescript
+await signup([
+    { "email": "test3@oursky.com" },
+    { "email": "test3@example.com" }
+], "12345678") // sent to test3@oursky.com and test3@example.com
+```
--- a/features/323-login-id/design.md
+++ b/features/323-login-id/design.md
@@ -1,0 +1,175 @@
+**NOTE: this specification deprecates #294: Login without specifiying login ID key**
+
+# Background
+
+## Current Design
+
+Under current design of authentication:
+
+1. Each **User** is associated with some **Principal**.
+3. Each **Principal** is associated with one **Identity Provider** (e.g. password authentication, OAuth).
+4. Password authentication principals is represented by tuple (**Login ID Key**, **Login ID**, password hash).
+5. App developer can specify a list of white-listed login ID key, and all login ID key is allowed by default.
+6. To signup using password authentication, user can provide a list of tuple (login ID key, login ID). e.g. `{ "username": "test", "email": "a@example.com" }`
+7. For password authentication principals, user can login using login ID. e.g. `{ "login_id": "a@example.com" }`
+8. Login IDs are globally unique in app, without considering associated login ID keys.
+9. Auth gear does not know the semantics of login IDs.
+
+
+## Problem
+1. Some features requires semantics of login IDs to function.
+2. Some authentication system designs may not be implemented easily in current design.
+
+
+# Use cases
+
+Considering some practical scenarios of authentication system:
+
+## GitHub
+1. Supports multiple emails per user, with email verification.
+2. User can add/delete any amount of emails.
+3. Only verified emails would be associated with the user in public.
+4. User can allow recovery email/security alert to be sent to all verified emails or a specific verified email.
+
+
+## Facebook
+5. User can login with email/phone number/username, without indicating what type of value it is.
+6. User can add/delete any amount of emails, which can be used to login after verification.
+7. Disposable email address is blocked.
+8. All emails/phones of user is verified.
+9. Email communications are personalized with user language & name.
+
+## Other systems
+10. User may choose to login as different roles (e.g. teacher/student, admin/member).
+
+
+# Design Objectives
+1. Allow auth related features to understand semantics of some login IDs
+2. Support multiple login ID with same login ID key
+3. Support compound login IDs
+4. Support creating/deleting login IDs
+
+
+# Proposed Design Overview
+
+## Use Implicit Login ID Key for Logging in
+In a deprecated design (#294), user would login using just login ID. i.e:
+```json
+{
+    "login_id": "a@example.com",
+    "password": "12345678"
+}
+```
+
+This feature would be kept to support implicit login ID key.
+
+This addresses use case 5.
+
+
+## Login ID Realm
+If developer would like user to login with compound login ID, current suggested
+approach is to use serialized compound login ID as login ID value.
+i.e.:
+```json
+{ "login_id": "student:a@example.com" }
+{ "login_id": "teacher:a@example.com" }
+```
+
+This approach would not allow auth features to understand semantics of the 
+login ID component, since the format of serialization is left unspecified.
+
+To support compound login ID better, a Realm would be associated with each
+password authenticated principal.
+- Realm is a user specified value, auth gear does not concern its semantics.
+- Default realm is 'default' if unspecified, for application without need of compound login ID.
+- Possible realm values are whitelisted.
+- Realm would be considered when matching login ID with principal.
+  i.e. password authenticated principals has unique (realm, login ID).
+- API consumer could specify realm along with login ID when realm is required.
+
+This solution should support majority of use case for compound login ID.
+
+For example, when logging in with realms:
+```json
+{ "realm": "teacher", "login_id": "a@example.com", "password": "12345678" }
+{ "realm": "student", "login_id": "a@example.com", "password": "12345678" }
+```
+If authentication succeed, both request would result in same user, with
+different principals.
+
+This addresses use case 10.
+
+## Create/Delete Login ID
+In a deprecated design (#294), user can create/update/delete login IDs.
+
+Updating login ID would be problematic, since it would change the identity of
+underlying principal. Therefore updating login ID is disallowed, allowing only 
+create/delete login IDs. Instead, API consumer should:
+- Create new login ID
+- Ensure new login ID is usable (e.g. verification)
+- Delete old login ID
+
+
+## Allow Multiple Login ID per Login ID Key
+In some website, user can associate multiple emails (potentially unlimited) with
+same user. These email may be used in auth features:
+- Login with multiple emails
+- Backup email
+- Welcome email
+
+A login ID key should allow having multiple associated login ID.
+When signing up, user should provide a list of login ID. e.g.
+```json
+{
+    "login_ids": [
+        { "username": "test" },
+        { "email": "test+1@example.com" },
+        { "email": "test+2@example.com" }
+    ],
+    "password": "12345678"
+}
+```
+
+This addresses use case 6.
+
+
+## Standard Keys
+With reference to OpenID Connect standard claims, a list of standard keys
+would be specified:
+- `username`
+- `email`
+- `phone`
+- etc.
+
+These keys would be used in user metadata.
+
+
+## Configurable Login ID Keys
+The login ID keys are whitelisted by developer. In additional, developer can
+configure each login ID key:
+- The corresponding standard key (normalization may be applied)
+- The minimum/maximum amount of login ID
+
+The auth gear would use configured keys in password recovery and welcome message.
+
+
+## Email/SMS Interpretation in Auth Features
+Different auth features will interpret the email/SMS according to login ID keys
+config:
+- Signup will set initial value of user metadata with provided login IDs.
+- Welcome message will be sent to email/phone login IDs when signup up.
+- In future, CMS/Auth UI kit may use standard keys in user metadata.
+
+
+# Design Detail
+This specification would be implemented in stages:
+1. [Implicit login ID key, Login ID realm](stage-1.md)
+2. [Multiple login IDs per key, Configurable login ID keys, Standard keys](stage-2.md)
+3. [Email/SMS Interpretation in Auth Features, Create/delete login ID](stage-3.md)
+
+Design detail of each stage can be found in corresponding documents.
+
+# Appendix
+
+**OpenID Connect: Standard Claims**: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+
--- a/features/323-login-id/stage-1.md
+++ b/features/323-login-id/stage-1.md
@@ -1,0 +1,84 @@
+# Implict login ID key, Login ID realm
+
+## Database Schema
+
+- In `_auth_provider_password` table:
+    - A column `realm` (`TEXT NOT NULL`) would be added.
+    - The existing unique index on `login_id` would be changed to include `realm`.
+
+## Configuration
+
+- Add field `allowedRealms` (list of strings, optional, default to `["default"]`).
+
+## HTTP API
+- `POST /login`:
+    - Change type of `login_id` from map of strings to a plain string value.
+    - Add parameter `login_id_key` (string, optional).
+    - Add parameter `realm` (string, optional).
+    - Example:
+      ```json
+        {
+            "login_id_key": "username",
+            "login_id": "test",
+            "password": "12345678"
+        }
+        {
+            "realm": "student",
+            "login_id": "test@example.com",
+            "password": "12345678"
+        }
+      ```
+- `POST /signup`
+    - Add parameter `realm` (string, optional).
+    - Example:
+      ```json
+        {
+            "realm": "teacher",
+            "login_ids": {
+                "email": "test@example.com"
+            },
+            "password": "12345678"
+        }
+        {
+            "realm": "student",
+            "login_id": {
+                "email": "test@example.com"
+            },
+            "password": "12345678"
+        }
+      ```
+
+- `POST /signup`, `POST /login`, etc.
+    - Remove list of login IDs from response. Method to obtain this list would
+      be specified in principal APIs later.
+
+- `POST /verify_request`
+    - Rename parameter `record_key` to `login_id_key`
+    - Add parameter `login_id` (string, optional).
+    - Example: 
+      ```json
+        {
+            "login_id_key": "email",
+            "login_id":"test@example.com"
+        }
+      ```
+
+
+## Auth Logic
+- Password Provider:
+    - Require (login ID key, login ID, realm) to lookup single principal.
+    - Allow using (login ID key, login ID) to lookup multiple principals.
+- Signup:
+    - If no realm is provided, default to `default`.
+    - If realm is not in list of allowed realms, reject the request.
+    - All provided login ID will be created with same realm.
+    - If any provided login ID exists (regardless of realm), reject the request.
+      (To add a new login ID in specific realm, use create login ID API (TBD))
+- Login:
+    - If no realm is provided, default to `default`.
+    - If realm is not in list of allowed realms, reject the request.
+- User Verification:
+    - Login IDs to verified is matched by (login ID key, login ID).
+    - Realm is ignored
+    - i.e. User can verified multiple login IDs at the same time if they
+      have same (login ID key, login ID), even if realms are different.
--- a/features/323-login-id/stage-2.md
+++ b/features/323-login-id/stage-2.md
@@ -1,0 +1,101 @@
+# Multiple login IDs per key, Configurable login ID keys, Standard semantics keys
+
+## Database Schema
+
+- Update `_core_user.verify_info` column:
+    - It is a string map keyed by login ID key,
+      with verified boolean flag as value.
+    - This format is incompatible with multiple login ID per login ID key.
+    - Migrate to keyed for login ID instead.
+    - All user would be treated as NOT verified after migration, and would need
+      to be verified again.
+
+## Configuration
+- Add field `loginIDKeys` (type `LoginIDConfigMap`, optional).
+    - Default value:
+      ```json
+      {
+          "username": { "type": "raw" },
+          "email": { "type": "email" },
+          "phone": { "type": "phone" }
+      }
+      ```
+
+- Let type `LoginIDConfigMap` be a string-keyed map,
+  where keys are allowed login ID keys, with value of type `LoginIDConfig`.
+
+- Let type `LoginIDConfig` be a string-keyed map, with entries:
+    - `type`: string, required;
+              The corresponding standard key, or `raw`.
+    - `minimum`: integer, optional, default to 0;
+                 The inclusive minimum amount of login IDs.
+                 This limit will be enforced on sign up / deleting login IDs.
+    - `maximum`: integer, optional, default to 1;
+                 The inclusive maximum amount of login IDs.
+                 This limit will be enforced on sign up / creating login IDs.
+
+Example:
+```json
+{
+    "loginIDKeys": {
+        "phone": true,
+        "login_email": {
+            "type": "email",
+            "minimum": 1,
+            "maximum": 5
+        },
+        "fingerprint": {
+            "maximum": 3
+        }
+    }
+}
+```
+
+## HTTP API
+- `POST /signup`
+    - Change type of `login_ids` parameter to list of login IDs.
+    - Example:
+      ```json
+        {
+            "login_ids": [
+                { "key": "username", "value": "test" }
+            ],
+            "password": "12345678"
+        }
+        {
+            "login_ids": [
+                { "key": "email", "value": "test+1@example.com" },
+                { "key": "email", "value": "test+2@example.com" }
+            ],
+            "password": "12345678"
+        }
+      ```
+
+## Auth logic
+
+- Password Provider:
+    - Require all password authenticated principals of same user
+      share same password.
+
+- Signup:
+    - If login ID key is not in the list of allowed keys, reject the request.
+    - If the number of login IDs does not match the configuration,
+      reject the request.
+    - Create a new principal for each login ID.
+
+- User Verification:
+    - Only allow verification on login ID keys that:
+        - is configured to be verifiable login ID key
+        - is configured as a supported standard key (`email`, `phone`)
+    - Save verified login ID in as key of `_core_user.verify_info` column.
+
+- Forgot Password:
+    - Only allow password recovery email/SMS to be sent to login IDs that:
+        - is configured as a supported standard key (`email`, `phone`)
+
+
+## Appendix
+
+### Standard Keys
+- `email`
+- `phone`
--- a/features/323-login-id/stage-3.md
+++ b/features/323-login-id/stage-3.md
@@ -1,0 +1,53 @@
+# Email/SMS Interpretation in Auth Features, Create/delete login ID
+
+
+## Configuration
+- Let `WelcomeEmailDestination` be an enum:
+    - `first`: First supported login IDs of user
+    - `all`: All supported login IDs of user
+- Update field `welcomeEmail`:
+    - Add field `destination` (enum `WelcomeEmailDestination`,
+      optional, default to `first`)
+
+
+## SDK API
+Following snippets use TypeScript:
+```diff
++ async createLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
++ async deleteLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
+```
+
+This backing HTTP API of this SDK API is to be designed in another issue.
+
+## Auth logic
+
+- Signup:
+    - Let first login IDs be the first login IDs for each login ID key.
+        - For example:
+          ```json
+          [
+              { "key": "email", "value": "test+1@example.com" },
+              { "key": "email", "value": "test+2@example.com" },
+              { "key": "secondary_email", "value": "test+3@example.com" },
+              { "key": "phone", "value": "+85299999999" },
+              { "key": "username", "value": "test" }
+          ]
+          ```
+          The first login IDs are:
+          ```json
+          [
+              { "key": "email", "value": "test+1@example.com" },
+              { "key": "secondary_email", "value": "test+3@example.com" },
+              { "key": "phone", "value": "+85299999999" },
+              { "key": "username", "value": "test" }
+          ]
+          ```
+
+- Welcome Email:
+    - In context of signing up, send email to destinations specified in config:
+        - `first`: Send to first email/phone login IDs,
+                   otherwise do nothing.
+            - For above example: `test+1@example.com` and `+85299999999`
+        - `all`: Send to all email/phone login IDs,
+                 otherwise do nothing.
+            - For above example: `test+1@example.com`, `test+2@example.com`, `test+3@example.com`, and `+85299999999`


features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+## Current Design
+
+Under current design of authentication:
+
+1. Each **User** is associated with some **Principal**.
+3. Each **Principal** is associated with one **Identity Provider** (e.g. password authentication, OAuth).
+4. Password authentication principals is represented by tuple (**Login ID Key**, **Login ID**, password hash).
+5. App developer specifies a list of white-listed login ID key.
+6. To signup using password authentication, user can provide a list of tuple (login ID key, login ID). e.g. `{ "username": "test", "email": "a@example.com" }`
+7. For password authentication principals, user can login using a tuple (login ID key, login ID). e.g. `{ "email": "a@example.com" }`
```
chpapa: login ID Key is not a must. User can login only with LoginID. As specified at [spec 294](https://github.com/SkygearIO/features/blob/master/features/294-login-without-specifying-login-id-key/api.md)

features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+- User can login with email/phone number/username, without indicating what type of value it is.
+- User can add/delete any amount of emails, which can be used to login after verification.
+- Disposable email address is blocked.
+- All emails/phones of user is verified.
+- Email communications are personalized with user language & name
+
+
+# Design Objectives
+- Allow auth related features to understand semantics of some login IDs / user metadata
+- Support multiple login ID with same login ID key
```
chpapa: I'm not sure why this becomes an objective? Or you plan to make the Login ID key meaningful semantically? (E.g. a key called `email` meant all value inside are email addresses)

features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+- Support adding/removing login ID
```
chpapa: [As specified in #294](https://github.com/SkygearIO/features/blob/master/features/294-login-without-specifying-login-id-key/api.md), there are `createLoginID`, `updateLoginID` and `removeLoginID`, so this objective should be in the existing design already.

features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+# Proposed Design
+
+## Use Principals in Auth API
+When login:
+- Specify which identity provider and principal identifier is used to authenticate.
+  For password authentication, there is no need to specify login ID key (login ID is already globally unique).
+  If identity provider is not specified, password identity provider is used by default.
```
chpapa: I'm a bit skeptical at having "identity provider" as a parameter for LoginIDs (or as a login parameter in your case), actually I have thought about it when the team was discussing #294, but I didn't bring it up finally because I can't make a good trade off between complexity and the use cases.

I think it is good if you can try to incorporate it. But it will be better if you can list out the use cases and how it works.

You can reference to the [#294 spec](https://github.com/SkygearIO/features/blob/master/features/294-login-without-specifying-login-id-key/api.md), one exmaple is the support of Passcode, if we have identify provider in loginID we may get rid of passcode (Basically like API key). So starting from jwt provider and passcode, maybe you can come up with a better design.

But if we can't, I would prefer we keep LoginIDs to password authentication, and leave JWT, SSO, API Key (passcode) to other APIs.

features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+```json
+{
+    "access_token": "...",
+    "user": {},
+    "principal": { "email": "a@example.com" }
+}
+```
+
+
+## Dedicated Principal Manipulation API
```
chpapa: More details are needed here.

And it shall improve based on `create/update/removeLoginID` in [#294](https://github.com/SkygearIO/features/blob/master/features/294-login-without-specifying-login-id-key/api.md)

features/323-login-id/design.md
```diff
@@ -0,0 +1,109 @@
...
+- User may create principals with limited permission (for automatic process/etc.)
+- User may want to add/remove login emails
+
+On sign up, only single principal is allowed to be created along with the new user.
+User can add/remove principals after signing up (e.g. login email)
+
+
+## Standard Semantics for Login ID Key & User Metadata
+With reference to OpenID Connect standard claims,
+a list of standard login ID key & user metadata would be specified,
```
chpapa: I think it is similar to the idea of in future, metadata will have some standardized fields like avatar, that other part of Skygear will respect and use it semantically?

features/323-login-id/design.md
louischan-oursky: Here is the response returned by login.

```json
{
  "result": {
    "user_id": "ddaa8775-ccd3-43a3-ac9f-4362bafc4321",
    "login_ids": {
      "email": "louischan@oursky.com"
    },
    "metadata": {},
    "last_seen_at": "2019-05-31T09:43:43.863154Z",
    "created_at": "2019-05-27T08:14:55.137184Z",
    "created_by": "ddaa8775-ccd3-43a3-ac9f-4362bafc4321",
    "updated_at": "2019-05-27T08:14:55.137184Z",
    "updated_by": "ddaa8775-ccd3-43a3-ac9f-4362bafc4321",
    "verified": false,
    "verify_info": {},
    "access_token": "redacted"
  }
}
```

We can see `login_ids` is dictionary where value is string.

Suppose the developer is doing the following

```js
signupWithLoginIDs({
  email: 'johndoe@example.com',
}, 'password');
createLoginID("email", "johndoe@mycompany.com", "another_password");
createLoginID("email", "johndoe@yourcompany.com", "secret");
```

The response should look like

```json
{
  "login_ids": {
    "email": ["johndoe@example.com", "johndoe@mycompany.com", "johndoe@yourcompany.com"]
  }
}
```

Therefore, the value should really be an array if we want to support multiple login id under the same login id key.

features/323-login-id/design.md
louischan-oursky: The current way of specifying provider is actually done in endpoint level. That is, `/login` is for password authentication while `/sso/<provider>/login` for each sso provider. I doubt the benefit of generalizing all endpoints into one.

features/323-login-id/design.md
chpapa: @louischan-oursky I understood the implication.

However, we actually have two ways to implement the above. One way is as you mentioned above, have ONE `email` key with an array of login IDs.

Another way is, in current spec [294](https://github.com/SkygearIO/features/blob/master/features/294-login-without-specifying-login-id-key/api.md), it would be:
```
{
  "login_ids": {
    "email": "johndoe@example.com",
    "email2": "johndoe@mycompany.com", 
    "email3": "johndoe@yourcompany.com"
  }
}
```

The differences is really about what we meant by LoginIDs. In the 2nd approach, `email`, `email1`, `email2`, or whatever key name, is just a name for user to annotate each loginID whatever they want.

In the new way of making each login key could have multiple login IDs, the only benefits I saw is we make some convention, of specific key name imply semantic meaning of Login IDs, such as email, or phone no?

****

BTW, please don't forget the case a lot of time login IDs could be a combination of different fields. One example is in a doctor platform app Oursky developed, doctor and patient share the same system but different logins (such an email address can have two accounts, doctor or patient).

Long story short, when we discuss #294 , these are the approaches considered:

Let's say we have the following as loginIDs:
```
{ "role": string, "email": string }
```

1. User need to serialize JSON, so Skygear LoginIDs are always string (choosen solution), and we plan to follow up in #296 Compound LoginIDs as first class citizen
1. Support Compound LoginIDs natively in the APIs. (many APIs need to redesign, hence left for #296)
1. Do not support compound loginIDs, instead we just have a attributes called "Domain" for each user. This is less flexible but good enough for 90% of use case?
1. Support using a combination of metadata as login IDs (too complicated, not preferred)

So if we decide to use Login ID keys to imply the semantic meaning, it will affect how we could implement compound login IDs.

louischan-oursky: Thanks @chpapa . You helped us to discover that we are encouraging the developer to encode compound login ids into JSON while our proposal cannot handle that case.

cc @kiootic

kiootic:
@chpapa @louischan-oursky Updated, please review again.

features/294-login-without-specifying-login-id-key/api.md
```diff
@@ -1,3 +1,5 @@
+**This document is deprecated, refer to #323 instead**
```
chpapa: If you plan to mark this document at #294 is deprecated (which is a good idea), please make sure the new doc #323 is self-contained (i.e. people no need to read #294 first to understand #323 in future)

features/323-login-id/design.md
chpapa: @kiootic please update this to reflect the current design

features/323-login-id/design.md
```diff
@@ -0,0 +1,253 @@
...
+1. Allow auth related features to understand semantics of some login IDs
+2. Support multiple login ID with same login ID key
+3. Support compound login IDs
+4. Support adding/removing login ID
+
+
+# Proposed Design Overview
+
+## Use Implicit Login ID Key for Logging in
+In current design, user would login using tuple (login ID key, login ID). i.e:
```
chpapa: again.. this is wrong, the spec allow login using login ID only

features/323-login-id/design.md
```diff
@@ -0,0 +1,253 @@
...
+        { "email": "kiootic+2@oursky.com" }
+    ],
+    "password": "12345678"
+}
+```
+
+This addresses use case 6.
+
+
+## Standard Semantics Keys
```
chpapa: How does the semantics keys relating with metadata? I suppose `username`, `email`, `phone` are login ID keys and it is okay to be in `principle`. However `name`, `picture`, `locale`, `zoneinfo` etc shall be in metadata of user object, and it seems redundant to be in principle. (Or what is principle actually?)

features/323-login-id/design.md
```diff
@@ -0,0 +1,253 @@
...
+{ "realm": "teacher", "login_id": "a@example.com", "password": "12345678" }
+{ "realm": "student", "login_id": "a@example.com", "password": "12345678" }
+```
+If authentication succeed, both request would result in same user, with
+different principals.
+
+This addresses use case 10.
+
+
+## Use Principals in Auth API
```
chpapa: I would suggest we either DO NOT introduce a new terms "Principle", or give a better definition here.

(apparently, the definition of principle is some data describing the current logged in session?)

features/323-login-id/design.md
```diff
@@ -0,0 +1,253 @@
...
+- The corresponding standard semantics key (normalization would be applied)
+- The minimum/maximum amount of login ID
+
+The password identity provider would return principal metadata according to
+the configuration.
+
+
+## User Metadata
+As specified in #256, auth gear stores 'user metadata'. However this name is
+misleading: this piece of data is technically NOT metadata.
```
chpapa: I can't understand what this meant. Metadata itself just meant some data describing another object. I think it is okay to say that's metadata of user.

features/323-login-id/design.md
```diff
@@ -0,0 +1,253 @@
...
+
+The current 'user metadata' would be renamed as 'user custom data'. Auth gear
+does not concern the content of user custom data.
+
+A new 'user metadata' would be added. This metadata is a map of strings,
```
chpapa: Does it meant the "metadata" in the user object would only allow the following two keys? If yes please be specific about it.

BTW, in that case, I don't think we should define two new terminologies `metadata` and `customdata`, it is too confusing. If it is only two keys I think we can just put it in user object?

The reason is to avoid introducing too many concepts which makes understanding of Skygear difficult for new comers / future developers...

features/323-login-id/stage-1.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Database Schema
+
+- In `_auth_provider_password` table:
+    - A column `realm` (`TEXT NOT NULL`) would be added.
+    - The existing unique index on `login_id` would be changed to include `realm`.
+
+## Configuration
+
+- Add field `realmWhitelist` (list of strings, optional).
```
chpapa: I think we should be specific on weather empty string (or default) is part of the whitelist or not.

e.g. if
`realmWhitelist = ["doctor", "patient"]`

would Signup or Login without realm (or default) return error?

features/323-login-id/stage-1.md
```diff
@@ -0,0 +1,91 @@
...
+        }
+      ```
+
+
+## SDK API
+Following snippets use TypeScript:
+```diff
+- async signup(loginIDs: { [key: string]: string }, password: string, data?: unknown): Promise<User>;
+- async signupWithUsername(username: string, password: string, data?: unknown): Promise<User>;
+- async signupWithEmail(email: string, password: string, data?: unknown): Promise<User>;
```
chpapa: Consider don't remove `withUsername` and `withEmail` shortcut? Those seems make new Skygear users understand much quicker.

features/323-login-id/design.md
chpapa: btw If my understanding of "Principle" here is correct -- the current logged in session, maybe give it a name that's more self explantory?

features/323-login-id/design.md
kiootic: The spec that allows login using only login ID is #294, which is deprecated by this document. Also, that spec had not yet implemented. Therefore I refer the design implemented as 'current design', and #294 as 'previous design'.

I suppose the terms I used is confusing. I'll change 'current design' to 'previous design', and 'previous design' to 'deprecated design'.

features/323-login-id/stage-1.md
kiootic: Yes I missed the interpretation of missing whitelist, I'll fix it: it should be `(list of strings, optional, default to ["default"])`.

The signup/login without specifying realm case is specified in 'Auth Logic' section.

features/323-login-id/stage-1.md
kiootic: It's just my personal opinion that it is redundant and not provide much convenience. I'll keep it in if you prefer.

kiootic:
I think there are some terms I need to clarify:

- 'User Metadata': There exists 'user metadata' in current implementation. However, auth gear does not concern it's content nor semantics; it is same as user profile in v1, and is indeed refer as user profile in code. Therefore even if `picture`, `locale` is present in this data, auth gear cannot use it. In my opinion, calling it 'metadata' does not describe it's real purpose, that it is custom data for apps without own database.

- Proposed 'User Metadata': Different from existing 'user metadata', its content is understood and controlled by auth gear. This metadata is not used in core auth processes, so it should not be put in the core user entity.

- 'Principal': This term is well-defined in context of auth, I don't think we should change its name. This concept is also actively used within current auth gear implementation, although not readily exposed publicly. To simplify, principal is an identity of user (so we have identity providers, as described in background section).

In the old version of spec, I did want to unify the login process, but eventually took it out since it's quite out of scope for this issue. So SSO should be left untouched at the moment, and passcode is not specified yet.

The design stages should be considered and implemented sequentially. For the principals and metadata changes (stage 3), I'd like to reach a general agreement before designing in detail (since it would be controversial).

features/323-login-id/design.md
chpapa: Just avoid using terms or sentences that don't convey any useful information.

features/323-login-id/stage-1.md
chpapa: Yea should help new users to use it. It is because for new users, they would read the API docs quickly to understand what features are available, having the common features as Function Name instead of an parameter is helpful.

chpapa:
> 'User Metadata': There exists 'user metadata' in current implementation. However, auth gear does not concern it's content nor semantics; it is same as user profile in v1, and is indeed refer as user profile in code. Therefore even if `picture`, `locale` is present in this data, auth gear cannot use it. In my opinion, calling it 'metadata' does not describe it's real purpose, that it is custom data for apps without own database.

@kiootic nope, metadata was not meant to be "custom data for apps without own database". In [#256](https://github.com/SkygearIO/features/blob/master/features/256-auth-gear-interaction-overview-in-next/api.md) it was meant to be "user attributes"; So I think metadata is a pretty good term for that. (From wikipedia: Metadata is data that provides information about other data", in the case of "User Metadata", it is data that provides information about User)

The proposed use case for metadata are like:
1. Auth Gear provide JWT token with custom claim support, it can be configured to include those metadata in custom claim.
1. Auth UIKit, can use some metadata like `avatar` for some specific purpose
1. If we will have a CMS, or for Skygear's User Management Dashboard, we will treat some metadata like `avatar` in a special way such as to display it in user listing.

We would encourage users of Skygear to always save their user profile in DB, and use Auth's webhooks if they need to sync some profile data from metadata.

> Proposed 'User Metadata': Different from existing 'user metadata', its content is understood and controlled by auth gear. This metadata is not used in core auth processes, so it should not be put in the core user entity.

In your case you're proposing two keys:
- `primary_principal_id`
- `recovery_principal_ids`

The reason why I think these attributes should be in core user entity is -- for users, as long as it is a value they are not supposed to modify without interacting with auth API, they would feel natural to see it in core user entity. The definition of `used in core auth processes` is non-relevant to the user. One may not even know if they're used or not -- e.g. a user cannot distinguish weather forgot password is a "core auth processes" or not. We don't have a definition of what is "Core".

If these values could be easily / freely modify by users without using auth API, I think we can keep it in metadata along with any other values. However, I tend to think this is not the case?

Another choices is, we can give it another name in the core user entity, such as something like:
`login_ID_setting: { "primary_login_ID": "...", "recovery_login_ID": "..." }`

Speaking of the wording choice between `Login_ID` and `Principal_ID`...

> 'Principal': This term is well-defined in context of auth, I don't think we should change its name. This concept is also actively used within current auth gear implementation, although not readily exposed publicly. To simplify, principal is an identity of user (so we have identity providers, as described in background section).

That's the problem, you're proposing to expose this name publicly, so we are introducing a new concept to our APIs. But I don't see the benefits -- it seems perfectly fine if we just keep using `LoginID` or a self-explanatry name in the case of HTTP Headers.

In fact, in the PR, principal was only mentioned when explaining the concept (without a definition), but not in the APIs itself, except in the case of having a key called "Principal" in authn response and the "primary" / "recovery" `principal_id`, which seems perfectly fine if we replace it with `login_id`

Do we really need this new terms and concept?

> In the old version of spec, I did want to unify the login process, but eventually took it out since it's quite out of scope for this issue. So SSO should be left untouched at the moment, and passcode is not specified yet.

Hmmm, I think introducing the concept of `Principal` from internal to external would somehow "unify" it. But I think it would be more clear when the spec is developed.

**** 
See if the above help clarify the usage of `metadata`? And see if my arguments about lack of the definition / clear design objective of principal as part of the API make sense or you can somehow further develop it in API level?

p.s. if my memory serve me right, the use of the terms "Principal" was originate from when we develop Skygear Auth, we borrowed a lot of concept/architecture from the auth modules in Pyramid, and they have an abstraction on top of users called "Principal". (I wasn't involved with Skygear Design as heavily when the team was doing it...) I think in that context the use of "Principal" in authn response is brilliant, but unless we need this concept else where in the APIs, I think a more self-explantory (and longer) wordings just make things easier.

kiootic:
After discussing with @louischan-oursky :
- User metadata would be kept, and the proposed one would be named 'user settings'
- The principals part would be cut out for future discussion

louischan-oursky:
> User metadata would be kept, and the proposed one would be named 'user settings'

I still prefer we keep its name.

kiootic:
@louischan-oursky @chpapa Updated, please review.

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
+# Multiple login IDs per key, Configurable login ID keys, Standard semantics keys
+
+## Database Schema
+
+- Normalize `_auth_provider_password` table:
+    - Require all password authenticated principals of same user share same password.
+    - Create `_auth_provider_password_credentials` table, with columns:
```
louischan-oursky: Do not introduce this table.

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
...
+        - `user_id`, `TEXT NOT NULL`, primary key, foreign key to `_core_user.id`
+        - `password`, `TEXT NOT NULL`
+        - `verification_info`, `JSONB NOT NULL`, default to `{}` (specified below)
+    - Remove `_auth_provider_password.password` column.
```
louischan-oursky: Drop this change.

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Remove `_core_user.verify_info` column:
```
louischan-oursky: Keep this column. Only change the format for our proposed changes.

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
...
+    - Use `_auth_provider_password_credentials.verification_info` instead, it is
+      keyed by login ID instead, with verified flag as value.
+
+## Configuration
+- Add field `loginIDKeys` (map of string to `LoginIDConfigMap`, optional).
+    - Default value:
+      ```json
+      {
+          "email": { "type": "email" },
+          "username": { "type": "username" }
```
louischan-oursky: Also add `"phone": { "type": "phone" }`

louischan-oursky: `"username": { "type": "raw" }`

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
...
+- Let type `LoginIDConfigMap` be a string-keyed map,
+  where keys are login ID keys, with value:
+    - `true`: the login ID key is allowed
+    - `false`: the login ID key is disallowed
+               (default if the login ID key does not exists in the map)
+    - value of type `LoginIDConfig`: the login ID key is allowed, with
+                                     additional configurations
+
+- Let type `LoginIDConfig` be a string-keyed map, with entries:
+    - `type`: string, optional;
```
louischan-oursky: required

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,138 @@
...
+      {
+          "email": { "type": "email" },
+          "username": { "type": "username" }
+      }
+      ```
+
+- Let type `LoginIDConfigMap` be a string-keyed map,
+  where keys are login ID keys, with value:
+    - `true`: the login ID key is allowed
+    - `false`: the login ID key is disallowed
```
louischan-oursky: No boolean is allowed. The value must be an object.

features/323-login-id/stage-3.md
```diff
@@ -0,0 +1,67 @@
...
+## Configuration
+- Let `WelcomeEmailDestination` be an enum:
+    - `primary`: Primary login ID of user
+    - `all`: All supported login IDs of user
+- Update field `welcomeEmail`:
+    - Add field `destination` (enum `WelcomeEmailDestination`,
+      optional, default to `primary`)
+
+
+## HTTP API
```
louischan-oursky: Let's not introduce this API to get/update settings for now.

features/323-login-id/design.md
```diff
@@ -0,0 +1,165 @@
...
+# Background
+
+## Current Design
+
+Under current design of authentication:
+
+1. Each **User** is associated with some **Principal**.
+3. Each **Principal** is associated with one **Identity Provider** (e.g. password authentication, OAuth).
+4. Password authentication principals is represented by tuple (**Login ID Key**, **Login ID**, password hash).
+5. App developer specifies a list of white-listed login ID key.
```
chpapa: In existing design, the Login ID Key is not really whitelisted -- default it allows any login key to avoid user confusion, and only whitelisted if user give it a setting.

IF we make login ID have semantic meaning -- i.e. most of the time people are expected to only use`email`, `phone`, `username` as login ID key, I think we can change to really whitelisted by default.

features/323-login-id/stage-1.md
```diff
@@ -0,0 +1,93 @@
...
++ async signup(loginIDs: { [key: string]: string }, password: string, data: unknown, realm?: string): Promise<User>;
++ async signupWithUsername(username: string, password: string, data?: unknown, realm?: string): Promise<User>;
++ async signupWithEmail(email: string, password: string, data?: unknown, realm?: string): Promise<User>;
+
+- async login(loginID: { [key: string]: string }, password: string): Promise<User>;
+- async loginWithUsername(username: string, password: string): Promise<User>;
+- async loginWithEmail(email: string, password: string): Promise<User>;
++ async login(loginID: string, password: string, realm?: string): Promise<User>;
```
chpapa: I think we should still support login with key, (and maybe loginWithUsername and loginWithEmail too)

The reason is, as my understand part of it is for semantic meaning of loginID, it would be more convenient if the end user login with Phone No / Email, and if it is known to be phone no / email, we can do the normalization for user to match with login ID.

(e.g. normalize of country code + area code for phone number login, or postfix match if login by phone no, etc)

features/323-login-id/design.md
```diff
@@ -0,0 +1,165 @@
...
+
+# Design Detail
+This specification would be implemented in stages:
+1. [Implicit login ID key, Login ID realm](stage-1.md)
+2. [Multiple login IDs per key, Configurable login ID keys, Standard keys](stage-2.md)
+3. [User settings](stage-3.md)
+
+Design detail of each stage can be found in corresponding documents.
+
+# Appendix
```
chpapa: @kiootic I think it will help future us to understand better, if we can include a sample of user object at appendix / somewhere in this document, which include all parameters of a user object, metadata sample, and the new user settings.

features/323-login-id/stage-3.md
```diff
@@ -0,0 +1,50 @@
...
+## Auth logic
+
+- Signup:
+    - Set user's primary login ID to first login ID with key configured as
+      supported standard keys (`email`, `phone`) in `login_ids` list.
+    - If no supported login ID is found, set primary login ID to
+      first login ID in list.
+
+- Welcome Email:
+    - Send email to destinations specified in config:
```
chpapa: Maybe it can be specified by config, and if not specified, could be specified in functions (I didn't give much thought on it on use cases, so just say "maybe".... not sure)

features/323-login-id/design.md
chpapa: Especially it feels a bit confusing on how verified/disabled would work after the changes above. And which attributes go to User Setting, which goes to the user object? (i.e. why verified is in user object, while primary_login_id goes to user setting?)

features/323-login-id/stage-3.md
kiootic: This would be out of scope for this document: we will need to revisit the welcome message feature in future anyways.

features/323-login-id/stage-1.md
kiootic: Design of login ID normalization is not in the scope of this issue. Since login ID key is not used during the login process in this moment, I'll prefer to add login ID key if needed later.

chpapa: So what is the scope of this issue actually...

The objective states above are:

```
1. Allow auth related features to understand semantics of some login IDs
2. Support multiple login ID with same login ID key
3. Support compound login IDs
4. Support adding/removing login ID
```

I started to feel confused with what is this about and why we are changing many things now.

The only reason for 1 and 2 to exists, to my understanding, is to allow normalization of login ID in the case of email addresses and phone number. Otherwise we have no reason to make auth aware of the semantic of some Login IDs; and hence no resaon why we need to support multiple login ID with same login ID key.

Support compound login IDs is good, but basically it is only adding realm.

Support adding/removing login ID was well covered in the current spec, just not implemented yet.

Can we clarify what we are trying to design here and move to implementation and delivery asap?

features/323-login-id/design.md
```diff
@@ -0,0 +1,185 @@
...
+
+**OpenID Connect: Standard Claims**: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+
+## Example user object
+```diff
+    {
+        "user_id": "5bf1e4d2-e1c4-4517-93c7-7dae89261da6",
+-       "login_ids": [
```
chpapa: So we don't support multiple login IDs any more after this change?

features/323-login-id/stage-3.md
chpapa: ok. Please remove things that's out of scope. (Or make it clear it is just some "reference for future design"

kiootic: The issue want to address where to send the welcome email if user has >=2 emails, so all/primary is within scope.
However, specifying the welcome email destination as cloud function would be out of scope, since it likely involve full customization of welcome message sending logic.

features/323-login-id/stage-1.md
kiootic: Point 1 is also for user verification & forgot password: they require understanding of login ID key so that they can send messages to user.
Point 2 is to address use case 6 as mentioned in design document: allow use case such as Facebook, where user can login with any verified email associated with the user.
I'll remove point 4 for future discussion on principal APIs.

features/323-login-id/design.md
kiootic: We do support multiple login ID. However, the list will not be shown in authn responses. Instead, it should be included as part of the principal APIs later.

chpapa: What is principal APIs?

features/323-login-id/stage-1.md
chpapa: @kiootic Actually in the current design it was not needed:

1. User verification / forgot password doesn't require understanding of login ID key to send message, as long as the login ID key's type (email, or phone no) is specified in the functions, e.g. in #294 

`requestVerification(LoginIDKey: String, type: ["email", "phone"])`

2. Use case 6 doesn't need pt 2 neither, it was supposed to be done in something like these in #294. (As login ID key could be arbitrary. 

```
login_ids: {
      email: 'example@example.com'
      email2: 'example2@example.com'
    },

```

chpapa: Please do not remove #4, and supplement how that could be done instead (you can even just copy and paste the API proposed in #294 .

I don't think we should have a new specification which provide less features than the proposal it deprecated.

Especially as it would change our APIs / SDK interfaces back-and-forth.

features/323-login-id/stage-3.md
chpapa: ok. We can keep it simple first.

chpapa:
@kiootic Please try to address all outstanding comments?

features/323-login-id/stage-1.md
kiootic: - I'm strongly against that the type of login ID key is determined by client side; we should not allow too much control by client side. The same goes for arbitrary login ID key; I'd prefer a sensible default white-list of keys rather than unrestricted login ID keys.
- I'll add back create/delete login ID API, however update login ID API has to go; it makes the identity of principals mutable and is dangerous from a security point of view.

features/294-login-without-specifying-login-id-key/api.md
kiootic: This document should fully cover #294 (I'll add back the login ID APIs), except the vaguely specified designs:
- Pass codes
- SSO auto link

features/323-login-id/design.md
kiootic: I'll add another document to summarize the new APIs and logic, with sample use cases.

features/323-login-id/design.md
kiootic: I'll open issues for this later, along with some other potential issues discovered during the design.

features/323-login-id/stage-1.md
chpapa: > * I'm strongly against that the type of login ID key is determined by client side; we should not allow too much control by client side. The same goes for arbitrary login ID key; I'd prefer a sensible default white-list of keys rather than unrestricted login ID keys.

I think in this design the Login ID Key is whitelist by a server-side configuration only so that's a non-issue? (In fact at #294 there is a option to whitelist it already)



>  * I'll add back create/delete login ID API, however update login ID API has to go; it makes the identity of principals mutable and is dangerous from a security point of view.

Can you be specific on why `updateLoginID` would be dangerous even for Client side with a logged in user session? Beware it is an API to update the LoginID, not other data of the principals.

But I'm fine with ignoring the updateLoginID one, coz technically it is equivalent to create and remove (see the API changes below).

BTW, when you add back create/delete login ID, please beware that in #294 , it is expected that each login ID key would have ONE login ID value; In this spec, you make each login ID key could have multiple Login ID value, so although we can have the same function name of `create/update/removeLoginID`, the parameter probably need to be changed.

Maybe something like this:

```
createLoginID(loginIDKey: String, loginIDValue: String) // this will add a new login ID to the existing array of login ID value if it already exists 
updateLoginID(loginIDKey: String, originalLoginIDValue: String, newLoginIDValue: String) // update the existing value to the new value.
removeLoginID(loginIDKey: String, loginIDValue: String) // remove a login ID from the key's array, and remove the login ID Key itself if the array is empty
```

Haven't give any in-depth thought on this, please feel free to revise/propose an alternative based on real use cases.

features/323-login-id/design.md
chpapa: Thanks. Ideally it is in the main doc as that should be the main thing we're designing.

(In fact, features spec DO NOT always need to discuss how to implement, we mainly want to discuss how users will use it...)

features/323-login-id/design.md
chpapa: As discussed, please make your case on why a principal APIs are needed. And why it is better.

A reminder is there was an effort to try to unify password authentication / SSO / and other auth methods before, but the result is too complicated and no one really uses that API, so turns out we fallback to make it simple to use PW/SSO (and passcode was proposed in #294) in separate APIs; So the principals APIs need to take care of these too. 

I'm just guessing as I still don't know what is the scope of principals API.

chpapa: On authn responses, shall we list all loginIDs actually? If not, I think at least we need to have the login ID used?

kiootic: This is another thing we need to discuss in principals APIs (sorry for deferring many things to it): if we had `primary_login_id`, do we have `primary_oauth_id`, `primary_custom_token_id`, etc.? It's better to unify as a single `primary_principal_id` (or whatever name we use).

Also, I thought about it in more details and seems we do not need this field at the moment. I'll discuss with @louischan-oursky and see if we can just remove this field.

features/323-login-id/stage-1.md
kiootic: > I think in this design the Login ID Key is whitelist by a server-side configuration only so that's a non-issue? (In fact at #294 there is a option to whitelist it already)

Yes, I just explain why I opted a server-side config rather than doing it client-side.

> Can you be specific on why updateLoginID would be dangerous even for Client side with a logged in user session? Beware it is an API to update the LoginID, not other data of the principals.

Ideally, I'd like an authn system where if I hold a reference to a principal of user (i.e. principal ID), I can rely on it always referring the the same principal (i.e. immutable identity). However if we allow user to update login ID (i.e. changing the identity of an existing principal), this guarantee will not hold.

features/323-login-id/design.md
chpapa: Actually, my question will be what is `primary_login_id` (or `primary_principal_id`), how would it be defined OR what would it be used for?

If it is configured as the "primary email to send welcome email/forgot password etc to", or "primary SMS number to send login code to", it seems make more sense to have a primary metatag to the login ID INSIDE each specific keys.

(so I still don't see why we need `principal` or `principal_id` at all, please don't defer anything to something we probably won't have / can't make the justification to have it yet)

kiootic: The use of `primary_login_id` is specified in stage 3: It is used to determine the destination of welcome email (all email login ID/primary email login ID).

chpapa: So except primary email, we might also have primary SMS number, so we don't have one single primary login ID, but primary login ID for each login ID Key type as described above?

kiootic: I didn't consider the multiple primary login ID for email & SMS case. I'll think about it.

features/323-login-id/stage-1.md
chpapa: > Yes, I just explain why I opted a server-side config rather than doing it client-side.

ic, but it comes out of nowhere, please start a different thread if it is a new discussion. So the discussion is traceable.

> Ideally, I'd like an authn system where if I hold a reference to a principal of user (i.e. principal ID), I can rely on it always referring the the same principal (i.e. immutable identity). However if we allow user to update login ID (i.e. changing the identity of an existing principal), this guarantee will not hold.

ic, but we don't have principals now. Instead, shall we have some data about the current session in authn response? No matter it is by password or other providers.

chpapa: Actually this thread is getting too long, let me resolve it and start a new thread. thanks.

features/323-login-id/design.md
chpapa: Notes from the other thread: please include createLoginID / deleteLoginID in it.

features/323-login-id/design.md
```diff
@@ -0,0 +1,184 @@
...
+        "metadata": {},
+        "last_seen_at": "...",
+        "created_at": "...",
+        "created_by": "...",
+        "updated_at": "...",
+        "updated_by": "...",
+        "verified": false,
+        "verify_info": {},
+        "access_token": "..."
+    }
```
chpapa: Shall we have some data about the current session in authn response? No matter it is by password or other providers. It seems the user will need to know some information about current session? (Or not?)

kiootic: For common use cases, there should be no need to know which login ID is used to login.

kiootic:
@chpapa @louischan-oursky Updated, please check.

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+- async loginWithEmail(email: string, password: string): Promise<User>;
++ async login(loginID: { [key: string]: string } | string, password: string, realm?: string): Promise<User>;
++ async loginWithUsername(username: string, password: string, realm?: string): Promise<User>;
++ async loginWithEmail(email: string, password: string, realm?: string): Promise<User>;
+
++ async requestVerification(loginIDKey: string, loginID: string): Promise<void>;
++ async createLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
++ async deleteLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
+
+  async whoami(): Promise<User>;
```
chpapa: As this is PR for #323 , should have email related features's API here too i.e. ForgotPassword, although it shouldn't need much change....

(Besides SDK APIs on client side, if you think there are interfaces on server-side APIs, HTTP response header etc, you can also put it under the "API summary", just FYI)

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+Assume `email` and `username` are white-listed login ID keys,
+with exactly one username allowed.
+
+```typescript
+const user = await signup([
+    {"email": "test@oursky.com"},
+    {"email": "test@example.com"},
+    {"username": "test"}
+], "12345678");
+// user.metadata is { "email": "test@oursky.com", "username": "test" }
```
chpapa: I think in existing design, loginID are not automatically added to `user.metadata`, is it changed or my understanding was incorrect?

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+// - following should succeed:
+await loginWithEmail("test@oursky.com", "12345678");
+await login("test@example.com", "12345678");
+await loginWithUsername("test", "12345678");
+
+// - following should fail:
+await login({
+    "email": "test@oursky.com",
+    "username": "test"
+}, "12345678"); // cannot login with more than 2 login ID
```
chpapa: more than 1

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+await signupWithEmail("test@example.com", "12345678"); // cannot reuse login ID
+await signup([
+    {"email": "test1@example.com"},
+    {"email": "test2@example.com"}
+], "12345678"); // username is not present
```
chpapa: by default username could be not present? This error only present when `username` was configured with `minimum` to 1?

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+// following should fail:
+await requestVerification("username", "test"); // unsupported login ID key
+await requestVerification("email", "test1@oursky.com"); // login ID does not belong to current user
+
+// following should succeed:
+await requestVerification("email", "test@oursky.com");
+
+// after verification:
+user = await whoami();
+// user.verified is false
```
chpapa: I think it is more common use case for at least ONE loginID is verified, the account is considered verified? (We can provide a configuration, but I think the default should be ONE or more login ID verified => verified is true.

features/323-login-id/stage-1.md
```diff
@@ -0,0 +1,103 @@
...
+    - Example: 
+      ```json
+        {
+            "login_id_key": "email",
+            "login_id":"test@example.com"
+        }
+      ```
+
+
+## SDK API
```
chpapa: Try to avoid repeating the content to create maintainability troubles, my 2 cents.

features/323-login-id/api.md
kiootic: By default there can be 0 - 1 username. The section is has exactly one user name configured as mentioned at top of section, to demonstrate validation of login ID amount.

features/323-login-id/api.md
kiootic: Yes, it is changed to avoid the concept of persistent 'primary login ID'; please refer to stage 3 for detail.

features/323-login-id/api.md
kiootic: There is a configuration for 'any'/'all' emails. However, there is no default value. I'll set it to 'any' if you prefer.

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
+// then:
+
+// - following should succeed:
+await loginWithEmail("test@example.com", "12345678", "teacher");
+await loginWithEmail("test@example.com", "12345678", "student");
+
+// - following should fail:
+await loginWithEmail("test@example.com", "12345678"); // realm not matched
+await signupWithEmail("test@example.com", "12345678", undefined, "admin"); // cannot create login ID (use createLoginID instead)
+await signupWithEmail("test@example.com", "12345678"); // cannot reuse login ID for new user
```
louischan-oursky: In this case, the error should be `realm not matched` since the function invocation does not specify `realm`.

Here are my expectation:

```typescript
await signupWithEmail("test@example.com", "12345678", undefined, "admin"); // "admin" is not a whitelisted realm.
await signupWithEmail("test@example", "12345678") // the default realm "default" is not a whitelisted realm.
await signupWithEmail("test@example", "12345678", undefined, "teacher"); // duplicate login ID for new user
```

features/323-login-id/api.md
```diff
@@ -0,0 +1,170 @@
...
++   password: string, data?: object,
++   realm?: string
++ ): Promise<User>;
++ async signupWithUsername(username: string, password: string, data?: object, realm?: string): Promise<User>;
++ async signupWithEmail(email: string, password: string, data?: object, realm?: string): Promise<User>;
+
+- async login(loginID: { [key: string]: string }, password: string): Promise<User>;
+- async loginWithUsername(username: string, password: string): Promise<User>;
+- async loginWithEmail(email: string, password: string): Promise<User>;
++ async login(loginID: { [key: string]: string } | string, password: string, realm?: string): Promise<User>;
```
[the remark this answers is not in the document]
kiootic: Normalization on input login ID may be applied, so:
- Provided login ID key: normalization may be applied based on the key
- Not provided login ID key: login ID is matched literally
I'll update documentation to reflect this.

features/323-login-id/api.md
kiootic: I intend the comment to be an explanation on why should it fail. I'll define error messages in the spec if you prefer.

features/323-login-id/api.md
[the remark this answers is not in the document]
louischan-oursky: So `/login` should optionally accept `login_id_key` as argument. The normalization should be performed at server-side.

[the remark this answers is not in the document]
kiootic: Yes it is accepting an optional `login_id_key`.

kiootic:
@chpapa @louischan-oursky updated, please check again.

features/323-login-id/api.md
```diff
@@ -0,0 +1,182 @@
...
++ async loginWithUsername(username: string, password: string, realm?: string): Promise<User>;
++ async loginWithEmail(email: string, password: string, realm?: string): Promise<User>;
+
++ async requestVerification(loginIDKey: string, loginID: string): Promise<void>;
```
chpapa: @kiootic Sorry I overlooked it before, shall we have:

`requestEmailVerification(loginID: string)` instead?

Reasons are:
1. In the app config, we already have semantic meaning of loginIDKey, so when it is about email addresses for example, we know where to look for and verify the email address is a valid login ID for the users, we don't need the developers to specify the loginIDKey again (and loginID is unique anyway)
2. We have different types of verification (at least email, SMS and shortcode have been thought of), so it is better to make it explicit on function name for which type of verification we are talking about.

features/323-login-id/api.md
```diff
@@ -0,0 +1,182 @@
...
++ async createLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
++ async deleteLoginID(loginIDKey: string, loginID: string, realm?: string): Promise<void>;
```
chpapa: similarly, since loginID is unique, `loginIDKey` is not needed?

features/323-login-id/api.md
chpapa: Okay. I consider stage1-3.md are about implementation details, not part of the spec. For the sack of convinence for reference in future, can we centralize all spec (user's interface related) in this markdown?

I'm a bit skeptical (or worried) about auto-populating metadata with login ID actually... we can try, not sure how users would react.

chpapa: But we need to decide if we still want to auto-populating metadata before public launch. @kiootic can you put up an issue about it and mark it in the launch milestone? Something like: "Decide if Skygear Auth should autopopulate metadata with loginIDs"

features/323-login-id/api.md
```diff
@@ -0,0 +1,182 @@
+# API Summary
+
+## SDK API
```
chpapa: I think we need an API to set primary login ID?

Primary ID now is based on the array inside loginIDkey, but we have no way to manipulate that.

Can we add a short section here about what is primary email/sms ID too? So what it is for, and how to set the primary login ID.

features/323-login-id/api.md
chpapa: I think it needs a default value? If it is not any / all not sure what the verified meant :)

yea I prefer any, but that's because it seems more common? (Most services I used, if you have ONE email address verified, you can act as a verified user)

features/323-login-id/stage-2.md
```diff
@@ -0,0 +1,101 @@
...
+- Update `_core_user.verify_info` column:
+    - It is a string map keyed by login ID key,
+      with verified boolean flag as value.
+    - This format is incompatible with multiple login ID per login ID key.
+    - Migrate to keyed for login ID instead.
+    - All user would be treated as NOT verified after migration, and would need
+      to be verified again.
+
+## Configuration
+- Add field `loginIDKeys` (map of string to `LoginIDConfigMap`, optional).
```
chpapa: Does it allows something like this?

```json
{
    "email": {"type": "email"},
    "email-addr": {"type": "email"},
}
```

If yes, it would violate the primary ID spec below. If not, we should make it clear that each type could only be assigned to one key.

chpapa: (Okay, but if each type could only be assigned to one key, why not we just enforce it must be called email, phone and username lol) @louischan-oursky @kiootic

features/323-login-id/api.md
kiootic: As mentioned in stage 3 spec, 'primary login ID' is a concept only existing during the signup process; it is not persisted. It is just a term (like a variable name) used to refer to the welcome email destinations.

features/323-login-id/stage-2.md
kiootic: Multiple key per type should be supported, I'll clarify how it interact with the welcome email feature.

features/323-login-id/api.md
chpapa: Hm okay, so it is only for welcome email.

features/323-login-id/api.md
kiootic: ref #331

chpapa:
@louischan-oursky feel free to merge it for @kiootic if you have no other comments (and confirmed the verification criteria as mentioned in comment was addressed else where)

features/323-login-id/api.md
```diff
@@ -0,0 +1,207 @@
...
+    {"email": "test1@example.com"},
+    {"email": "test2@example.com"}
+], "12345678");
+// "login ID 'username' is not valid": username is not present (exactly one required)
+```
+
+### User verification
+Assume `email` and `username` are white-listed login ID keys,
+with maximum 2 emails.
+Assume verification criteria is set as `all`.
```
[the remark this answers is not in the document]
kiootic: The criteria config is already implemented, but default value is missing. I think this should be specified in another issue for the user verification feature.

PR #326, opened 2019-06-21
Revamp tenant config

louischan-oursky:
[Rendered Version](https://github.com/louischan-oursky/features/blob/config/features/324-config/index.md)


features/270-next-skycli/commands.md
```diff
@@ -170,48 +170,46 @@ the cluster instead.
...
-  ALLOWED_CALLBACK_URLS:
+$ skycli app view-user-config
+masterkey: xxxxx
+apikey: xxxxx
+sso:
+  allowedcallbackurls:
   - http://127.0.0.1:5000/
   - http://localhost:5000/
-  MERGE_USERS_WITH_IDENTICAL_EMAIL: true
+  mergeuserswithidenticalemail: true
```
kiootic: We would allow camelCase right?

features/324-config/index.md
```diff
@@ -0,0 +1,56 @@
...
+- `https://tag1.myapp.skygearapis.com`
+- `https://tag2.myapp.skygearapis.com`
+
+If the signup request is `https://tag1.myapp.skygearapis.com/_auth/signup`, the hooks configuration injected by the gatway for the auth gear should be specific to the `tag1`.
+
+## Proposed solution
+
+The new TenantConfiguration is solely for gear consumption. When it is injected by the gateway, it is msgpack serialized. It can also be deserialized from environment variable in case the gear is running in standalone mode.
+
+The old TenantConfiguration will be splitted into AppConfiguration and UserConfiguration. AppConfiguration is used to store non-editable configuration while UserConfiguration is intended to be edited by the developer.
```
kiootic: I suppose user configuration would be associated with a deployment, instead of an app?

louischan-oursky: The current design is associated with the app.

features/270-next-skycli/commands.md
louischan-oursky: No. If we allow camelCase then the developer can write `foobar` and `fooBar` in YAML legally but they conflict when we lowercase them.

features/324-config/index.md
kiootic: How would the solution address the 'Stored deployment-specific data' problem you mentioned?

features/270-next-skycli/commands.md
kiootic: I don't think it's a good UX; it is hard for human to understand and does not look good. Can we validate key uniqueness after normalization instead?

features/324-config/index.md
louischan-oursky: The skycli command `skycli app update-user-config` does not take deployment id as argument so logically the config should apply to the app, not a specific deployment. However, hooks are configured in `skygear.yaml`, so they must be deployment-specific.

features/270-next-skycli/commands.md
louischan-oursky: After a more throughout consideration, I am going to drop support for configuring via environment variable due to the following reason. It is hard to encode dictionary and list in environment variable, especially when the element is itself a struct. Instead, I am going to support configuring standalone mode via a YAML file.

chpapa: So rest of the spec will change back to _ separator?

chpapa:
Next step is probably to actually apply the principles to the new tenant config spec? (Give a complete sample spec / do the code changes and copy/paste the samples? @louischan-oursky

louischan-oursky:
> I think except the config name in underscore, it is fine.

`skygear.yaml` (the deployment config file read by skycli) is using `snake_case` so we better stick with `snake_case`.

> Next step is probably to actually apply the principles to the new tenant config spec? (Give a complete sample spec / do the code changes and copy/paste the samples?

I am implementing this feature and will post sample of how the YAML look like.

louischan-oursky:
@chpapa I have added [examples](https://github.com/louischan-oursky/features/blob/config/features/324-config/index.md#appendix) of the new config.

features/270-next-skycli/commands.md
kiootic: @louischan-oursky The implemented config has _ separators. Can you update the config key name to the correct format?

louischan-oursky: Updated.

Issue #329, opened 2019-06-26
Identity APIs


kiootic:
# Description

Main objective: Ensure all identity providers (e.g. password, oauth, custom token) are well-supported on API level.

- Consistent API on the principal associated with current session (e.g. `current_login_id`, `current_oauth_id`, `current_custom_token_id`)
- Consistent API on getting list of associated principals (e.g. getting login IDs, linked OAuth accounts, associated custom tokens)
- Consistent API on creating/deleting principals (e.g. create/delete login IDs, link/unlink OAuth accounts)
- Specify the consistent method to obtain metadata about the principal for internal/external usage (e.g. sending welcome email to OAuth accounts, synchronizing user metadata from external profiles)

# Related Issues

- Spec PRs
  - [x] https://github.com/SkygearIO/features/pull/334
- Server Issues
  - [x] https://github.com/SkygearIO/skygear-server/issues/974
- Client Issues
  - [ ] https://github.com/SkygearIO/skygear-SDK-JS/pull/553
- Guides Issues
  - TODO

chpapa:
For the record, the idea of a Principal APIs is highly debatable and unless there are very clear arguments, demonstrated by sample code and API spec that it is better than existing APIs interfaces, I think we shouldn't assume this will be implemented.

PR #334, opened 2019-07-03
Identity API design

kiootic:
ref #329 

[Design](https://github.com/kiootic/skygear-features/blob/329-principal-api/features/329-identity-apis/design.md)

[API](https://github.com/kiootic/skygear-features/blob/329-principal-api/features/329-identity-apis/api.md)


features/329-principal-apis/design.md
```diff
@@ -0,0 +1,100 @@
...
+# Glossary
+
+**Identity Provider**:
+A method to assert the identity of user. For example, providing a password or
+perform OAuth flow.
+
+**Principal**:
+An identity associated with user. A user can have multiple principals. Since
+the term 'Principal' is defined differently across many frameworks and
+standards, we will instead use a simpler term 'Identity' for public interfaces.
```
chpapa: Please rename #329 and #334 from `Principal` to `Identity` for consistency than.

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+# Proposed Design
```
chpapa: Can you add the function names related with each section/features below? Would make it easier to references with the api.md thanks!

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Deleting Principal
+We need to provide a way for developer to delete new principal for the user:
+- Remove login ID
+- Unlink OAuth account
+
+Each identity provider should have their specific API to delete principal. APIs
+that delete principals should have common behaviors:
+- Should respect re-authentication configuration
+- Should not allow a verified user to become unverified through the operation
```
chpapa: May you clarify why this behavior is desired?

What if in a case:

1. User A signed up with `a@a.com`
1. User A verified with `a@a.com`
1. User A linked with `b@b.com`
1. User A logged in with `b@b.com` again
1. User A deleted the email address of `a@a.com`

In the case above, user should become unverified? 

Or you intend to enforce the behavior of before `a@a.com` is deleted, `b@b.com` must have been verified? (We need to consider this seriously and what are the implications)

features/329-principal-apis/api.md
```diff
@@ -0,0 +1,270 @@
...
+async function addLoginID(loginID: {[key: string]: string}, realm?: string): Promise<void>;
+async function removeLoginID(loginID: string, realm?: string): Promise<void>;
+
+async function changePassword(newPassword: string, oldPassword?: string): Promise<User>;
+
+```
+
+
+## Configuration
+`reauthForSecurity` (boolean; default `true`):
```
chpapa: LOVE that we have these configurations! Great idea @kiootic

features/329-principal-apis/api.md
```diff
@@ -0,0 +1,270 @@
...
+        metadata: {
+            "email": "test@example.com"
+        }
+    },
+]);
+
+await addLoginID("secondary_email", "test@oursky.com");
+// throws error "access token is not issued recently"
+
+// after re-authentication:
```
chpapa: Which api the user should use for re-authentication? It might be tricky for users to reconstruct the login API from the current identity, and worth us to provide a shortcut API at client side.

kiootic: Just the normal login API. I expect the flow would be similar to Google's change password: user is prompt to login again when attempting to change password, and user can enter just the new password after logging in.

A dedicated API cannot be provided. User may be logged in using password/OAuth; it is up to app developer to provide suitable UI for re-authentication. App developer can get the way user logged in through current identity.

features/329-principal-apis/design.md
kiootic: It’s just my feeling that it make sense for most app. We can remove this for now and discuss when we specify user verification feature in #336.

features/329-principal-apis/api.md
```diff
@@ -0,0 +1,270 @@
...
+    };
+}
+interface CustomTokenIdentity extends IdentityBase {
+    type: 'custom';
+    providerUserID: string;
+    metadata: {
+        raw_profile: object;
+    };
+}
+type Identity = PasswordIdentity | OAuthIdentity | CustomTokenIdentity;
```
louischan-oursky: - Remove `IdentityBase`
- Move `raw_profile` to top-level
- For OAuth `metadata: { email?: string }`
- For Custom Token `metadata: { email?: string }`
- For Password, `metadata: { email?: string, phone?: string }`

Specify `metadata` in Identity is always derived. Specify how it is derived for each Identity.

features/329-principal-apis/api.md
chpapa: Right, that should be a feature when we start develop a AuthUI Kit.

features/329-principal-apis/design.md
chpapa: Sure. Let's postpone it to later / with more real world use cases.

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,100 @@
...
+- Should not allow deleting the current principal
```
chpapa: Doesn't seems we should have this behaviour?

I'm thinking about the following use cases:

0. Context: A simple system that allow ONE login ID per user only
1. User `loginIDA` logged in, and want to change it's username
2. Since it can't rename a Login ID, the user need to delete the old login ID and add a new one.
3. Since it does not allow deleting the current principal, it also can't delete the old Login ID.
4. The only way to implement this is add the new username as new login ID, ask the user to login again, and delete the old login ID automatically.

The above behaviour seems to contradict with most "change my username"'s operation.

So either:
1. We allow rename of loginID
2. We allow delete of current principal.

kiootic: Both option is undesirable:
- allow renaming: we cannot ensure a principal is always referring the same identity
- allow deleting current principal: we cannot return a meaning current principal after the operation.

Instead, I'll add a `replaceLoginID` API, and practically it performs the suggested procedure (i.e. add new login ID then delete old login ID) atomically with necessary checks.

chpapa: Any reason why we should call it `replaceLoginID` instead of `renameLoginID` or `changeLoginID`? From usability perspective we should limits the usage of verb, so better check other verbs used in auth API and pick the common one. (We used to have `change` and `update` in different places and it was very confusing for users)

features/329-principal-apis/api.md
```diff
@@ -267,4 +292,56 @@ expectEquals(identities, [
...
+    id: "1431CB1E-8A2F-4A44-874E-70C3DB3EE043",
+    type: "password",
+    loginIDKey: "username",
+    loginID: "test1",
+    realm: "default",
+    metadata: {}
+});
+
+// after re-authentication (otherwise the call would fail):
+await replaceLoginID("test1", {"username": "test2"});
```
chpapa: If we care about security, maybe we should include in this spec, that we will keep an audit trail of updating / adding / delete of login ID? It is a common practice for system with security requirements.

(Similar measure goes to username, but that would be another features from loginID, it is even more common for username to have some notes like "last changed N months ago" etc)

features/329-principal-apis/design.md
kiootic: 'rename' or 'change' implies a mutable principal, which is not the case; this API creates a new principal and delete the old one as an atomic operation.

features/329-principal-apis/api.md
kiootic: Yes I agree all security critical operations, including adding/removing login ID and linking/unlinking OAuth account, should produce audit trails. However, I don't see this feature specified for version next, and the existing spec for v1 seems to be not detailed enough. Would we list all events we track when we specify audit trail in future?

chpapa: Maybe can help create a feature issue about user operation audit, and list current known operations like the above there first?

chpapa: @kiootic pls mark it as resolve when the issue mentioned is created

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,136 @@
...
+
+For change password, we support the common case of providing old password to
+implicitly authenticate the user.
+
+Both the re-auth requirement and time to considered 'recent' can be configured.
+Developer can turn off the re-auth requirement in configuration. However,
+developer should understand the potential security risk.
+
+> Related config: `reauthForSecurity` and `reauthIntervalSeconds`
+> Related API: `changePassword` function
```
chpapa: add/remove login id, link/unlink sso provider?

features/329-principal-apis/design.md
chpapa: Further comments below, resolve this first.

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Current Principal
```
chpapa: :s/Principal/Identity/g

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,136 @@
...
+- Which login ID is used to login?
+- Which OAuth provider is used to login?
+
+When logging in using a principal, the ID of the principal (i.e. the current
+principal) would be embedded in the access token along with the user ID.
+Information about the current principal would be returned in user object.
+
+> Related API: `identity` field in `User` object
+
+## Principal List
```
chpapa: :s/Principal/Identity/g

features/329-principal-apis/design.md
```diff
@@ -0,0 +1,136 @@
...
+- Link OAuth account
+
+Each identity provider should have their specific API to create principal. APIs
+that create principals should have common behaviors:
+- Should respect re-authentication configuration
+
+When signup as with multiple login IDs, the first login ID will be used as
+current identity in the newly created session.
+
+> Related API: `addLoginID` function for password identity provider
```
chpapa: link OAuth account?

features/329-principal-apis/design.md
chpapa: I think consistency of verbs is more important, can we look at the list of verbs we use and use a common one for renaming of LoginID?

Actually I'm not sure what mutable meant here. In the above cases, if a logged in user changed the username (login ID), its identity's login ID is updated already? Even if you create a new auth token, for the end user it is still just like the identity's parameter was changed.

chpapa:
@kiootic please try to address all outstanding comments so we can consider this spec mergable asap?

features/329-principal-apis/api.md
kiootic: ref #340

features/329-principal-apis/design.md
kiootic: In this document, we use `Principal`. `Identity` is used for public interface, i.e. in SDK/HTTP API or config.

features/329-principal-apis/design.md
kiootic: API for SSO is specified in another document, so it is not mentioned here.

features/329-principal-apis/design.md
kiootic: API for SSO is specified in another document, so it is not mentioned here.

features/329-principal-apis/design.md
kiootic: It is not renaming login ID, it is replacing old login ID with new login ID. The ID of the current principal will be different.

chpapa: So it is still not `replaceLoginID`, it could be `replaceIdentity` but then it is very difficult to understand for newcomers.

For consistency and usability reason, it is better for us to follow the verbs of similar APIs in our platform like `changeLoginID` or `updateLoginID` or whatever we have been using for consistency.

features/329-principal-apis/design.md
chpapa: However it is mentioned in the paragraph above; I assume if "changePassword" is related APIs, do you meant any APIs that need to be reauthenticated? That seems include link/unlink SSO and add/remove Login ID? In that case why `changePassword` is listed as related APIs but not the others?

I have written another comment on review about the difficult to read reasons and recommendation.

features/329-principal-apis/design.md
chpapa: Please only use Principal when appropriate and needed. You're describing a public interface here.

features/329-principal-apis/design.md
chpapa: Similar problem with the above, what is this design document about actually?

features/329-principal-apis/design.md
kiootic: I'll name it `updateLoginID`. However, the documentation would need to note that the ID of identities would be changed; most user would not expect an 'update' operation would change the ID of an entity.

chpapa: You can note it as a side effect in spec, and we can mention it in documentations.

chpapa: Actually, what are the use cases for exposing the id of Identity?

features/329-principal-apis/api.md
```diff
@@ -0,0 +1,363 @@
...
+    lastLoginAt: Date;
+    isVerified: boolean;
+    isDisabled: boolean;
+    metadata: Metadata;
+
+    identity: Identity;
+}
+
+interface PasswordIdentity {
+    id: string;
```
chpapa: actually, what is the use case for exposing the ID of Identity?

kiootic:
@chpapa @louischan-oursky updated, please check again.

chpapa:
@kiootic thanks for the work! I took a brief look, I think should be okay (ok I admit I'm getting a bit lazy here...) 

Let's move forward and get this thing out in real life testing to see anything need to be changed :)

I hope you found the journey of writing spec interesting.

Actually last year we discussed a new way to do new feature design / changes to Skygear, we can discuss it together later.

chpapa:
@louischan-oursky given I didn't take a look in details, can you help read the latest in details just to make sure everything make sense?

features/329-identity-apis/api.md
```diff
@@ -0,0 +1,363 @@
...
+security-critical operations:
+- add/remove login ID
+- link/unlink SSO provider (except auto-link when logging in with OAuth)
+- change password (if old password is not provided)
+
+`reauthIntervalSeconds` (integer; default 300, i.e. 5 minutes):
+The period after the access token is issued, that is considered authenticated
+for security-critical operations.
+
+`updateLoginIDEnabled` (boolean; default `false`):
```
chpapa: I think this should default to `true`? Any reason why we need to disable it actually?

Reason is allow changing of username seems a basic enough use case? Or there are special reason why this feature is problematic? (As it should be obvious that we don't want to have feature flag for users to enable/disable each and every features)

features/329-identity-apis/design.md
```diff
@@ -0,0 +1,134 @@
...
+recent.
+
+For change password, we support the common case of providing old password to
+implicitly authenticate the user.
+
+Both the re-auth requirement and time to considered 'recent' can be configured.
+Developer can turn off the re-auth requirement in configuration. However,
+developer should understand the potential security risk.
+
+> Related config: `reauthForSecurity` and `reauthIntervalSeconds`
```
chpapa: `disableReauthForSecurity` instead?

features/329-identity-apis/api.md
kiootic: This feature has some important implications: many users & developers expect the user name would not be changed, so it requires much consideration on UX (e.g. user profile link slug, display name impersonation).

chpapa: Agree we should have this option. Not sure if we should have default to true or false. Let's stick with it first than if you guys think false is better.

features/329-identity-apis/api.md
```diff
@@ -0,0 +1,363 @@
...
+        realm: "default",
+        claims: {}
+    },
+    {
+        id: "E65CA45C-7ADF-4F37-AF8C-1F99D526D6F1",
+        type: "oauth",
+        providerID: "some-site",
+        providerUserID: "9999999999999999",
+        claims: {
+            raw_profile: {
```
louischan-oursky: This should be moved 1 level up.

features/329-identity-apis/api.md
```diff
@@ -0,0 +1,363 @@
...
+let identities: Identity[];
+identities = await listIdentities();
+expectEquals(identities, [
+    {
+        id: "1C1340E0-2A4E-44FA-AFC1-F8EF01297DF6",
+        type: "oauth",
+        providerID: "some-site",
+        providerUserID: "9999999999999999",
+        claims: {
+            raw_profile: {
```
louischan-oursky: 1 level up.

features/329-identity-apis/api.md
```diff
@@ -0,0 +1,363 @@
...
+
+identities = await listIdentities();
+expectEquals(identities, [
+    {
+        id: "1C1340E0-2A4E-44FA-AFC1-F8EF01297DF6",
+        type: "oauth",
+        providerID: "some-site",
+        providerUserID: "9999999999999999",
+        claims: {
+            raw_profile: {
```
louischan-oursky: 1 level up.

features/329-identity-apis/design.md
```diff
@@ -0,0 +1,134 @@
...
+# Proposed Design
+
+## Re-authentication for Security
+Many identity manipulation operations are critical for security:
+- Add/remove/update login ID
+- Link/unlink SSO provider
+- Change password
+
+By default, user are required to re-authenticate before performing these
+operations. This is achieve by checking the issue time of the access token is
```
louischan-oursky: s/achieve/achieved

features/329-identity-apis/design.md
louischan-oursky:
```YAML
reauthentication:
  # zero value is false so enabled by default
  disabled: false
  # zero value to be converted to default value
  interval: 30
```

features/329-identity-apis/design.md
```diff
@@ -0,0 +1,134 @@
...
+- Which OAuth provider is used to login?
+
+When logging in using a identity, the ID of the identity (i.e. the current
+identity) would be embedded in the access token along with the user ID.
+Information about the current identity would be returned in user object.
+
+> Related API: `identity` field in `User` object
+
+## Listing Identity
+We need to provide a way for developer to get a list of identitys belonged to
```
louischan-oursky: s/list of identitys belonged to the user/list of identities of the user

features/329-identity-apis/design.md
```diff
@@ -0,0 +1,134 @@
...
+the user:
+- What login IDs the user has?
+- Which OAuth providers had the user linked?
+
+A list of identitys belonged to current user would be returned through an API.
```
louischan-oursky: s/list of identitys belonged to the user/list of identities of the user

features/329-identity-apis/design.md
```diff
@@ -0,0 +1,134 @@
...
+that create identitys should have common behaviors:
+- Should respect re-authentication configuration
+
+When signup as with multiple login IDs, the first login ID will be used as
+current identity in the newly created session.
+
+> Related API: `addLoginID` function, `linkOAuthProviderWithPopup/Redirect` function (#332)
+
+## Deleting Identity
+We need to provide a way for developer to delete new identity for the user:
```
louischan-oursky: s/new/an existing

kiootic:
@louischan-oursky updated, please check.

features/329-identity-apis/api.md
```diff
@@ -59,6 +57,19 @@ async function updateLoginID(oldLoginID: string, newLoginID: { [key: string]: st
 
 async function changePassword(newPassword: string, oldPassword?: string): Promise<User>;
 
+class AuthContainer {
+    // NOTE:
+    // Access token should be persist more securely than common auth data.
+    // Access token is separated from user object, so developer can save user to
+    // app state storage (e.g. redux store) directly.
+    get currentAccessToken(): string;
```
louischan-oursky: `get currentAccessToken(): string | null | undefined;`

features/329-identity-apis/api.md
```diff
@@ -59,6 +57,19 @@ async function updateLoginID(oldLoginID: string, newLoginID: { [key: string]: st
...
+    // NOTE:
+    // this field is not contained in user object to be consistent with access
+    // token: user has many identity, identity has many session. Only top-level
+    // entity (i.e. user) would be returned from APIs.
+    get currentIdentity(): Identity;
```
louischan-oursky: `get currentIdentity(): Identity | null | undefined;`

chpapa:
Thanks @kiootic ! I think this is good to merge.

PR #335, opened 2019-07-03
Specify SSO

louischan-oursky:
[View rendered version here](https://github.com/louischan-oursky/features/blob/332-sso/features/332-sso/api.md)

### DIFF
--- a/features/332-sso/api.md
+++ b/features/332-sso/api.md
@@ -20,9 +20,8 @@ The flow does not change the location of the current window, instead, it opens a
 
 ```typescript
 interface LoginOptions {
-  // The developer can only set it to false to disable merging, given that
-  // the server has enabled merging.
-  mergeExistingUser?: false;
+  // Specify how the server should handle user duplicate. Default to "abort".
+  onUserDuplicate?: "abort" | "merge" | "create";
   // Which realm to consider during merging. Default to the "default" realm.
   mergeRealm?: string;
 }
@@ -132,23 +131,51 @@ askConfirmation().then(confirmed => {
 
 Some users may forget that they have signed up with email long ago. When they return to the application again, they may login with OAuth 2.0 provider in which the same email is used.
 
-There is a boolean server config `merge_existing_user` to offer two options.
+The client side configuration `onUserDuplicate` has three options `abort`, `merge` and `create`.
+
+The server side configuration `on_user_duplicate_allow_merge` is a boolean flag whether `merge` is allowed.
+
+The server side configuration `on_user_duplicate_allow_create` is a boolean flag whether `create` is allowed.
 
 Since login ID by itself is not unique, the client has an option `mergeRealm` to specify the realm to match against.
 
-## Abort on duplicated email
+## Abort on duplicate
 
 - The Auth gear will infer the email from the provider user info.
 - The email and `mergeRealm` are used to match an existing user.
-- If such match is found and `merge_existing_user` is `false`, the process is aborted and user duplicated error is returned to the client.
+- If such match is found and `onUserDuplicate` is `abort`, the process is aborted and user duplicated error is returned to the client.
 - It is the developer's responsibility to handle this error, probably by showing an error message explaining the situation such as `Seems that you have signed up before. Click here to login.`
 - Preferably after the user has logged in with their email, the application can offer an option to link with the provider.
 
-## Merge on duplicated email
+### Example
+
+```typescript
+try {
+  await loginOAuthProviderWithPopup("google");
+} catch (e) {
+  if (skygear.error.isUserDuplicate(e)) {
+    const navigateToLogin = await alert("Seems that you have signed up before. Click here to login.");
+    if (navigateToLogin) {
+      history.push("/login");
+    }
+  }
+}
+```
+
+## Merge on duplicate
 
 - The Auth gear will infer the email from the provider user info.
 - The email and `mergeRealm` are used to match an existing user.
-- If such match is found and `merge_existing_user` is `true`, that provider becomes a new Identity of that existing user.
+- If such match is found and `onUserDuplicate` is `merge`, that provider becomes a new Identity of that existing user.
+
+### Example
+
+```typescript
+loginOAuthProviderWithPopup("google", {
+  // The server must have enabled merge otherwise this is an error.
+  onUserDuplicate: "merge",
+});
+```
 
 However, merging has security implications as illustrated by the following scenario.
 
@@ -157,9 +184,22 @@ However, merging has security implications as illustrated by the following scena
 3. Alice signs up with OAuth 2.0 Provider and her account is shared with Mallory.
 4. Mallory can now access Alice's profile of that provider.
 
-Therefore, `merge_existing_user` is `false` by default.
-
 > In addition to email, should we consider phone number?
+
+## Create new user on duplicate
+
+- The Auth gear will infer the email from the provider user info.
+- The email and `mergeRealm` are used to match an existing user.
+- If such match is found and `onUserDuplicate` is `create`, a new user is created.
+
+### Example
+
+```typescript
+loginOAuthProviderWithPopup("google", {
+  // The server must have enabled create otherwise this is an error.
+  onUserDuplicate: "create",
+});
+```
 
 # Removal of v1 API
 
@@ -250,9 +290,10 @@ custom_token:
   issuer: ''
   # The share secret with the foreign authentication system.
   secret: ''
-  # Whether or not login should merge existing user by email.
-  # Default to false.
-  merge_existing_user: false
+  # Whether merge is allowed. Default to false.
+  on_user_duplicate_allow_merge: false
+  # Whether create is allowed. Default to false.
+  on_user_duplicate_allow_create: false
 oauth:
   url_prefix: ''
   js_sdk_cdn_url: ''
@@ -262,9 +303,10 @@ oauth:
   # A whitelist of allowed callback URLs
   allowed_callback_urls:
   - 'http://example.com/login'
-  # Whether or not login should merge existing user by email.
-  # Default to false.
-  merge_existing_user: false
+  # Whether merge is allowed. Default to false.
+  on_user_duplicate_allow_merge: false
+  # Whether create is allowed. Default to false.
+  on_user_duplicate_allow_create: false
   providers:
     # "type" chooses the implementation. Required.
     # Valid values are "google", "facebook", "instagram", "linkedin" and "azureadv2"
@@ -318,7 +360,7 @@ Applications that have specific needs to add or remove providers as part of thei
 
 ```typescript
 interface LoginOptions {
-  mergeExistingUser?: false;
+  onUserDuplicate?: "abort" | "merge" | "create";
   mergeRealm?: string;
 }
 
--- a/features/332-sso/api.md
+++ b/features/332-sso/api.md
@@ -303,6 +303,8 @@ oauth:
   # A whitelist of allowed callback URLs
   allowed_callback_urls:
   - 'http://example.com/login'
+  # Whether external access token flow is enabled. Default to false.
+  external_access_token_flow_enabled: false
   # Whether merge is allowed. Default to false.
   on_user_duplicate_allow_merge: false
   # Whether create is allowed. Default to false.
--- a/features/332-sso/api.md
+++ b/features/332-sso/api.md
@@ -282,6 +282,19 @@ oauth:
     # Space-delimited scopes. Required.
     # Note that the scope must be large enough to fetch user info.
     scope: 'profile email'
+    # When "type" is "azureadv2"
+    # The developer must specify "tenant".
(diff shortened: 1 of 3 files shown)


features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+  // Which realm to consider during merging. Default to the "default" realm.
+  mergeRealm?: string;
+}
+
+interface PopupOptions {
+  // If provided, it overrides to the default scope.
+  // Note that the scope must be large enough to fetch user info.
+  // The default scope typically requests access to information
+  // like email.
+  scope?: string[];
```
kiootic: Also, this scope should be a server-side config.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+getAccessTokenViaFacebookSDKSomehow().then(accessToken => {
+  loginOAuthProviderWithAccessToken("facebook", accessToken).then(user => {
+    // login suceedeed
+  });
+});
+```
+
+# Remove OAuth Provider
+
+If the removal will result in no login principals, then removal is forbidden. Such case happens when the user logged in with Google and has no other login means.
```
kiootic: As mentioned in #334, removal of current principal is disallowed.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+
+It **SHOULD** have the following standard claims.
+
+- `aud`: If present and the configured value is not empty, they must match.
+- `exp`: If present, the Auth gear will validate it against the current time.
+- `nbf`: If present, the Auth gear will validate it against the current time.
+
+The signing algorithm is `HS256`. The signing key is shared between the Auth gear and the foreign authentication system.
+
+The token can optionally contain a claim named `skyprofile` which **MUST** be an JSON object.
```
kiootic: Can we use another name? For email, we can use OIDC standard JWT claim `email`. For custom profile, we can use prefixed name like `skygear:profile` instead.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+url_prefix: ''
+custom_token:
+  # By default Custom Token is disabled.
+  enabled: false
+  # The "iss" in JWT claims.
+  issuer: ''
+  # The share secret with the foreign authentication system.
+  secret: ''
+oauth:
+  js_sdk_cdn_url: ''
```
kiootic: Should this configurable by app developer? I think this URL would be set by the hosting admin.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+  skyprofile: {
+    email: "user@example.com",
+  },
+}, secret, { algorithm: "HS256" });
+```
+
+# Tenant Configuration
+
+```YAML
+url_prefix: ''
```
kiootic: Should this configurable by app developer? I think this URL would be set by the hosting admin.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+    skygear.auth.removeOAuthProvider("google");
+  }
+});
+```
+
+# Merging existing user for login
+
+Some users may forget that they have signed up with email long ago. When they return to the application again, they may login with OAuth 2.0, creating a new user identity unexpectedly. To mitigate this issue, the developer can instruct the Auth gear on how to merge existing user.
+
+All login APIs accept `LoginOptions`. By default, merging is enabled. If `mergeRealm` is not in `allowedRealms`, it is an error and the login process is aborted. Otherwise, the implementation will merge existing user by inferring email from the provider user info and use it and the realm to match existing user.
```
kiootic: I think this option should exist as a server side config; app developer may want to disable auto-linking.
Also, I think this option should default to `false`; it may have security implications. Consider this case:
1. A signup a legit website using B's email.
2. A send a link to the website to B and ask B to sign up/sign in.
3. B sign up with SSO and merged with user created by A.
4. A can now access B's SSO profile.

features/332-sso/api.md
louischan-oursky: This is an escape hatch to allow the client to override the server config.

features/332-sso/api.md
louischan-oursky: This is actually following the old spec. So you are proposing we lift `email` up a level so it become a top-level claim? Other than `email`, we are not doing anything special with `skyprofile` so probably we can drop it.

features/332-sso/api.md
louischan-oursky: I leave it undocumented intentionally. We may move it to AppConfiguration but I do not want to do that now.

features/332-sso/api.md
louischan-oursky: Same as `js_sdk_cdn_url`. Actually this setting should be put in `oauth`, let me correct this first.

features/332-sso/api.md
louischan-oursky: I should have stated why this is a client side option. If the application is configured `allowedRealms` to be `["teacher", "student"]`. The client should have a UI to let the user to select the realm and a button to login with a provider. So this option is part of the user input and cannot be configured at server side.

For the security implication, @chpapa any thoughts?

features/332-sso/api.md
kiootic: I don't think we should allow client side to override scopes. What are the use cases that require customization of scopes at client side?

features/332-sso/api.md
kiootic: I should clarify the option I refer to is `mergeExistingUser`; I don't oppose allowing `mergeRealm` at client side.

chpapa: Agreed for `mergeRealm`.

For `mergeExistingUser`, there was a discussion before on the use cases, forgive me for the laziness to dig out the related issues, but the conclusion was basically, we need to support the following use cases:

1. Default: no merging of users, non-unique email addresses
    1. Have an API for client-side to indicate it is an email address logged in, so the application can redirect the user to login to their original account, and link the SSO back to the existing account.
    1. Alternatively, the application can be designed in a way it ignores the above or let users to choose, and have multiple accounts even with identical email addresses for different SSO.
1. No merging of users, unique email address
    1. Same with above, except we would stop people 
1. Always auto-merge users login if email addresses match. (It will have security implication, but some applications think this trade-off worth they can still do it)

So basically the idea is to have a server side config, but need some changes of the API to support use case 1.i and 2, would be great if we can consider it as well.

To clarify, basically use case 1.i is like the following, from a user's perspective:

1. User A already have an account, say email address `a@a.com`
2. User A login with another SSO provider, say gmail, with `a@a.com`
3. User A see a screen say it already have an account, he can choose to login first
4. User A login with his account as usual
5. User A link the SSO provider after login.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+
+# Removal of v1 API
+
+The following v1 API is removed.
+
+```typescript
+function getOAuthProviderProfiles(): Promise<{[providerID: string]: any}>;
+```
+
+The existing implementation simply return all snapshots of user info of all providers of the logged in user. The snapshot of user info most likely will be stale. So the function unlikely return what the developer expects. If we want to make the function always return the fresh user info, we need to implement refresh token of OAuth 2.0. However, to implement refresh token, we have to drop support for `withAccessToken`, not to mention that some providers like Facebook does not support refresh token.
```
chpapa: So we don't handle refresh of data first right? (Is ok, I don't think we handled it before neither actually)

features/332-sso/api.md
chpapa: I will look into #334 which is still a PR now later.

However, we definitely need a way to make unlinking a OAuth login to a user account possible.

features/332-sso/api.md
```diff
@@ -0,0 +1,311 @@
...
+
+# External access token flow
+
+This is not a flow defined by OAuth 2.0. The main usecase is for native applications that have already acquired an access token, for example, via [Facebook Login for iOS](https://developers.facebook.com/docs/facebook-login/ios/).
+
+## Possible Vulnerability
+
+Suppose Alice grants access to Mallory, a malicious application, via OAuth 2.0. Mallory now has the bearer access token of Alice, which is considered a proof of identity. Mallory can now impersonate Alice and login our application with the identity of Alice.
+
+Since this API is vulnerable by design, it is disabled by default. The developer is strongly recommended to use the standard OAuth 2.0 Authorization Code flow even the application is native and has already had acquired an access token.
```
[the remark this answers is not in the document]
louischan-oursky: > Make it very easy and obvious to enable per tenant if the user needs to do that.

Yes, the developer can simply `skycli edit-user-config` to enable for their app only. This is a per tenant configuration.

features/332-sso/api.md
louischan-oursky: > mergeExistingUser

Both the server and the client has `mergeExistingUser`. If it is true in server config, then merging is enabled. The client can only turn it to false to disable merging per request. If it is false in server config, then the client must set it to false only, setting to true will result in error. `mergeExistingUser` is false by default in server.

> support use case 1.i and 2

We can include the email in the error info so the developer can redirect the user to the login screen with email prefilled. Though I am not sure if it is appropriate to reveal the email in this way.

Or we just return an error complaining user duplicated without any email. Then it is the developer's responsibility to catch this error and act accordingly. I prefer this option.

features/332-sso/api.md
louischan-oursky: Yes, v1 did not implement refresh token. So I am proposing to remove this confusing API first. We can add it back when refresh token is implemented.

features/332-sso/api.md
louischan-oursky: The customization of scope is related to `getProviderProfiles`. If the developer can customize the scope, then `getProviderProfiles` will include data with respect to the requested scope. If we remove `getProviderProfiles`, then the developer should have no need to override the scope in the client side.

cc @chpapa

features/332-sso/api.md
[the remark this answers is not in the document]
chpapa: Okay, we need a big remarks in our guides and docs for this lol

features/332-sso/api.md
chpapa: I tends to think it needs to be a parameters over configuration.

If the concern is server side or client side, we can make function (`getProviderProfiles`) with this parameter to be a server side call instead of Client side APIs.

(BTW, in Skygear v1, at client side we have JS/iOS/Android SDK, and part of the JS SDK intended for use at server side with Cloud Functions only; At Skygear v2, I remember there were decisions iOS/Android/JS SDK would be for client side; And server side functions would be as a REST API so we can be programming language agnostic on server side)

features/332-sso/api.md
chpapa: Sure @louischan-oursky , can you write that flow into spec so we make sure our APIs actually support this flow nicely?

features/332-sso/api.md
chpapa: I'm a little bit concern if our current spec are clear enough in specifiying which APIs are client side SDK, which is server side... @louischan-oursky @kiootic

features/332-sso/api.md
louischan-oursky: Sure.

louischan-oursky:
@chpapa @kiootic Comments addressed, except that removal of OAuth 2.0 Provider is still not referencing https://github.com/SkygearIO/features/pull/334

features/332-sso/api.md
```diff
@@ -0,0 +1,343 @@
...
+
+```typescript
+askConfirmation().then(confirmed => {
+  if (confirmed) {
+    skygear.auth.removeOAuthProvider("google");
+  }
+});
+```
+
+# Options on duplicated email during login
```
chpapa: Two comments on this section:

1. Better if each case is illustrated with some code
2. What if the application allows duplicated email with different accounts? (e.g. LoginID `chpapa@gmail.com` and Gmail Login `chpapa@gmail.com` is treated as TWO accounts) Is it possible?

louischan-oursky: > What if the application allows duplicated email with different accounts?

Not possible because `merge_existing_user` is a boolean. To support this use case, we need to replace `merge_existing_user` with `on_user_duplicated: abort | merge | create_new` where `abort` means `merge_existing_user: false`, `merge` means `merge_existing_user: true` and `create_new` means this use case.

chpapa: @louischan-oursky woudl that be a better design?

louischan-oursky: Yes, as it covers one more use case. Let me update the spec to reflect this.

louischan-oursky:
@chpapa @kiootic Updated the spec to support `abort`, `merge` and `create` on user duplicate.

chpapa:
@louischan-oursky I feel like the duplicate one is too complicate, but I have no better idea lol. I think when we actually implement it, it is better if we have a sensible default both on server side and client side. 

No further comments.

louischan-oursky:
No worry, I will try my best to write helpful guide!

Issue #287, opened 2019-02-25
External DB user profile synchronization 


kaojohnny:
# Description

When desigin #256, we considered syncing auth data for user's external DB (such as for some attributes in their user profile). There were 3 approaches discussed:

1. User provide an external DB, and we sync it for the user
1. Use an "eventually consistent hooks"
1. Use a two-phase commit hook for real time data update

### Approach of  "eventually consistent hook"
| Action | Hooked Cloud Function |
| -------- | -------- | 
| `user_object` | `blocking_user_sync(user)`<br>`user_sync(user)`<br> |

and their behavior as:

1. `user_sync` is an `AFTER`, `ASYNC` hook, its implementation could be simple after async hook like other hooks, or it can be optimized by other mechanisms.
2. `blocking_user_sync` is an `AFTER`, `SYNC` hook. `blocking_user_sync` is especially useful when consistency is important to the application. Skygear next will rollback DB if hook failed.

In this issue, we should complete the following tasks:
1. naming of hooks
2. implementation detail of hooks
3. Consider concurrent use cases, like:

| Req 1 | Req 2| Remarks 
| ---- | ---- | ---- |
| enable_op |  | Req 1 called |
| after_enabled_sync() | disable_op | Req 2 called |
|  | after_disabled_sync() | |
|  | [external system update user status to disabled] | For network latency reason, req 2's hook got called first |
| [external system update user status to enabled] | | Oops

this issue happens when two requests happens concurrently in race condition. e.g. one users logged in two devices at the same time and call the ops above.

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
@kaojohnny updated the description to follow the issue format

Issue #285, opened 2019-02-19
Auth Hooks in 2.0


kaojohnny:

chpapa:
#206 change password hook is needed, make sure it is considered in NEXT. @kiootic

Issue #206, opened 2018-04-18
Change password hook


yipsang:
# Description

The developer wants to perform certain actions after the end user changes their password.

## Scenario

Client required an email to be sent after the password is changed by end-user.

chpapa:
@carmenlau I remember it was discussed before but I forgot the content... do you have any idea?

carmenlau:
@chpapa We discussed about this, last time we need to validate the password so we come up with password policy and password history. We thought that can cover most of the use cases. This time we need to send email after user reset password which is not covered. We can further observe if this use case common.

chpapa:
@carmenlau so actually would it be easier if we just have an event hook after user password change for Cloud Functions?

chpapa:
Close as it should be reflect in #285

Issue #340, opened 2019-07-09
Specify auth operation audit


kiootic:
# Description

- Operations to log in audit trail (event, payload, context, etc.)
- Audit trail log retention
- Audit trail retrieval

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
Updated a purpose section for this feature issue since it will supersede / replace a lot of features at #202

PR #341, opened 2019-07-12
Amend design on web-hooks

kiootic:
[Main Document](https://github.com/kiootic/skygear-features/blob/285-webhook/features/285-design-webhooks/design.md)

Refs #285 #206 #287


features/285-design-webhooks/api.md
```diff
@@ -0,0 +1,47 @@
+# Event Management APIs
+
+These API are intended to be consumed at server-side (e.g. in a cron job), no
+client SDK API is provided.
+
+## GET /_auth/events
```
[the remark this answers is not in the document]
kiootic: Duplicated query parameter name would be problematic. I'll use comma-separated values for statuses instead.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,248 @@
...
 
-There are two types of webhook:
+Each operation will trigger two events: BEFORE and AFTER:
+- BEFORE events would be triggered before the operation is performed; the
+  operation can be failed by web-hook handler.
+- AFTER events would be trigger after the operation is performed.
 
-- `SYNC` - gear's operation will be blocked until a response is received. Gear should respect `SYNC` hook's response, if it indicates failed, operation should abort and return a formatted error response to the client.
-- `ASYNC` - gear's operation won't be blocked and the response from the webhook is ignored.
+Both events has the same event payload.
```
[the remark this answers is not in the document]
kiootic: No we can't.

features/285-design-webhooks/config.md
```diff
@@ -0,0 +1,36 @@
...
+## Sample configuration
+In `skygear.yaml`:
+```yaml
+hooks:
+    before_signup:
+        - "https://www.mydomain.com/api/hooks/before_signup"
+        - /profile/before_signup
+    after_signup:
+        - /profile/after_signup
+```
```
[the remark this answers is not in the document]
kiootic: Most of the parameters in that spec is no longer needed, so I updated the config to a more simple format. I'll update other documents if the discussion goes well.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,248 @@
...
+delivered in unspecified order.
 
-```json=
+Web-hook handler must be idempotent.
+
+Web-hook handler must return a status code within the 2XX range. Responding with
+status code outside the range, including 3XX & 5XX, would be considered a
+failed delivery.
+
+The time spent in a delivery must not exceed 5 seconds, otherwise would be
```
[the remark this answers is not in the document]
kiootic: I'll say 5 seconds is already too long; most interactive production web-site would abort the request much faster (e.g. in 1 sec) to prevent long-lasting request from hogging the resources. For example, GitHub's unicorn page.

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,259 @@
...
+        { /* an Identity object */ },
+        { /* an Identity object */ }
+    ]
+}
+```
+
+### before_identity_create, after_identity_create
+```json
+{
+    "is_user_creating": false,
```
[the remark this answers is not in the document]
kiootic: I think it's better to refer to the specific operation (user creation) in the flag's name.
---
I intended this flag to be used to distinguish whether the event is triggered along with user_create events; developer should do user create specific actions in user_create handlers, and may want to avoid duplicate works in identity_create handler.

Most of the time, the handler would only need to user ID from the event, which is present in the identity object. I don't think the user object is needed in the event.

features/285-design-webhooks/api.md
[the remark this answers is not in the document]
louischan-oursky: Why would it be problematic?

features/285-design-webhooks/config.md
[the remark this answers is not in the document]
louischan-oursky: But in case we do have configuration for each hook registration, the original format is better.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
louischan-oursky: Please state this in the document.

features/285-design-webhooks/config.md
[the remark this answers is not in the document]
kiootic: Ok I'll keep the original format and remove the unneeded parameters.

kiootic:
@louischan-oursky updated, please check. Also, the full list of web-hook events is in events.md.

features/285-design-webhooks/config.md
```diff
@@ -0,0 +1,38 @@
...
+
+If the web-hook handler endpoint is a full URL, it must use HTTPS scheme.
+Otherwise, the configuration will be failed to validate.
+
+If the web-hook handler endpoint is a absolute path, it will be resolved to full
+URL with app domain by auth gear at runtime.
+
+To update hooks configuration, a new deployment would be created.
+
+## User Configuration
```
chpapa: Sorry am a bit confused here, what is "user configuration" referring to?

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,263 @@
...
+Each event type can have multiple handlers; the order of deliveries is
+unspecified.
 
-All event payloads follows following payload format:
+BEFORE events will always be delivered before AFTER events. BEFORE events will
+be delivered in well-defined order during a request. AFTER events will be
+delivered in unspecified order.
 
-```json=
+Web-hook handler must be idempotent.
```
chpapa: This one is tricky, what is the reason for it? Is it because we want to relax the constraint of we may call a hook multiple times unexpectedly?

This is a huge trade-off in my opinion, as it impose a constraint of how user write their code (e.g. god know what kind of DB operation they might do with their web hook), and I think it deserves a detailed explanation for us to decided if this trade-off is justified.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: Can we make it configurable? I agree in 70% of case, 1 sec is too long, but there are also occasion people might want longer timeout. 

Another factor is for async webhook, the timeout could be even longer.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,263 @@
...
+- If it is within the operation transaction,
+  then the transaction should rollback on web-hook delivery failure,
+  otherwise async AFTER can be used instead. So:
+    - If the handler has no side-effect, sync BEFORE can be used instead
+      (e.g. custom validation)
+    - If the handler has side-effect, it is subject to distributed transaction
+      problem. We use eventual consistency so async AFTER should be used.
 
-A `SYNC` hook is considered failed if it can't response within 5 seconds, where the value of timeout is configurable of each hook (up to 10 seconds).
+Therefore, we do not offer sync AFTER events.
```
chpapa: In Skygear v1 experience, sync after is rarely useful except when some data are only available AFTER the operation is completed (let's say loginWithOAuth and we only fill in the profile afterwards)

I guess it is a bit hard to think of those cases now, but is it possible we leave the possibility if one day we really need a SYNC after, it could be done?

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,259 @@
...
+  timestamp of when this event is generated; retried deliveries would not
+  affect this field.
+- `request_id`:
+  ID of the HTTP request that generated this event; if the event is not
+  generated by HTTP request, this field would be `null`.
+- `user_id`:
+  the ID of user performing the request that generated this event; if request
+  is not authenticated or event is not generated by user action, this field
+  would be `null`.
+- `principal_id`:
```
chpapa: identity?

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,263 @@
...
 {
-    "event": "event_name",
+    "is_allowed": true
+}
+
+// Disallowed
+{
+    "is_allowed": false,
+    "reason": "the metadata does not match the required format.",
     "data": {
```
chpapa: Will our hook spec include the data return value?

In the old spec, there was a very implicit assumption (which I think is not so good) -- after the hook is completed, the hook can return the DB object (say the user object) for the updated value.

It is useful for cases like:
* if we don't auto-populate the avatar from Facebook Login, a user may want to write a hook that populates the metadata with an avatar from Facebook.
* with some crazy manual condition, like any email address, ends at @oursky.com, the user is always verified or disabled by default -- whatever crazy rules users want to write.

It is helpful as it:
1. Avoid calling more APIs to modify the data, hence lower the call roundtrip time and improve the performance of each webhook
2. Avoid potential recursive lock user may create

Do we want to include the same spec for the return data of auth hooks? That they can update some attributes?

features/285-design-webhooks/config.md
kiootic: The user accessible part of tenant config, i.e. the yaml produced by `skycli app view-user-config` command.

features/285-design-webhooks/events.md
kiootic: Oops I'll fix it.

features/285-design-webhooks/config.md
chpapa: noted and thx.

features/285-design-webhooks/design.md
kiootic: Writing idempotent handler is a best-practice in web-hook design. If the handler is not idempotent, we cannot retry the web-hook (otherwise unexpected things happen), so the web-hook delivery would be much more unreliable than expected.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
kiootic: If we allow setting a longer timeout, people will almost always set it to be the longest possible. So I prefer it to be non-configurable.

If the handler takes more than 5 seconds, it is more than likely some nontrivial work is being performed (e.g. intensive computation, heavy database query). User should perform these work in background instead.

features/285-design-webhooks/design.md
kiootic: This data is just the mentioned 'additional information' for the disallow action; it will be returned in the HTTP response error body.

I don't plan to support modifying the request using web-hook handler.

One of my concern is the audit feature. If web-hook can modify the request arbitrarily, audit logs can no longer be relied on to reflect real action of users.

features/285-design-webhooks/design.md
kiootic: As explained, I believe that a sync after event is useless and cannot be used safely. Therefore I don't think we should consider supporting sync after in future.

features/285-design-webhooks/design.md
chpapa: Best practices vs imposing a constraint to the system are two things.

If we impose a constraint, I expect there is some reason we have to do it; And we have to figure out how we make sure no one makes the mistake / aware of it when they write their own web-hook.

if it is just the best practices, that's just something we would mention in docs somewhere, and maybe warn some features shouldn't be used if they don't do it.

Specifically:

- we don't retry before hook from the spec? So before hook doesn't need to be idempotent if retry is the only reason?
- after hook would be retried automatically so it needs to be idempotent, but we have no way to ensure it is.

In that case it seems more appropriate to just write in spec: "Async after hook is expected to be idempotent"

kiootic: Sorry but I don't see any difference between 'must' and 'expected to be'?

For before events, it would be delivered at most once, so technically there is no need for idempotency. However, it is recommended that no side effect to be performed in it, otherwise it'll need to consider what to do when the operation failed.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: I know it is "better" to perform some long task in the background instead.

However, it is a decision we need to make for the pros and cons trade-off (pros and cons of having a short fixed default timeout):

* For before sync hooks:
    * Pros:
        * Avoid disastor since it could be locking some resources
        * Improve end-users' latency
    * Cons:
        * Developers need to write the code carefully
        * Developers may encounter unexpected error for short timeout, worse experience
* For after async hooks:
    * Pros:
        * Improve end-users' latency
    * Cons:
        * Need developer awareness to write background job code

Let me know if you think the above list missed anything. Otherwise I think it is not so obvious for after async hooks that a short fixed timeout is better. We can left the trade-off to users (hence make the timeout configurable)

@kiootic I hope the above help you make more design decision in future and improve our quality of discussion here.

----

Or another way to think about it is, you also said "people will almost always set it to be the longest possible", that meant there is a strong reason why people want to do it, and we need to take that into account. Why people almost always set it??? Is that reason really bad enough that we should dismissed it?

The subjective decision here is if the risk of "locking resources and affect scalability of the system" meant we should take away the choice from users. In this case I think it is, I have to admit it is gut feeling but if I need to provide a reason, one could be because the cluster is multi-tenant, it could take down other people's app at the same time.

features/285-design-webhooks/design.md
chpapa: @kiootic as suggested, there are cases it could fit the criteria of, say data are only populated after the action, and the hooks want to be called with the data after the action, so your assumption of sync after the event is useless and cannot be used safely could be invalid (at least it is not a logical proof that I'm very confident in)

I didn't go into the details of all possible use cases and hooks we will provide yet. So I'm suggesting if it is an unknown, maybe we want to make sure our design can cater that if it come back to us.

Maybe let me give a sample which I'm unsure if it will happen:
-- e.g. Let's say we have an API for facebook login, only after the login is done (after identities created? after_user_create?) we will have the claims from FB.

However, the user wants to avoid any Facebook login with email address ends with "@gmail.com" to create an account, it seems a perfect case with after_user_create but sync?

features/285-design-webhooks/design.md
chpapa: However, if we actually specify what data could be modified for each hook, aren't we can include that in audit log too?

features/285-design-webhooks/design.md
chpapa: must is stronger, it meant obligations.

Expected / recommended / best practices etc, are way weaker.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
kiootic: For after async events: they're delivered asynchronously, so the request/response latency is not affected.

I'm considering the worst case here: people will likely take advantage of given resources. Also, let's say the default timeout is 1 sec and maximum 5 sec: if developer encountered the timeout, they'll almost surely just increase the timeout, instead of considering whether they're doing too much work in the handler.

The multi-tenant nature of auth gear is indeed my main reason that I prefer it to be a short and non-configurable timeout.

features/285-design-webhooks/design.md
kiootic: I agree 'recommended' and 'best practice' are weaker words. However, in my opinion, both 'must' and 'expected to' means the same: if this requirement is not fulfilled, the system will not work as intended - which is what I want to indicate here.

features/285-design-webhooks/design.md
kiootic: Please note that as stated in the web-hook events document, the before events are delivered right before the commit into database: all possible information are already available.

In your use case, developer should use `before_identity_create` and disallow when the email in event payload is undesired.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: > For after async events: they're delivered asynchronously, so the request/response latency is not affected.

Oh I should have separate it, the latency is not just requested latency, it could be a developer well aware of the timeout and write better code and hence their eventually consistency whatever counter respond faster. So I included that on my comparison table. But yea I should have specify "request latency" and "latency" separately.



> I'm considering the worst case here: people will likely take advantage of given resources. Also, let's say the default timeout is 1 sec and maximum 5 sec: if developer encountered the timeout, they'll almost surely just increase the timeout, instead of considering whether they're doing too much work in the handler.

Yea but again, if we consider the pros/cons, in the case of after sync hook, the only resources we got "locked" is the connection if the users keep increase the timeout to maximum. That won't bother us much?

features/285-design-webhooks/design.md
chpapa: okay I think we are sync here, but just the wording could be misleading. Please make sure the spec is read in a way we want their hook is idempotent, not that it must be idempotent.

features/285-design-webhooks/design.md
chpapa: BTW, I think another advantage of having a return value to update some data is:

1. If we don't have it, in any case the hooks need to update, say metadata, it will have some side effect.
2. If the before sync hook is failed, the users need to take care of the inconsistency of side effect.

If we restore the feature of updating some common data to them, it greatly reduce the complexity as the users can rely on Skygear to update some common data only if the hook is being called and returned successfully.

kiootic: If we specify the fields that can be modified (e.g. metadata in before user create, but not password), the same problem remains: we cannot sure the performed operation is exactly as user intended.

If we include the modified/original fields in audit log, some problems may be considered:
- Audit log is supposed to be immutable and keep for a long time, personal information should not be included.

features/285-design-webhooks/design.md
chpapa: ok. God bless us this assumption holds in future :)

features/285-design-webhooks/design.md
chpapa: I don't get it, logically speaking, no matter it is login hook call update metadata, or login hook return metadata, we could still make an audit log with two entries: Login, and Metadata Updated.

Whether to include those metadata in logs or not is another decision.

louischan-oursky: Agree that we can a separate entry to reflect the fact that the before hook modified part of the payload.

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,260 @@
+# Web-hook Events
```
chpapa: @kiootic @louischan-oursky Have we looked at our list of all current auth API, and make sure use cases are covered?

(Actually, where is the current latest list of all auth APIs? I want to look at it and make some comparisons myself...)

I notice at least one hook is not covered -- after password change. We had a real use case before: https://github.com/SkygearIO/features/issues/206

Can we add the above back? (and also make sure if the whole design still make sense) And also review and see if the web hooks are probably designed with common use cases.

kiootic: Yes I've considered password changed event before. However, since we'd like to implement the minimum required amount of events at the moment due to schedule, so I left it out.

louischan-oursky: > (Actually, where is the current latest list of all auth APIs? I want to look at it and make some comparisons myself...)

I always want to consolidate all auth gear features into a single document so that it is easier for us to see what auth gear currently offers. I also want to archive obsolete or outdated documents so the repo reflects the implemented feature and behavior.

---

https://github.com/SkygearIO/features/tree/master/features/323-login-id

This document covers signup and login.

---

https://github.com/SkygearIO/features/pull/334

This document covers add/remove/replace login id, change password, list identities and the auth response shape.

---

https://github.com/SkygearIO/features/tree/master/features/332-sso

This documents covers SSO, including OAuth 2.0, Custom Token and external access token.
This documents obsoletes

- https://github.com/SkygearIO/features/tree/master/features/5-basic-sso
- https://github.com/SkygearIO/features/tree/master/features/56-login-with-custom-token

---

User verification is not specified in next. Though we have the following documents from v1

- https://github.com/SkygearIO/features/tree/master/features/20-sms-verification
- https://github.com/SkygearIO/features/tree/master/features/22-email-verification

---

Disabling user is not specified in next. Though we have the following document from v1

- https://github.com/SkygearIO/features/tree/master/features/50-disable-user

---

There are some next documents that may be outdated

- https://github.com/SkygearIO/features/tree/master/features/253-auth-sdk-callback-in-next
- https://github.com/SkygearIO/features/tree/master/features/256-auth-gear-interaction-overview-in-next
- https://github.com/SkygearIO/features/tree/master/features/285-design-webhooks
- https://github.com/SkygearIO/features/tree/master/features/294-login-without-specifying-login-id-key

--- 

The following documents are record related so they should be irrelevant now.

- https://github.com/SkygearIO/features/tree/master/features/48-user-profile-discovery

chpapa: > Yes I've considered password changed event before. However, since we'd like to implement the minimum required amount of events at the moment due to schedule, so I left it out.

Ic. It is okay to create spec which doesn't implement yet. Usually, our flows are:

1. Create feature issues
2. Write the high level requirements / APIs
3. Create spec for the detailed user interface
4. Create repo's issues, put those repo issues number as link in feature issue
5. Implement
6. Complete API docs
7. Release
8. Close feature issues

So don't worry about writing a complete spec but implement only a portion of it.

On the other hand, we can also open feature issues for bigger features that seems too large to mix together.

chpapa: > I always want to consolidate all auth gear features into a single document so that it is easier for us to see what auth gear currently offers. I also want to archive obsolete or outdated documents so the repo reflects the implemented feature and behavior.
> 
> [](https://github.com/SkygearIO/features/tree/master/features/323-login-id)

@louischan-oursky sure, go ahead... Actually it is good if we can consolidate the "releasable version" of Skygear Auth's remaining features / spec in the original tracking feature issue: https://github.com/SkygearIO/features/issues/256

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
kiootic: > Yea but again, if we consider the pros/cons, in the case of after sync hook, the only resources we got "locked" is the connection if the users keep increase the timeout to maximum. That won't bother us much?

I think you mean after async? I'm not sure if the connection socket is the only resource user's holding while the event delivery is in progress: we need to consider memory, processing power, bandwidth, storage, etc.

Even if we concluded that the amount of resource is trivial, do we want to encourage developer to do incorrect thing (long-running job in event handler), by providing an easy way out (increasing timeout)?

features/285-design-webhooks/design.md
kiootic: Still, I'm not seeing how using 'must' is misleading: 'must' is a well-known word used in spec (see RFCs). Would you consider 'should' a better word?

features/285-design-webhooks/design.md
kiootic: Whether updating metadata should produce an audit log entry is another thing to specified in #340, but we can put it aside for now.

After discussing with @louischan-oursky, I'll allow mutation on some limited fields in before sync event handler.

chpapa: Sure. BTW, I think it is better if all auth gear have the same set of fields that users can update for better developer experiences.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: So it sounds like my assumption of after async hook would only involve connection resources on our side is right? In that case my recommendation would be:

1. For before sync webhook, we have a default value of 5 sec and it is not configurable (@louischan-oursky suggested 5 secs is too short originally, so he may have another opinion on this)
2. For after async webhook, we have a default value of 30 sec and it is configurable to at most 60 secs.

> Even if we concluded that the amount of resource is trivial, do we want to encourage developer to do incorrect thing (long-running job in event handler), by providing an easy way out (increasing timeout)?

As explained above, we need to consider the pros and cons of each design decision, not just gut feeling of rights or wrong. In this case, the "easy way out" is a major improvement to developer experience when they needed. The "incorrect thing" do not have any major drawback besides the resources to hold the connection -- Unless there are more reason to justify why it is an "incorrect thing" for async hook to be longer

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,307 @@
...
+specified time.
 
-app.post('/before_signup', function(req, res){
-    var user = req.body.data.user;
+If the delivery continue to fail after 3 days from the time of first attempted
+delivery, the event is marked as permanently failed and will not be retried
+automatically.
 
-    user.metadata.loveCat = true;
+## Mutations
```
chpapa: I notice different handlers have different mutation now.

But would it be better if we have the same set of data that users can mutate for all auth hooks?

The main reason is to just make it easier for developers, so they don't need to go back to the doc every time to understand what fields are mutable. 

On the other hand, it also doesn't seem we could imagine what are the possible use cases anyway, like god knows maybe users want to put some default metadata `before_user_create` and currently it is not mutable. It would drives us crazy to figure out all possible use cases at this stage.

kiootic: Initial metadata can be mutated in `before_user_create` event handler. At the moment, the possible mutations are based on the use cases you mentioned before: populating metadata according to SSO profile and auto-verify/disable based on identity.

I don't think we need to allow mutations of same fields for all events. Developer need to know what events are available, and what fields is contained in payload anyways.

features/285-design-webhooks/design.md
kiootic: I don't think it's better to use 'expect to'. I found 'must' much more clearer:
1. the web-hook system is designed with idempotent handlers in mind.
2. if developer don't write idempotent handlers, the system will not work as we indented.
3. therefore, developer must write idempotent handlers.

Also, the point 2 you mentioned ('the webhook won't works if it is not XXX'), is in fact true; the webhook won't work properly if it's not idempotent.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
kiootic: I'd agree on a longer timeout for after async event deliveries, even 60 seconds would be okay by default; I mainly oppose configurable/long timeout for before sync events.

Discussed with @louischan-oursky :
- For before sync events: 5 seconds by default, configurable in app config (can be modify by admin only)
- For after async events: 60 seconds, non-configurable for now.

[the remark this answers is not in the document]
chpapa: Agree if before sync is configurable in app config by admin only, seems more flexible especially in the case if the Skygear Cluster is owned by, say an Enterprise, they can do whatever...

But I'm curious, the main reason for after async is non-configurable per app is....? Is there any reason why one might want to shorten it?

features/285-design-webhooks/design.md
chpapa: Logically it is not a must -- even if the hook is not idempotent, but if the users don't care about the consequence, it doesn't matter then.

On the other hand, the next sentence in the same paragraph "Web-hook handler must return a status code within the 2XX range.", that must is trivial and easy to understand, because if it doesn't return a 2XX status code, we treat it as failure.

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
kiootic: 60 seconds is more than enough for most use cases, so just make it non-configurable for now.

features/285-design-webhooks/design.md
chpapa: As mentioned above, it doesn't seem we could imagine all possible use cases.

I can simply put another story up, let's say someone want to have a metadata to count the number of times a user in auth got disabled. `metadata` is not mutatable in `user_disabled_status_update` right?

Are you going to add that back to mutations of `disabled_status_update`?



> I don't think we need to allow mutations of same fields for all events. Developer need to know what events are available, and what fields is contained in payload anyways.

Developer "need to know" != Good developer experience. It is just way more simple to have a single set of fields that we can mutate for the developer in all auth hooks. 

On the other hand, do you have any reason why you want to complicate it with different data are mutable for different hooks?

features/285-design-webhooks/design.md
kiootic: I'll compromise and say:
`Web-hook handler should be idempotent, since the same event may be delivered multiple times due to retries.`
Does that sounds okay to you?

chpapa: Sounds fine. thx

features/285-design-webhooks/design.md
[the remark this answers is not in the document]
chpapa: Sounds fine.

features/285-design-webhooks/design.md
kiootic: For your story, I'd say you should not do this in before events: you cannot be sure the user will be disabled in before events. After events should be used instead, and we do not encourage user to save this in metadata anyways.

I'd say allowing mutation of the same set of fields in all events would complicate things:
- what'd happen if user is disabled in `before_session_create`?
- do we generates `user_metadata_update` events if user modify metadata in any hooks?
To put it simple, I can't ensure allowing arbitrary fields in before events would make sense, maintainable, or have no security implications. If this is allowed, we have to support it for a long time until next major version change. I'd prefer keeping the bare minimum useful mutations at the moment.

chpapa: > For your story, I'd say you should not do this in before events: you cannot be sure the user will be disabled in before events. After events should be used instead

So are you suggesting we have different mutation fields in before and after hook?

> and we do not encourage user to save this in metadata anyways.

Why we do not encourage user to save this in metadata anyways?

>     * what'd happen if user is disabled in `before_session_create`?

It seems obviously logically the session won't be created? Or what is the problem?

>     * do we generates `user_metadata_update` events if user modify metadata in any hooks?

This one is interesting, have you think about it for `user_create` hooks?



>     * To put it simple, I can't ensure allowing arbitrary fields in before events would make sense, maintainable, or have no security implications. If this is allowed, we have to support it for a long time until next major version change. I'd prefer keeping the bare minimum useful mutations at the moment.

I don't against try to keep it minimal, but could we have a subset that is small enough and common for every hooks? I think the additional mental burden for developers to check mutation is too high. If we have a minimal set of attributes that is mutable, we can still consider their implication and most hook's implication are relatively similar? (Like the metadata update event above)

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,310 @@
...
+
+For the detail on signature generation and validation, refer to #300.
+
+> For advanced end-to-end security scenario, some network admin may wish to
+> use mTLS for authentication. We do not support this at the moment.
+
+
+# Considerations
+
+### Recursive Web-hooks
```
chpapa: Something popup in my mind. I remember in the early days when we discuss calling between gears / microservices at Skygear, we talk about adding a field in the header as a counter, to avoid recursive calling of functions (it doesn't have to be web-hooks)

Has this design been written in spec finally? (Sorry actually I don't recall if we have even designed the mechanism for a cloud functions calling another cloud functions, like how to verify the data are trusted, etc) Maybe that can be re-used here? (Since webhook is technically another way a cloud functions -- auth gear calling another cloud functions if it is inside Skygear)

louischan-oursky: Technically speaking, the developer is free to call our endpoints by hand. In that case, they do not know they should include that counter header. We could make that header mandatory. However I can predict that every developer will make the mistake of omitting the header.

chpapa: DX-wise, we would probably need to provide a set of library/SDK (for some supported languages like Typescript, Python etc) to handle at least two things for web-hook calling / call between microservices / cloud functions:

1. How to authenticate the caller's identities (users identity or some sort of master key)
2. Avoid an infinite recursive loop of calling.

Anyway it seems beyond the scope here. I created another feature issue #342 for that. Feel free to comment there.

features/285-design-webhooks/design.md
kiootic: > So are you suggesting we have different mutation fields in before and after hook?
We do not support mutations requested by after event handlers: they're asynchronous.

> Why we do not encourage user to save this in metadata anyways?
We encourage developer to store application-specific data in their own database right?

> It seems obviously logically the session won't be created? Or what is the problem?
Your suggested behavior is logically sound. The problem is the behavior would be different for different events. If it is signing up, would disabling the user fail the signup, or return the user object without a session? If user is already logged in, would disabling the user failed the operation, revoke the session, or whatever?

> This one is interesting, have you think about it for user_create hooks?
This has been noted in the spec: `user_create` is not updating metadata, but creating. So `user_metadata_update` event would not be generated.

> I don't against try to keep it minimal, but could we have a subset that is small enough and common for every hooks?
I think the current list of mutation is reasonable: developer can mutate whatever the event is mutating. When user is creating, the whole user object is mutated so all reasonable fields are available. When an API is used to modify a field, that specific field would be available for mutation.

chpapa: > We do not support mutations requested by after event handlers: they're asynchronous.

Why async event handler = no mutations?

> We encourage developer to store application-specific data in their own database right?

I think the line is actually quite blurred, just like I'm not sure what you meant by application-specific data neither. Practically we would limit the functionality of metadata and users need to judge what they want to do with it, and probably there are 3 reasons they might want to put something in metadata:
1. Because it needs to be displayed in User Managemenet Dashboard or Auth UIKit that Skygear's code needs it to be accessible.
2. Because it is used as claims or something else for identity provider (the feature is not there yet)
3. Some other reason that is beneficial, e.g. we want to have built-in field-level encryption per-key in metadata

So in that case, maybe number of disable is a value they want to see in User Management Dashboard (which makes perfect sense, maybe admin user need it to decide if the user is fishy or for customer support reason)

> The problem is the behavior would be different for different events. If it is signing up, would disabling the user fail the signup, or return the user object without a session? If user is already logged in, would disabling the user failed the operation, revoke the session, or whatever?

Sounds like something we can still design for?

1. If user is already logged in and disabled, it won't affect the current operation / revoke the session, just not able to login (Unless our disable user function will revoke all sessions immediately, in that case we should be consistent and revoke the session when disabled is set true)
1. disabled during signup: this one is more tricky, if we don't revoke sessions after a disabled field is set as currently, I suggest we should just be consistent and return the session first.

> This has been noted in the spec: `user_create` is not updating metadata, but creating. So `user_metadata_update` event would not be generated.

I don't see it being specified anywhere. If you meant the words "initialize" meant "not updating", I think you need to be specific with it in the spec.

Similar, maybe we should also specify would metadata itself being updated by mutations at `user_metadata_update` webhook would not trigger another update? (or would it?)

> I think the current list of mutation is reasonable: developer can mutate whatever the event is mutating. When user is creating, the whole user object is mutated so all reasonable fields are available. When an API is used to modify a field, that specific field would be available for mutation.

You're missing the point, it would be much easier for developers to use Skygear if the mutations list is common, so I'm suggesting a minimal list of **common** mutations.

Another point is the current list doesn't seems cover enough use case. If I'm not sitting down to do some serious work at thinking about the use case, but I can still think of a lot of use cases that current design can't cater, I won't think the list is "reasonable".

Here are a random list of use cases...
* Count number of disabled happens for a user (as mentioned above)
* Manually update `verified` status when new identity is created, maybe some special rules for an app
* Disable a user automatically after a certain number of login event

So maybe we need to think about the use cases carefully, and make sure the list is reasonable.

OR, since there are not many fields in User Object actually, we can enumerate it and make a common list that make sense and should cover most use cases. (In fact I don't recall there are other fields we want to let users to mutate besides disabled, verified, metadata (or verified_info).

@louischan-oursky please help chip in.

kiootic: > Why async event handler = no mutations?

We cannot guarantee the requested mutation would be successful in async event handler. Developer should call auth gear API directly and handle possible errors.

> I don't see it being specified anywhere.

It is noted in the description of `user_metadata_update` event:
```
**NOTE**
- To handle metadata creation, use `before/after_user_create` instead.
```

> we should also specify would metadata itself being updated by mutations at user_metadata_update webhook would not trigger another update?

It would not; we would not keep generating `before_XXX_update` events until the data become stable. As stated in design document:
```
If an entity is mutated, subsequent events will used the mutated value.

Developer is responsible for ensuring correct order of event delivery. For
example, in most case, developer would like mutating event-handlers (e.g. 
populating default value for some fields) before non-mutating event-handlers
(e.g. validating field values).
```

kiootic:
@chpapa @louischan-oursky updated, please check.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,324 @@
...
 
-```json=
+Web-hook handler should be idempotent, since the same event may be delivered
+multiple times due to retries.
+
+Web-hook handler must return a status code within the 2XX range. Responding with
+status code outside the range, including 3XX & 5XX, would be considered a
+failed delivery.
+
+### BEFORE Events
```
louischan-oursky: Mention mutation in this section by providing a link to that section.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,324 @@
...
-## Overview
+Web-hooks would be used to notify external services of events in auth gear.
 
-skygear sends webhooks for events that happen in your app, and are designed to help a developer to monitor events, sync database, and many other use cases.
 
-When one of events is triggered, skygear sends a HTTP POST with a payload to the webhook's configured URL.
+# Proposed Designs
 
-## SYNC and ASYNC hooks
+## Events
```
louischan-oursky: Mention all kinds of hooks here. BEFORE, AFTER and USER_SYNC. This section is the very first part of this document so it should be an overview.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,324 @@
...
+                    }
+                }
+            ]
         }
     }
 }
 ```
 
-Note that, `context` is above format is still under discussion.
+The time spent in a BEFORE event delivery must not exceed 5 seconds, otherwise
```
louischan-oursky: Move the sentence `This timeout can be configured in app config` to the end of this paragraph. Rewrite it to `Both timeout are configurable in app config`.

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,324 @@
...
-If webhook wants to include additional information for an error case, it could return with payload:
+- If a field is absent: no mutation would be performed for that field.
+- If a field is present: the field would be set to the provided value.
+
+If failed to perform the mutations, the operation would be failed.
+If the operation failed, the mutations are rollbacked and not visible.
+
+Mutations would not generate additional events (except `user_sync`).
+
+If a field is mutated, subsequent events will use the mutated value.
```
louischan-oursky: `The mutated value is be propagated along the chain of hooks.`

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,348 @@
...
+
+**NOTE**
+- When user is performing authentication (e.g. signing up/logging in),
+  the request will not be treated as authenticated with the user to
+  sign up/log in, in both BEFORE/AFTER events.
+
+
+## Mutations
+
+For BEFORE events about a user, following fields can be mutated:
```
louischan-oursky: Can we also add `is_verified`?

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,348 @@
...
+```json
+{
+    "user": { /* a User object */ }
+}
+```
+
+- `user`: the user object after the operation
+
+**NOTE**
+- This user object may not be updated for consecutive events; even if the
```
louischan-oursky: > The event would be generated unconditionally whenever a mutating API is called. For example, if the user is initially `disabled = false`, setting `disabled = false` will generate this event.

Adding example seems easier to understand.

features/285-design-webhooks/events.md
```diff
@@ -0,0 +1,348 @@
...
+  operation does not actually modify the user state, this event would be
+  generated.
+
+
+## Sample Web-hook Event
```
louischan-oursky: The example has the `version` field placed in wrong position. Please correct them.

features/285-design-webhooks/events.md
kiootic: I'd prefer to add this when we provide a way to turn off auto-update of the flag.

kiootic:
@chpapa @louischan-oursky updated, please check.

features/285-design-webhooks/use-cases.md
```diff
@@ -0,0 +1,285 @@
...
+## Synchronize Metadata to Self-Managed Profile
+
+Developer may want to synchronize metadata when it is updated by user:
+- When user update metadata, an external service should be notified to
+  synchronize their managed profile according to user metadata in auth gear.
+- External service may want to validate & reject invalid user metadata.
+
+### Suggested Solution
+
+- In `before_user_metadata_update`:
```
carmenlau: From events.md, it should be `before_user_update`?

kiootic: Yes it should be.

features/256-auth-gear-interaction-overview-in-next/api.md
```diff
@@ -54,173 +54,6 @@ Following function will be removed from client SDK, a developer should create it
 
 For security reason, Cloud Function should invoke admin related API with master key.
 
-## Execution flow of hooked Cloud Function
```
chpapa: @kiootic just for readability sake, can you:

1. Add a section here called "Web Hook", with a sentence mention details of hook is in another spec (path)
2. Add a table similar to the one below, with all hooks supported.

features/285-design-webhooks/config.md
```diff
@@ -0,0 +1,46 @@
...
+The top-level key `hook_secret` is a string. This is the value of web-hook
```
chpapa: Add a sentence mention how the secret is used? Maybe just say refer to Security in 285/design.md?

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,329 @@
...
+Each event type can have multiple handlers; the order of deliveries is
+unspecified for AFTER events, and same order as in `skygear.yaml` for BEFORE
+events.
 
-```json=
+BEFORE events will always be delivered before AFTER events. BEFORE events will
+be delivered in well-defined order during a request. AFTER events will be
+delivered in unspecified order.
+
+Web-hook handler should be idempotent, since the same event may be delivered
```
chpapa: Only after event will be delivered multiple times?

features/285-design-webhooks/design.md
```diff
@@ -1,129 +1,329 @@
...
 ```
 
-If webhook wants to include additional information for an error case, it could return with payload:
+- If a field is absent: no mutation would be performed for that field.
+- If a field is present: the field would be set to the provided value.
+
+If failed to perform the mutations, the operation would be failed.
+If the operation failed, the mutations are rollbacked and not visible.
+
+Mutations would not generate additional events (except `user_sync`).
```
chpapa: `user_sync` is an after event I think, so there are no mutation?

features/285-design-webhooks/design.md
kiootic: Currently just after events. However, not sure if we would want to retry before events too in future, so I used 'may'.

features/285-design-webhooks/design.md
kiootic: Yes, I mean mutations would not generate additional events, however `user_sync` would still be generated.

chpapa: But logically `user_sync` have no mutation, so it is werid to say (except `user_sync`)?

features/285-design-webhooks/design.md
chpapa: Let's be consistent first. Either make it clear it is talking about `after webhook` only, OR explain here it is for `after webhook` now but maybe for `before webhook` in future and why... (I suggest no need to explain something we are not sure if we will add in future)

features/285-design-webhooks/design.md
kiootic: Yes I've further clarified the except clause in the spec.

kiootic:
@louischan-oursky @chpapa @carmenlau updated, please review again.

features/285-design-webhooks/design.md
chpapa: The clarification still read werid, readability and logically-wise seems doesn't make sense to specially mention `user_sync` which is an after async event with no mutation supports.

I guess you actually meant something like this?

> Mutations would not generate additional events unless users call other APIs to update the data independent from returning mutation data in the web hook.

kiootic: I'll remove that note instead, since `user_sync` is not generated based on mutations.

kiootic:
@chpapa updated, please review again.

PR #343, opened 2019-07-19
MFA

louischan-oursky:
[Authentication Session](https://github.com/louischan-oursky/features/blob/26-mfa/features/26-mfa/authentication-session.md)
[API](https://github.com/louischan-oursky/features/blob/26-mfa/features/26-mfa/api.md)
[Implementation](https://github.com/louischan-oursky/features/blob/26-mfa/features/26-mfa/implementation.md)


features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+
+interface RecoveryCodeAuthenticator {
+  id: string;
+  type: "recovery_code";
+  createdAt: Date;
+}
+
+interface BearerTokenAuthenticator {
+  id: string;
+  type: "bearer_token";
```
kiootic: I'm not sure if bearer token should be a standalone authenticator: if I login with a TOTP authenticator and get a token, the token should be invalidated if the TOTP authenticator is deleted.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+async function regenerateRecoveryCode(): Promise<RegenerateRecoveryCodeResult>;
+```
+
+# Use Cases
+
+## Registration
+
+```typescript
+try {
+  await skygear.auth.login("user@example.com", "password");
```
kiootic: You mean `signupWithEmail`?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+
+```typescript
+// Present an UI to let the user to choose authenticator.
+// Suppose the user chose OOB.
+
+// Present an UI to instruct the user to input a phone number.
+const e164number = textInput.value;
+const { authenticatorID } = await skygear.auth.mfa.createNewOOB({ phone: e164number });
+
+// Present an UI to instruct the user to check SMS.
```
kiootic: I assume `createNewOOB` will automatically do `triggerOOB`?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+  #
+  # 'optional' means the user is free to register a new authenticator.
+  # When a user has at least one active authenticator, the user must login with MFA.
+  # This is like how other online services such as Google and Facebook offer MFA.
+  #
+  # 'required' means the user must login with MFA.
+  # APIs like change password will return MFA required error.
+  # This is like how banking services offer MFA.
+  #
+  # Default is 'optional'
```
chpapa: Alternatively, I think we need a configuration for the idea of "when mfa is needed if available", and the choices could be:

1. Always -- if the user have MFA, it is always required (default)
2. When not in IP address ranges
3. When country change
4. If metadata is set as X 

I think we can only support 1 and 2 to avoid complexity, but at least we have this concept in it?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+  if (!skygear.isMFARequiredError(e)) {
+    throw e;
+  }
+
+  let done = false;
+
+  const bearerToken = await getBearerToken();
+  if (bearerToken) {
+    try {
+      await skygear.auth.mfa.authenticateWithBearerToken(bearerToken);
```
kiootic: Do we save the authentication state (i.e. email & password) internally in SDK?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+
+  // Persist the bearer token
+  if (bearerToken) {
+    await storeBearerToken(bearerToken);
+  }
+}
+```
+
+The above example demonstrates the flow of authenitcation in details.
+The bearer token should be handled transparently by the SDK so the actual code written by the developer should not need to be aware of bearer token.
```
kiootic: The above flow seems to be the explicit flow using bearer token. We should also show a sample flow for transparent flow using bearer token.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+//     type: "totp",
+//     activatedAt: new Date("2019-07-19T00:00:00.000Z"),
+//     displayName: "totp-2edfdca2-3f05-41d2-bfd8-502725ee09a6",
+//   },
+//   {
+//     id: "2",
+//     type: "oob",
+//     activatedAt: new Date("2019-07-19T00:00:00.000Z"),
+//     channel: "sms",
+//     phone: "+85223456789",
```
kiootic: Not sure if showing the full phone here would have some security implications?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+// Present an UI to show the list of authenticators and
+// let the user to choose which one to delete.
+
+// Suppose the user wants to delete the first one.
+const { id } = authenticators[0];
+await skygear.auth.mfa.deleteAuthenticator(id);
+
+// The user can also delete a bearer token if they believe
+// the device storing it has been compromised.
+const bearerTokens = authenticators.filter(a => a.type === "bearer_token");
```
kiootic: I don't think we need to provide a way to invalidate a specific bearer token. I just checked google only has option for revoking all trusted devices.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+  const { bearerToken } = await skygear.auth.mfa.authenticateWithOOB({ authenticatorID, code }, { requestBearerToken: true});
+
+  // Persist the bearer token
+  if (bearerToken) {
+    await storeBearerToken(bearerToken);
+  }
+}
+```
+
+The above example demonstrates the flow of authenitcation in details.
```
chpapa: Could we have two set of codes here?

1. Maybe REST APIs / some sort of pesudocode like yours now, to present how it works underlying or on REST API level (like the bearer token)
2. A JS sample code to show the expected client side SDK.

Reason is I think besides bearer token, there are multiple things we may want to provide at client sdk which affect developer experience heavily. (e.g. The secret tokens of OTP should also be in the form of QR code too)

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+// Present an UI to show the list of recovery codes to the user.
+```
+
+# Access Token
+
+The access token includes a new claim [amr](https://openid.net/specs/openid-connect-core-1_0.html).
+
+## Mapping between existing authentication method and factor name
+
+|Method|Factor Name|
```
kiootic: There exists a list of defined factors: https://tools.ietf.org/html/rfc8176

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+```JSON
+{
+  "amr": ["password", "oob"]
+}
+```
+
+The injected header is
+
+```
+x-skygear-auth-amr: password, oob
```
kiootic: I think there should not be a space after the comma.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+The usefulness of this injection is to allow microservice to determine whether MFA is required.
```
kiootic: I think you mean determine whether the user is logged in with MFA?

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
+# Overview
+
+The following factors are supported:
+
+- Time-Based One-Time Password [(TOTP)](https://tools.ietf.org/html/rfc6238)
+- Out-of-band
+  - SMS
```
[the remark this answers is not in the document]
louischan-oursky: I intentionally left it out because access to an email is not an additional factor. Access to an email is "something they know" which is the same factor as password.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+- Recovery code
+- Bearer token
+
+In the future, the below factors will be supported:
+
+- [WebAuthn with CTAP2 or FIDO U2F](https://fidoalliance.org/specifications/)
```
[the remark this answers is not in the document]
louischan-oursky: I have read it and our design should be able to implement it in the future.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+  # 'optional' means the user is free to register a new authenticator.
+  # When a user has at least one active authenticator, the user must login with MFA.
+  # This is like how other online services such as Google and Facebook offer MFA.
+  #
+  # 'required' means the user must login with MFA.
+  # APIs like change password will return MFA required error.
+  # This is like how banking services offer MFA.
+  #
+  # Default is 'optional'
+  enforcement: 'optional'
```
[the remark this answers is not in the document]
louischan-oursky: You mean within the same app, some users (maybe they are admin) require MFA while some other users (maybe they are casual users) do not? Since we do not have authorization now and the auth gear does not have the concept of admin (we use master key), the application should use the injected `amr` header to perform authorization themselves.

features/26-mfa/api.md
louischan-oursky: There are more rules we cannot think of. Case 2 and 3 can be done if the gateway inject `x-real-ip` and the gateway can infer country from IP address. I hope injecting informational headers can help application to perform their complex logic. However, the downside is that the API of auth gear is not affected by this mechanism.

features/26-mfa/api.md
louischan-oursky: > Could we have two set of codes here?

OK. I will add it.

> Reason is I think besides bearer token, there are multiple things we may want to provide at client sdk which affect developer experience heavily. (e.g. The secret tokens of OTP should also be in the form of QR code too)

Display the secret in form of QR code should be done by another library. I think the document should definitely teach the developer how to do this for each SDK we support.

features/26-mfa/api.md
louischan-oursky: Good point! Maybe for bearer token authenticator it must have a reference to an existing authenticator (only "totp", "oob" and "recovery_code", does not make sense to depend on another bearer token authenticator). When the parent authenticator is deleted, the dependent bearer token authenticator(s) must also be deleted.

features/26-mfa/api.md
louischan-oursky: I really meant `login`. The user should be able to sign normally. The access token they get will have `"amr": ["password"]`.

features/26-mfa/api.md
louischan-oursky: Yes. You're right.

features/26-mfa/api.md
kiootic: I think it does not make sense to depends on a recovery code authenticator too?

features/26-mfa/api.md
kiootic: I see, so the section title means MFA registration, not user registration.

features/26-mfa/api.md
louischan-oursky: The only things we store are

- access token
- This newly introduced MFA bearer token

Why would you ask this question?

features/26-mfa/api.md
louischan-oursky: OK. I will add it.

features/26-mfa/api.md
louischan-oursky: Maybe we can mask it? In that case, I would rename it to `maskedPhone` and it should look like

```
maskedPhone: "+8522345****"
```

features/26-mfa/api.md
kiootic: When I see an error occured during the `login` call, I assume the login failed and no access token is returned. So I'm thinking how the password credentials is stored in the MFA process.

features/26-mfa/api.md
louischan-oursky: So we should have a first-class function like `deleteAllBearerTokens()`? In the future when we design the session API, should we delete the associated bearer token when a session is revoked?

features/26-mfa/api.md
louischan-oursky: Thanks for letting me know! If the user is authenticated with password and OOB, then `amr` should be `["pwd", "mfa", "sms"]`

features/26-mfa/api.md
louischan-oursky: Oops. Typo.

features/26-mfa/api.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+The injected header is
+
+```
+x-skygear-auth-amr: password, oob
+```
+
+The usefulness of this injection is to allow microservice to determine whether MFA is required.
+The developer is expected to detect as condition in client-side and guide the user to
```
louischan-oursky: s/detect as/detect this/

features/26-mfa/api.md
```diff
@@ -176,10 +177,11 @@ try {
 // Present AN UI to let the user to optionally name the TOTP authenticator.
 
 const displayName = textInput.value;
-const { authenticatorID, secret } = await skygear.auth.mfa.createNewTOTP(displayName);
+const { authenticatorID, secret, otpauthURI } = await skygear.auth.mfa.createNewTOTP(displayName);
 
 // Present to the secret to the user and let the user
 // to add the secret to their TOTP application, such as Authy and Google Authenticator.
+// Or preferably use a OR code to display otpauthURI, it is compatible with Google Authenticator and Authy.
```
kiootic: I don't think we should provide a otpauthURI; the app should generate it at client side with the provided secret, since most of the parameters in URI seems app-specific (e.g. issuer, account name).

Also, typo: OR should be QR?

features/26-mfa/api.md
kiootic: Yes, I think a function named `revokeAllBearerTokens` is appropriate. If a session is revoked, I don't think we need to revoke the associated bearer token. However, if the API is 'logout this device', then we should revoke the associated bearer token.

features/26-mfa/api.md
chpapa: SDK is client-side, and for usability we definitely need otpauthURI. In fact, I think on Client-side, we should have generated the QR code for users to directly display it, instead of "generate QR code from the otpauthURI".

(and it should depend on the platform, so on Web we probably need server-side to generate the QR Code and return the URL in the form of a URL; On iOS / Android we can generate the QR code on client-side devices and return it as some sort of object that users can use to display easily)

features/26-mfa/api.md
chpapa: So....?

features/26-mfa/api.md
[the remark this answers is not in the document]
chpapa: @louischan-oursky so do you meant the way it should works is:

1. Developers set the enforcement to optional
2. Developers write the code so that in the `amr` header, if there are no mfa but requires for the users or other situation, the code should return error to client side, and the client side will trigger the 2FA login manually?

If that's what you meant (and what current design can support) I think is okay.

features/26-mfa/api.md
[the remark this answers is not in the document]
chpapa: It depends, if the users use loginWithSMS / forgotPasswordWithSMS something like that, than SMS is not another factor lol. But yea this is not very important... just nice to have if not complicated.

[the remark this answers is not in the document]
louischan-oursky: OK. Let me add it as a channel of Out-of-band.

features/26-mfa/api.md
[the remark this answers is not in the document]
louischan-oursky: Yes.

features/26-mfa/api.md
louischan-oursky: > since most of the parameters in URI seems app-specific (e.g. issuer, account name).

I believe the parameters should be relatively static so they can be configurable in tenant config.

> so on Web we probably need server-side to generate the QR Code and return the URL in the form of a URL

Good suggestion! It should be possible to add an endpoint to serve a QR code image. Let me include this.

kiootic: > I believe the parameters should be relatively static so they can be configurable in tenant config.

Account name can be customized, e.g. user display name from self-managed database, role of user, etc.

louischan-oursky: The specification of otpauth is here https://github.com/google/google-authenticator/wiki/Key-Uri-Format

The only configurable input to generate otpauth URI is "Issuer" and "Account Name". Since a user can have many login IDs, the Auth gear cannot find a good candidate for "Account Name".

I think instead of generating the otpauth URI on server-side, the client SDK should offer an utility function to generate otpauth URI.

To generate an QR code image, the developer should first use our client SDK to generate otpauth URI. The server offers an endpoint to accept a otpauth URI and return an QR code image.

chpapa: @louischan-oursky the specification didn't specify what is account name, only say it is for identifying the account uniquely.

I'm a bit confused, if Auth gear cannot find a good candidate for Account Name, how would Client SDK able to choose it? (or how Developers are supposed to choose it?)

louischan-oursky: Both the "Issuer" and "Account Name" is for display purpose in the Authenticator Application such as Authy and Google Authenticator. Only the secret is used in deriving the Time-based one-time password. The primary purpose of issuer and account name is for the user to identify registered secrets. If the application uses email as login ID, then most likely the developer should use the email of the user as "Account Name". Similarly "Issuer" should be a user-facing application name chosen by the developer.

It is perfectly fine to have duplicate "Issuer" and "Account Name" pair as illustrated by the following screenshot.

IMG_0185

chpapa: ic, sounds like we should let user define their own issuer and account name.

I'm thinking if there is any security implication... would some how a social attack could be engineered to make users fake a OTP with a phishing website' issuer and account name? hmmm

features/26-mfa/api.md
```diff
@@ -325,60 +315,39 @@ const authenticators = await skygear.auth.mfa.listAuthenticators();
...
+// require MFA.
+await skygear.auth.mfa.revokeAllBearerTokens();
 ```
 
-## Regenerate recovery codes
+## List and regenerate recovery codes
 
 ```typescript
+// If listing is enabled, the user can list the existing recovery code.
+cpnst { recoveryCodes } = await skygear.auth.mfa.listRecoveryCode();
```
kiootic: typo

louischan-oursky: Which one?

kiootic: cpnst should be const?

features/26-mfa/api.md
louischan-oursky: The MFA token is included in the error.

kiootic: Yes I understand, but the error is not passed to MFA function? (e.g. `authenticateWithTOTP`)

features/26-mfa/api.md
```diff
@@ -0,0 +1,882 @@
...
+async function activateOOB(authenticatorID: string, code: string): Promise<ActivateOOBResult>;
+
+// Error inspection
+
+function isMFARequiredError(err: unknown): boolean;
+
+// Authenticate
+
+interface TOTPAuthenticateOptions {
+  authenticatorID?: string;
```
carmenlau: Per our discussion, in most of the case user will only have 1 authenticator. Maybe the authenticatorID can be optional if there is only one authenticator for the user?

features/26-mfa/api.md
louischan-oursky: Let me add a paragraph to explain how the client SDK should interact with the error.

features/26-mfa/api.md
louischan-oursky: > I'm thinking if there is any security implication for client side, would some how a social attack could be engineered to make users fake a OTP with a phishing website' issuer and account name? hmmm

The one-time password is short (6 digits) and valid for 30+-30s (1.5 min). So TOTP is subject to brute force attack if we do not have authentication attempt limits. We can discuss authentication attempt limits in another issue.

The details of security of TOTP (which is an extension of HOTP) can be found in the RFC

- https://tools.ietf.org/html/rfc4226#appendix-A.1
- https://tools.ietf.org/html/rfc4226#appendix-B.1

features/26-mfa/api.md
louischan-oursky: The value `optional` corresponds to Case 1.

---

Case 4 is impossible now because metadata can be freely set by the user. We need to have labels/tags that can be set on users only by the developer to achieve this.

---

For Case 2, I am not sure about the actual cases and here are my guess.

If the request is not from whitelisted IP address ranges, MFA is required. So the user must associate at least one MFA factor.

If the request is from whitelisted IP address ranges, MFA is optional.

This situation corresponds to `required` but conditionally `optional` when the request is whitelisted.

It seems to me that the application is targeting regional users. Do you know any real world examples?

One use case to whitelist IP so that MFA can be skipped I can think of is programmatic usage.

Google has a feature called App Password. App Password is available only when the account has MFA turned on. Authentication with App Password does not require MFA. App Password can also be revoked at wish.

If we have Google App Password in Skygear, then it would be a kind of Identity Provider, just like password, OAuth and Custom Token. But it is a special case that does not require MFA.

---

For Case 3, I think it is about re-authentication.

When a user is created, we must remember the request IP and the inferred country. The user does not setup MFA and is able to login as long as the country is the same. The user travels to another country. MFA is required now and the user has to setup MFA. It seems unnatural to me that MFA setup is delayed until the user has traveled to another country. When the user travels back to the original country, MFA has been setup but is not required during login. This contradicts with how real world applications work.

A more natural way is to require re-authentication when the inferred country is different from the country when the access token was issued. The access token includes the request IP and the inferred country. The server compares the request country and the claimed country in the access token. If they are different, then re-authentication is required. The re-authentication can start from the first factor such as password or directly from the second factor like TOTP.

louischan-oursky:
@chpapa @carmenlau @kiootic 

Most comments were addressed except for those that are not marked as resolved in this PR conversation.

features/26-mfa/implementation.md
```diff
@@ -0,0 +1,75 @@
...
+|code|TEXT|
+|created_at|timestamp|
+|consumed|BOOLEAN|
+
+- `consumed` default to `false`.
+- When a code is used to authenticated, it is marked as `consumed = true`.
+- A consumed code cannot be used anymore.
+- Recovery codes are never regenerated automatically. The user must regenerate themselves.
+
+## _auth_authenticator_bearer_token
```
kiootic: Do we associate a bearer token with single authenticator, so that it gets invalidated if that authenticator is removed?

features/26-mfa/implementation.md
```diff
@@ -0,0 +1,75 @@
...
+
+|Column|Type|
+|------|----|
+|id|TEXT|
+|user_id|TEXT; foreign key to `_core_user.id`|
+|code|TEXT|
+|created_at|timestamp|
+|expire_at|timestamp|
+
+- A OOB code is generated when trigger is called.
```
kiootic: Is that a non-expired OOB code would remains valid even if the authenticator that generated it is removed?

features/26-mfa/implementation.md
```diff
@@ -0,0 +1,75 @@
...
+- A OOB code is valid if it does not expire yet.
+- If a valid OOB code is found, no new code is generated. The valid code is delivered again.
```
kiootic: I'm not sure if this is common practice, but my intuition is that a new code is generated each time the user request a SMS code.

features/26-mfa/implementation.md
louischan-oursky: Sorry. Forgot this. I'll add it back.

features/26-mfa/implementation.md
louischan-oursky: My original thought is the OOB code is shared across all OOB authenticators. So trigger via different channels will deliver the same OOB code. If we want to have the behavior you mentioned, OOB code should be authenticator specific.

features/26-mfa/implementation.md
louischan-oursky: From my experience, the OOB code of other online services seems to be the same if triggered multiple times within a short period of time.

features/26-mfa/implementation.md
kiootic: So a user can use a SMS delivered code to authenticate as Email OOB authenticator? It seems strange that the `amr` claim does not necessarily reflect the actual delivery channel.

features/26-mfa/implementation.md
kiootic: Okay then.

features/26-mfa/implementation.md
louischan-oursky: `amr` does reflect the channel.

louischan-oursky:
- [ ] Add `off` to enforcement (default is `off`)
- [ ] Configure maximum authenticator per type. (upper bound is 10) (Default is 1)
- [ ] Design a way to omit authenticator ID in `authenticateWithXXX` API.
- [ ] If maximum OOB authenticator is 1, then `trigger` does not require authenticator ID.

features/26-mfa/api.md
chpapa: @louischan-oursky oh I should describe more about the use cases.

For Case 2, the use cases could be for example, if the user login from a company's premises, it doesn't need 2FA. It could be very meaningful if the 2FA is a hardware dongle (say when we support WebAuthn)

For Case 3, it is not about re-auth. It is supposed for end-users who already have 2FA setup, but usually it is not needed. But if it is logging in from another country, we will require 2FA again. In fact I'm kind of randomly come up with this requirement, because ultimately I think this option should change to "Only require MFA for abnormal login", so we should combine multiple factors (AI-blah-blah-blah) of login context to requires 2FA login. Kind of like what Google Login normally do after you login the first time on acomputer.

For Case 4, maybe it is still okay if the metadata can be changed? It is for more complex logic like, for example in Oursky.com. If a user don't have 2FA enabled, he can only access Email, but not Calendar or Google Drive. After a user have 2FA enabled, he can access everything. So programatically, I think we can have a metadata like role, and only certain value will require 2FA, and the authorization code can check if mfa is present for certain access control.

(But of coz, if the developer doesn't want role is changeable by the user alone, he will need something else...)

======

Alternatively, I think we could have 3 options for this option:

`required` / `optional` / `smart` (only for abnormal login, but now our logic to detect abnormal could be dumb lol)

features/26-mfa/api.md
chpapa: Oh sure... but it seems very important and we should mention the limits in this spec and create issues for repo?

louischan-oursky: I'll mention the limits in this spec and open a feature issue for authentication attempt limits.

louischan-oursky: Issue opened here https://github.com/SkygearIO/features/issues/344

louischan-oursky:
@chpapa @carmenlau @kiootic

I have updated the spec to include various improvements starting from [Document more TOTP characteristics and brute force attack](https://github.com/SkygearIO/features/pull/343/commits/249eecdeb3ce7c3ad7b48a122112280334b08328)

The remaining issue is whether to specify `enforcement: 'smart'` at this moment.

features/26-mfa/api.md
```diff
@@ -336,6 +336,10 @@ No input is required.
 
 No response body.
 
+### Note
```
kiootic: I assume if MFA is required/optional, this operation would still be denied if not yet authenticated with MFA?

features/26-mfa/api.md
```diff
@@ -765,6 +765,30 @@ const user = await skygear.auth.mfa.authenticateWithRecoveryCode(recoveryCode);
...
+### SDK
+
+```typescript
+// Suppose the user has lost their device or changed their phone number.
+// Under default settings, the number of authenticators is 1.
+// So the user should first remove the old authenticator otherwise
+// new registration will never succeed.
+
+// Present an UI to let the user to choose an authenticator to delete.
+const authenticators = await skygear.auth.mfa.getAuthenticators();
```
kiootic: If maximum number of authenticator is one, and that authenticator is lost, the user cannot be authenticated to perform this action?

features/26-mfa/api.md
```diff
@@ -37,6 +37,16 @@ mfa:
   #
   # Default is 'off'
   enforcement: 'off'
+  totp:
+    # The maximum number of TOTP authenticator the user can register.
+    # Valid values is [0, 10].
+    # Default is 1.
+    maximum: 1
```
kiootic: Do we need to provide a default value on maximum number of authenticators? If the user just opt-in with enforcement = optional/required, the default settings would require user to implement TOTP/OOB flow UI.

features/26-mfa/implementation.md
```diff
@@ -43,12 +43,14 @@
...
 |id|TEXT|
 |user_id|TEXT; foreign key to `_core_user.id`|
+|authenticator_id|TEXT; foreign key to `_auth_authenticator.id`|
 |code|TEXT|
 |created_at|timestamp|
 |expire_at|timestamp|
 
 - A OOB code is generated when trigger is called.
 - A OOB code is valid if it does not expire yet.
+- A OOB code is specific to an authenticator.
```
kiootic: We may want to note that the all valid OOB codes are unique, so that the correct OOB channel can be inferred just using the code.

features/26-mfa/api.md
louischan-oursky: They should authenticate with recovery code.

features/26-mfa/api.md
louischan-oursky: Introduce `maximum` to limit the number of authenticators. The default value is 1. Cap at 20. Change the default value of `totp.maximum` and `oob.maximum` to 0 so the developer must choose which authenticator they want to enable as enabling the authenticator requires the developer to implement a UI flow. Effectively, the developer must opt-in the authenticator they plan to support.

## At most 1 TOTP or 1 OOB

```yaml
mfa:
  enforcement: "optional"
  maximum: 1
  totp:
    maximum: 1
  oob:
    maximum: 1
```

## At most 1 TOTP and 1 OOB

```yaml
mfa:
  enforcement: "optional"
  maximum: 2
  totp:
    maximum: 1
  oob:
    maximum: 1
```

## At most 1 TOTP

```yaml
mfa:
  enforcement: "optional"
  totp:
    maximum: 1
```

## At most 1 OOB

```yaml
mfa:
  enforcement: "optional"
  oob:
    maximum: 1
```

features/26-mfa/api.md
```diff
@@ -153,12 +154,13 @@ mfa:
 ```JSON
 {
   "authenticator_id": "xxx",
-  "otp": "123456"
+  "otp": "123456",
```
kiootic: No need to have comma?

features/26-mfa/api.md
```diff
@@ -500,24 +526,42 @@ interface ActivateOOBResult {
...
+}
+
+function decodeAuthenticationSession(value: unknown): AuthenticationSession | undefined;
 
 // Authenticate
 
-interface MFAAuthenticateOptions {
+interface AuthenticateWithTOTPOptions {
+  otp: string;
+  authenticationSessionToken: string;
```
kiootic: I think the authentication session token can be handled automatically by SDK? So something like `currentAuthenticationStep` would be exposed, and developer does not need to pass the token around the different screens.

features/26-mfa/api.md
```diff
@@ -1040,26 +950,33 @@ async function login(loginID: string, password: string): Promise<User> {
...
   }
+  const { token: authn_session_token } = authnSession;
 
   // Retrieve the saved bearer token
-  let bearerToken;
+  // The token may be in cookie so it cannot be retrieved.
+  let bearer_token;
   try {
-    bearerToken = await getBearerToken();
+    bearer_token = await getBearerToken();
```
kiootic: Since the bearer token may be stored in cookie/local storage, I think it's better to let SDK manage it without exposing the token. The token itself should be meaningless to developer anyways.

features/26-mfa/api.md
louischan-oursky: The container is already very stateful. One of my experience on using the v1 SDK is that the record API is full of mutation. Maybe we do not expose the token but require the developer to pass in the authentication session. The authentication session in the SDK is like a opaque handle.

features/26-mfa/api.md
```diff
@@ -0,0 +1,1030 @@
...
+
+### Response body
+
+```JSON
+{
+  "recovery_codes": ["xxx", "yyy"]
+}
+```
+
+## POST /mfa/bearer_token/revoke_all
```
chpapa: Do we need to / have a bearer token default revoke time?

chpapa:
@louischan-oursky I notice this commit contradict with what we just discussed:

https://github.com/SkygearIO/skygear-server/commit/740af7b90ef39329a5975879151917831583be88

Shall we allow enforcement of MFA = off, but developers can still manually trigger MFA and treat it as one of the session header for authorization of specific APIs?

features/26-mfa/api.md
louischan-oursky: Yes. The configuration `mfa.bearer_token.expire_in_days` is `30` by default.

louischan-oursky:
> Shall we allow enforcement of MFA = off, but developers can still manually trigger MFA and treat it as one of the session header for authorization of specific APIs?

The purpose of `off` is to entirely forbid MFA. Since it is possible to register MFA authenticator if the configuration is not `off`, a clever user can register MFA authenticator via HTTP API. When the user login in the application, the application is not expecting MFA so the MFA flow is not properly handled.

To fulfill your use case, I think we should add one more option to enforcement.

`off`: No MFA authenticator can be added. MFA is not required in the authentication.
`optional`: MFA authenticator can be freely added. If the user has at least one activated authenticator, MFA is required in authentication.
`required`: MFA authenticator must be added. The user must authenticate with MFA.
`<name-to-be-confirmed>`: MFA authenticator can be freely added. But MFA is not required in the authentication even if the user has at least one activated authenticator.

The authentication is now driven by a authentication session token and the generation of such token is now implementation details. If the developer wants to use `<name-to-be-confirmed>`, we need to introduce an HTTP API to generate the authentication session token. The API is supposed to be used by microservice and requires master key. The API returns the authentication session token.

Since microservice is free to return any response they like, the client SDK **CANNOT** participate in manual MFA flow triggered by microservice. The current client SDK API design does not assume MFA can solely triggered by the auth gear. For example, it does not store the authentication session token nor the secret for the developer. The MFA part of the client SDK is stateless now, it does not store any state behind the scene. Therefore, as long as the developer can somehow get a authentication session token, they can make use of the client SDK to perform MFA when the enforcement is `<name-to-be-confirmed>`

---

If we want to introduce `<name-to-be-confirmed>`, I would like to rename some options to make them appear more intuitive.

`off` is still `off`
`required` changes to `always_required`
`optional` changes to `required_if_activated`
`<name-to-be-confirmed>` changes to `never_required`

features/26-mfa/api.md
```diff
@@ -0,0 +1,1030 @@
...
+  displayName,
+  authenticationSessionToken,
+});
+
+// Present to the secret to the user and let the user
+// to add the secret to their TOTP application, such as Authy and Google Authenticator.
+// Or preferably generate an otpauth URI.
+
+const otpauthURI = skygear.auth.mfa.generateOTPAuthURI({
+  secret,
```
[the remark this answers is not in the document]
louischan-oursky: I prefer we do not store `secret`.

One option is to require the developer to provide `issuer` and `accountName` in `createTOTP`. So the key URI is included in the result of `createTOTP`. The developer does not need to call `generateOTPAuthURI` themselves.

```typescript
const { otpauthURI, authenticatorID } = await skygear.auth.mfa.createNewTOTP({
  displayName,
  issuer: "My application",
  accountName: "user@example.com",
});

```

features/26-mfa/api.md
```diff
@@ -0,0 +1,1030 @@
...
+
+## Registration
+
+### SDK
+
+```typescript
+try {
+  await skygear.auth.login("user@example.com", "password");
+} catch (e) {
+  const authnSession = skygear.decodeAuthenticationSession(e);
```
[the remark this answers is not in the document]
louischan-oursky: I have removed `decodeAuthenticationSession` and instead store the authentication session in commit https://github.com/SkygearIO/features/pull/343/commits/0f556dbd6f65dedf6f42400e4c23b07573aa5e8e

The API is different from your suggestion because we do not have the convention of having a dedicated error class for each kind of error.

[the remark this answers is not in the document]
chpapa: Sorry, I didn't make it clear... my comment was more about, can we make `authenticationSession` transparent to developer users?

Right now they still need to understand the concept and use it from `skygear.auth.authenticationSession`?

The new code sample segment below make me feel really strange:

```
try {
  await skygear.auth.login("user@example.com", "password");
} catch (e) {
  const authnSession = skygear.auth.authenticationSession;
  if (authnSession != null && authnSession.step === "mfa") {
    navigateToRegisterMFAScreen(authnSession);
    return;
  }
  // Handle any other error.
}
```

The MFA login is triggered by `try catch`, but the hints of it is MFA related is totally unrelated with the exception?

So what is the convention of how we return each kind of error? Can we just follow that convention?

Apparently, we really have no reason why we need to make the user aware of the `authenticationSession`...

features/26-mfa/api.md
```diff
@@ -0,0 +1,1003 @@
...
+  enforcement: 'optional'
+  # The total maximum number of authenticator the user can register.
+  # Valid values is [0, 15].
+  # Default is 1.
+  maximum: 1
+  totp:
+    # The maximum number of TOTP authenticator the user can register.
+    # Valid values is [0, 5].
+    # Default is 0.
+    maximum: 0
```
chpapa: should we make a default value that once `enabled` is `true`, the default value could be used for normal case?

features/26-mfa/api.md
```diff
@@ -0,0 +1,1003 @@
...
+## Registration
+
+### SDK
+
+```typescript
+try {
+  await skygear.auth.login("user@example.com", "password");
+} catch (e) {
+  const authnSession = skygear.auth.authenticationSession;
+  if (authnSession != null && authnSession.step === "mfa") {
```
chpapa: As comments on the old version, the `catch` without using authentication feel weird.

And would love to see if we can eventually remove the whole `authenticationSession` concept from users

features/26-mfa/api.md
```diff
@@ -0,0 +1,1003 @@
...
+}
+
+async function getAuthenticators(): Promise<Authenticator[]>;
+async function deleteAuthenticator(authenticatorID: string): Promise<void>;
+async function regenerateRecoveryCode(): Promise<RegenerateRecoveryCodeResult>;
+async function listRecoveryCode(): Promise<ListRecoveryCodeResult>;
+async function revokeAllTrustedDevices(): Promise<void>;
+```
+
+# Use Cases
```
chpapa: Will we add a session about "Step Up MFA from an existing session"?

features/26-mfa/api.md
louischan-oursky: I want the developer to choose which authenticator their application supports. The UI of TOTP authenticator and Out-of-Band authenticator are different. If `enable` is `true`, the default value of TOTP authenticator and Out-of-Band authenticator both become `1`, the application is assumed to support both authenticators. I do not want to make such assumption.

If either one kind of authenticator becomes `1`, then the developer have to explicitly set it to `0` to disable it. That also means we flavor that kind of authenticator.

features/26-mfa/api.md
louischan-oursky: The reason why I added `authenticationSession` is that we want to support some application that may reload during the authentication session.

If authentication session is only stored in memory and the application reloads, the authentication session is clear. Effectively the user can never login.

If we choose not to support such use case, the authentication session can be hidden from the developer. We can use `e` and a helper function to tell the developer the step.

features/26-mfa/api.md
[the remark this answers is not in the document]
louischan-oursky: Please see the comment https://github.com/SkygearIO/features/pull/343#discussion_r324970903

louischan-oursky:
@chpapa @carmenlau 

I have added two changes to the spec that may need your attention.

The first one https://github.com/SkygearIO/features/pull/343/commits/ec4d670682370e2ef652416b6a27117081347770

The commits explains that the authentication session token is stateless and not invalidated after session is created. So it can be reused to create more sessions.

The second one https://github.com/SkygearIO/features/pull/343/commits/3229b44adaf56431bdce5da5e55036b0ed6f0f05

The commit limits the concurrency of registering an authenticator. In order to avoid the user to create many authenticators but never activate them, the API is changed to allow at most one pending inactive authenticator per kind. I think this limitation should not be a problem for common use cases because the user rarely registers multiple MFA authenticators concurrently.

carmenlau:
@louischan-oursky Both looks good to me:)

kiootic:
@chpapa the implementation is already merged. do you think this spec is okay?

features/26-mfa/api.md
```diff
@@ -0,0 +1,961 @@
...
+
+### Description
+
+Activate the just created inactive TOTP authenticator.
+
+## GET /mfa/totp/qrcode
+
+## Query parameters
+
+- `otpauth_uri`: The otpauth URI
```
chpapa: usability-wise, would it be better if we let people do either `otpath_uri`, or:

`label`: Lable of the account
`secret`: secret (or `authenticator_id`?)
`issuer`: provider or the service this account is associated with

features/26-mfa/api.md
```diff
@@ -0,0 +1,961 @@
...
+- `display_name`: The display name of the new TOTP authenticator.
+- `authn_session_token`: Required only when there is no access token.
+
+### Response body
+
+```JSON
+{
+  "authenticator_id": "xxx",
+  "authenticator_type": "totp",
+  "secret": "secret"
```
chpapa: in real world use cases, is it easier if user provide a optional `label` and `issuer` to us, we can respond the QR Code image URL at once here?

features/26-mfa/api.md
```diff
@@ -0,0 +1,961 @@
...
+
+### Response body
+
+```JSON
+{
+  "recovery_codes": ["xxx", "yyy"]
+}
+```
+
+## GET /mfa/recovery_code
```
chpapa: Am I right to assume to get recovery code / regenerate, a recent auth session is needed (so users will need to type password again if their sessions is not recently authed?)

How these requirements are recorded in spec for now? @louischan-oursky

features/26-mfa/api.md
louischan-oursky: We have [this function](https://skygeario.github.io/skygear-SDK-JS/docs/web.generateotpauthuri) in the client SDK, which effectively turn [label, secret and issuer](https://skygeario.github.io/skygear-SDK-JS/docs/web.generateotpauthurioptions) into `otpauth_uri`.

`authenticator_id` should **NOT** be used because it should be impossible to reveal the secret after the creation. The secret is only revealed once during creation.

features/26-mfa/api.md
louischan-oursky: Agree we should have re-authentication for security-related operation like this.

The re-authentication is briefly mentioned [here](https://github.com/SkygearIO/features/blob/master/features/329-identity-apis/api.md#configuration). Maybe we should move the spec of re-authentication to its own document.

Should I open a feature issue for this?

features/26-mfa/api.md
louischan-oursky: Your two comments are about the convenience of accepting either `otpauth_uri` or `(accountName, issuer)`. I agree it is more ergonomic to have this convenience. Let me include in the spec and open an issue to implement it.

features/26-mfa/api.md
```diff
@@ -188,7 +188,15 @@ Activate the just created inactive TOTP authenticator.
...
 
-- `otpauth_uri`: The otpauth URI
+- `otpauth_uri`: The [key URI](https://github.com/google/google-authenticator/wiki/Key-Uri-Format).
+- `account_name`: The [account name](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#label).
+- `secret`: The [Base32 encoded secret](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#secret).
+- `issuer`: The [issuer](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#label).
+
+If `otpauth_uri` is given, it is used as is. Otherwise `otpauth_uri` is generated from `account_name`, `secret` and `issuer`.
+
+This endpoint accepts `secret` instead of `authenticator_id` because this endpoint is supposed to be used directly,
```
chpapa: Oops, I totally didn't get it before you spell it out.

Ummm, do we really want to encourage that? Coz I think embed the secret in image URL is not a good practices and we shouldn't encourage users to do so... (since URL could be tracked by browsers / extensions / server side logs etc)

I think it is better to display the QR code by:

1. Users call the APIs, get some sort of `data:image/png` format, and render it, OR
2. Get the png from a stateful image URL

chpapa: Or you have researched and it is commonly done in this way? @louischan-oursky

features/26-mfa/api.md
chpapa: Sure Thx.

features/26-mfa/api.md
```diff
@@ -0,0 +1,975 @@
...
+- `authn_session_token`: Required only when there is no access token.
+
+### Response body
+
+```JSON
+{
+  "authenticator_id": "xxx",
+  "authenticator_type": "totp",
+  "secret": "secret",
+  "otpauth_uri": "otpauth://totp/MyApp:user@example.com?secret=secret&issuer=MyApp"
```
chpapa: Following up from the comment of avoiding secret in GET parameter as part of URL, do we really need to send the secret as string to client side at all? Maybe make secret "optional" and only return in response if the client want it? In normal flow I think the client side only need the QR Code (which itself have the secrets already)

I might have missed something! Just brainstorming...

features/26-mfa/api.md
```diff
@@ -0,0 +1,975 @@
...
+
+- `otpauth_uri`: The [key URI](https://github.com/google/google-authenticator/wiki/Key-Uri-Format).
+- `account_name`: The [account name](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#label).
+- `secret`: The [Base32 encoded secret](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#secret).
+- `issuer`: The [issuer](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#label).
+
+If `otpauth_uri` is given, it is used as is. Otherwise `otpauth_uri` is generated from `account_name`, `secret` and `issuer`.
+
+This endpoint accepts `secret` instead of `authenticator_id` because this endpoint is supposed to be used directly,
+such as setting it to the `src` of `<img>`. So it is stateless and does not require authentication.
```
chpapa: Oops, I totally didn't get it before you spell it out.

Ummm, do we really want to encourage that? Coz I think embed the secret in image URL as GET Parameter is not a good practices and we shouldn't encourage users to do so... (since URL could be tracked by browsers / extensions / server side logs etc)

I think it is better to display the QR code by:

1. Users call the APIs, get some sort of data:image/png format, and render it, OR
1. Get the png from a stateful image URL

@louischan-oursky or you have done more research and found that it is the best practices?

features/26-mfa/api.md
louischan-oursky: Opened [a feature issue](https://github.com/SkygearIO/features/issues/367)

features/26-mfa/api.md
louischan-oursky: > Users call the APIs, get some sort of data:image/png format, and render it

Here is my proposal

-  Remove `/mfa/totp/qrcode` endpoint so there is no chance the secret appear in URL. 
-  The QR code is a image representation of the `otpauth_uri`. `otpauth_uri` includes `secret` in its query string. The three are equivalent. Knowing one of them allows deriving the secret. We want to make the API both flexible and ergonomic. Therefore, in `/mfa/totp/new`, we should 
    - Include `secret`, which allows the developer to display it in a way they want.
    - Include `otpauth_uri`, which is correctly formatted Key URI so that it is ready to generate QR code in the developer's own way.
    - Include `qr_code_image_uri`, which is data URI of a server generated QR code image of the secret. In most case, the developer just needs this.

None of the above is optional. They always appear in the response.

> Get the png from a stateful image URL

The secret should not be ever revealed again. Its sole appearance is in the response of `/mfa/totp/new`

louischan-oursky:
Is it OK to merge this?

Issue #255, opened 2018-11-06
Use cookies and sessions instead of LocalStorage and JWT for session


chpapa:
# Description

Would like to see if we will consider doing this for Skygear NEXT since it will break some backward compatibility:

Reasons for the switch:
* To make it easier/possible to invalidate user login session from server side
* Support this feature: https://github.com/SkygearIO/features/issues/202
* (Maybe?) Make SSR easier: https://github.com/SkygearIO/skygear-SDK-JS/issues/301

Some reference read:
* http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/
* https://kev.inburke.com/kevin/things-to-use-instead-of-jwt/

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

Issue #202, opened 2018-04-09
List login sessions and invalidate all other / individual session from SDK


chpapa:
# Description

It is a common requirement to manage login sessions / invalidate all other or individual session. 

## Scenario

1. Have a setting page which lists all login user session with the following data. And able to log each session out, or have a button to log ALL other session out.
    * Last location used
    * Last time used (or Active now)
    * Last IP addresses
    * Device type(?) / Browser Type
1. After some operation, such as validate a new phone number, need to invalidate all sessions.

# API Design

Something like these?

* Client side SDK (JS, iOS, Android):
  * `skygear.auth.listSessions()` => return an array of dicts.
  * `skygear.auth.invalidateSession(sessionToken)`
  * `sykgear.auth.invalidateAllSessions(exceptCurrentSession=true)`
* The above API should be usable at Cloud Functions too

# Open Questions

* Maybe it doesn’t make sense to make JWT stateful just for invalidate sessions? Could consider alternatives such as using a short auth token + long refresh token with blacklist on refresh token / login. Or other idea (search online?)
* For login sessions history, shall we use redis or postgresql?

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

carmenlau:
Would like to bring this up, we had a short discussion within the team. For the session management, we may also want to support refresh token. If so, we may need to have it before the first launch as this should be a breaking change. Can further discuss in the product meeting.

louischan-oursky:
The client sdk should have a callback mechanism to handle token expiration error.

PR #350, opened 2019-08-09
Session management

kiootic:
ref #202 
ref #255 

[Rendered Version](https://github.com/kiootic/skygear-features/blob/202-session-management/features/202-session-management/design.md)

### DIFF
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -75,13 +75,13 @@ information may not be backward compatible.
 
 ## SDK Extra Information Collection
 Extra information collection is disabled by default. If enabled, the collected
-information would be sent in header `X-Skygear-Extra-Info` as JSON encoded object.
+information would be sent in header `X-Skygear-Extra-Info` as base64 encoded JSON encoded object.
 
 Client SDKs should persist the extra information configuration locally.
 
 Example header:
 ```
-X-Skygear-Extra-Info: { "device_name": "My Phone" }
+X-Skygear-Extra-Info: eyAiZGV2aWNlX25hbWUiOiAiTXkgUGhvbmUiIH0K
 ```
 
 ## Web-hook Events
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -34,3 +34,14 @@ class AuthContainer {
     get currentSessionID(): string | null;
 }
 ```
+
+## SDK User Agent
+
+In general, for native SDKs, the user agent would have following format:
+```
+[App ID]/[App Version]([Device Model]; [Device OS]) [SDK Library Name]/[SDK version]
+```
+
+Examples:
+- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
+- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -2,6 +2,18 @@
 
 ## SDK API
 ```typescript
+interface UserAgent {
+    raw: string;
+    name: string;
+    version: string;
+    os: string;
+    osVersion: string;
+    app: string;
+    appVersion: string
+    device: string;
+    deviceID: string;
+}
+
 interface Session {
     id: string;
     identityID: string;
@@ -9,8 +21,9 @@ interface Session {
     lastAccessedAt: Date;
     createdByIP: string;
     lastAccessedIP: string;
-    createdByUserAgent: string;
-    lastAccessedUserAgent: string;
+    createdByUserAgent: UserAgent;
+    lastAccessedUserAgent: UserAgent;
+    name: string;
     data: object;
 }
 
@@ -21,8 +34,9 @@ function listSessions(): Promise<Session[]>;
 // for master-key, can use for any user
 function getSession(sessionID: string): Promise<Session>;
 
-// this is master-key only API
-function updateSession(sessionID: string, data: object): Promise<void>;
+// update specific session name for current user
+// for master-key, can use for any user and also update custom attributes
+function updateSession(sessionID: string, name: string | undefined, data?: object): Promise<void>;
 
 function revokeOtherSessions(): Promise<void>;
 
@@ -39,9 +53,14 @@ class AuthContainer {
 
 In general, for native SDKs, the user agent would have following format:
 ```
-[App ID]/[App Version]([Device Model]; [Device OS]) [SDK Library Name]/[SDK version]
+[App ID]/[App Version]([Device]; [OS]; [Device ID]) [SDK Library Name]/[SDK version]
 ```
 
 Examples:
-- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
-- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
+- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0; 46FF79AA-4E2E-475F-B58F-E7E37BAB771C) SKYKit/2.0.1`
+- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0; 32DDE92083E1AA54) io.skygear.skygear/2.2.0`
+
+The parsed user agent information is provided in best-effort basis, and its
+information may not be backward compatible.
+Some parsed user agent information may not be available, and empty string
+would be used instead.
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -21,31 +21,11 @@ interface Session {
     createdByIP: string;
     lastAccessedIP: string;
     userAgent: UserAgent;
-    name: string;
-    data: object;
 }
 
 interface ExtraSessionInfoOptions {
-    collectDeviceName: boolean; // indicate SDK should collect device name information
-    deviceName?: string; // if falsy, SDK would try to detect it if possible
+    deviceName?: string;
 }
-
-// list all sessions of current user
-function listSessions(): Promise<Session[]>;
-
-// get specific session information for current user
-// for master-key, can use for any user
-function getSession(sessionID: string): Promise<Session>;
-
-// update specific session name for current user
-// for master-key, can use for any user and also update custom attributes
-function updateSession(sessionID: string, patch: { name?: string; data?: JSONObject }): Promise<void>;
-
-function revokeOtherSessions(): Promise<void>;
-
-// revoking current session is disallowed
-// for master-key, can use for any session
-function revokeSession(sessionID: string): Promise<void>;
 
 class AuthContainer {
     get currentSessionID(): string | null;
@@ -55,7 +35,23 @@ class AuthContainer {
     // after updating extra session info options, call this to save it to
     // persistent storage.
     saveExtraSessionInfoOptions(): Promise<void>;
+
+    // list all sessions of current user
+    listSessions(): Promise<Session[]>;
+
+    // get specific session information for current user
+    // for master-key, can use for any user
+    getSession(sessionID: string): Promise<Session>;
+
+    revokeOtherSessions(): Promise<void>;
+
+    // revoking current session is disallowed
+    // for master-key, can use for any session
+    revokeSession(sessionID: string): Promise<void>;
 }
+
+// get current device name; return empty string if cannot detect.
+function getDeviceName(): Promise<string>;
 ```
 
 ## SDK User Agent
@@ -107,5 +103,3 @@ added:
 
 For `session_delete` event, the `reason` field will have one more possible
 value `revoke`.
-
-
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -51,6 +51,10 @@ class AuthContainer {
     get currentSessionID(): string | null;
 
     extraSessionInfoOptions: ExtraSessionInfoOptions;
+
+    // after updating extra session info options, call this to save it to
+    // persistent storage.
+    saveExtraSessionInfoOptions(): Promise<void>;
 }
 ```
 
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -54,8 +54,12 @@ In general, for native SDKs, the user agent would have following format:
 ```
 
 Examples:
-- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
-- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
+- `io.skygear.test/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
+- `io.skygear.test/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
+
+For web SDK, the user agent would be controlled by browser, for example:
+
+`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36`
 
 The parsed user agent information is provided in best-effort basis, and its
 information may not be backward compatible.
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -1,0 +1,26 @@
+# Session Management APIs
+
+## SDK API
+```typescript
+interface Session {
+    id: string;
+    identityID: string;
+    createdAt: Date;
+    lastAccessAt: Date;
+    createdByIP: string;
+    lastAccessIP: string;
+    createdByUserAgent: string;
+    lastAccessUserAgent: string;
+    data: object;
+}
+
+function listSessions(): Promise<Session[]>;
+
+function revokeOtherSessions(): Promise<void>;
+
+function revokeSession(sessionID: string): Promise<void>;
+
+class AuthContainer {
+    get currentSessionID(): string | null;
+}
+```
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -25,8 +25,8 @@ interface Session {
     data: object;
 }
 
-interface ExtraSessionInfoConfig {
-    deviceNameEnabled: boolean; // indicate SDK should collect device name information
+interface ExtraSessionInfoOptions {
+    collectDeviceName: boolean; // indicate SDK should collect device name information
     deviceName?: string; // if falsy, SDK would try to detect it if possible
 }
 
@@ -39,7 +39,7 @@ function getSession(sessionID: string): Promise<Session>;
 
 // update specific session name for current user
 // for master-key, can use for any user and also update custom attributes
-function updateSession(sessionID: string, name: string | undefined, data?: object): Promise<void>;
+function updateSession(sessionID: string, patch: { name?: string; data?: JSONObject }): Promise<void>;
 
 function revokeOtherSessions(): Promise<void>;
 
@@ -50,7 +50,7 @@ function revokeSession(sessionID: string): Promise<void>;
 class AuthContainer {
     get currentSessionID(): string | null;
 
-    extraSessionInfoConfig: ExtraSessionInfoConfig;
+    extraSessionInfoOptions: ExtraSessionInfoOptions;
 }
 ```
 
@@ -58,12 +58,12 @@ class AuthContainer {
 
 In general, for native SDKs, the user agent would have following format:
 ```
-[App ID]/[App Version]([Device]; [OS]) [SDK Library Name]/[SDK version]
+[App ID]/[App Version](Skygear; [Device]; [OS]) [SDK Library Name]/[SDK version]
 ```
 
 Examples:
-- `io.skygear.test/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
-- `io.skygear.test/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
+- `io.skygear.test/1.0.1 (Skygear; iPhone11,8; iOS 12.0) SKYKit/2.0.1`
+- `io.skygear.test/1.3.0 (Skygear; Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
 
 For web SDK, the user agent would be controlled by browser, for example:
 
@@ -79,6 +79,10 @@ information would be sent in header `X-Skygear-Extra-Info` as JSON encoded objec
 
 Client SDKs should persist the extra information configuration locally.
 
+Example header:
+```
+X-Skygear-Extra-Info: { "device_name": "My Phone" }
+```
 
 ## Web-hook Events
 In event context, a `session` attribute with value type `Session` would be
@@ -100,14 +104,4 @@ added:
 For `session_delete` event, the `reason` field will have one more possible
 value `revoke`.
 
-A new event `session_update` would be added with payload:
-```json
-{
-    "name": "",
-    "data": {}
-}
-```
-`name` and `data` are the new name and custom attributes respectively. If the
-field is not updated, `null` would be used as value. The session attributes in
-event context is the old value.
 
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -65,3 +65,36 @@ The parsed user agent information is provided in best-effort basis, and its
 information may not be backward compatible.
 Some parsed user agent information may not be available, and empty string
 would be used instead.
+
+
+## Web-hook Events
+In event context, a `session` attribute with value type `Session` would be
+added:
+```json
+{
+    "timestamp": 1565859018,
+    "request_id": "66014B80-0813-481B-9CD9-7C5A12B4671F",
+    "user_id": "F973C930-D357-428E-8E6D-A32B0FF1CFF4",
+    "identity_id": "1658AC0E-1D62-4612-BDE3-31464121C9A3",
+    "session": {
+        "id": "FEAD3339-4A26-4C43-8737-27AC88D1173E",
+        "identity_id": "1658AC0E-1D62-4612-BDE3-31464121C9A3"
+        // ...
+    }
+}
+```
+
+For `session_delete` event, the `reason` field will have one more possible
+value `revoke`.
+
+A new event `session_update` would be added with payload:
+```json
+{
+    "name": "",
+    "data": {}
+}
+```
+`name` and `data` are the new name and custom attributes respectively. If the
+field is not updated, `null` would be used as value. The session attributes in
+event context is the old value.
+
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -2,6 +2,7 @@
 
 ## SDK API
 ```typescript
+// Field value would be empty string if not available
 interface UserAgent {
     raw: string;
     name: string; // browser name or app ID
@@ -11,6 +12,11 @@ interface UserAgent {
     device: string;
 }
 
+// Field value would be empty string if not available
+interface Device {
+    name: string; // device name
+}
+
 interface Session {
     id: string;
     identityID: string;
@@ -18,10 +24,15 @@ interface Session {
     lastAccessedAt: Date;
     createdByIP: string;
     lastAccessedIP: string;
-    createdByUserAgent: UserAgent;
-    lastAccessedUserAgent: UserAgent;
+    userAgent: UserAgent;
+    device: Device?;
     name: string;
     data: object;
+}
+
+interface ExtraSessionInfoConfig {
+    enabled: boolean; // indicate whether extra information is enabled, default false
+    deviceName: string; // if empty, SDK would try to detect it if possible
 }
 
 // list all sessions of current user
@@ -43,6 +54,8 @@ function revokeSession(sessionID: string): Promise<void>;
 
 class AuthContainer {
     get currentSessionID(): string | null;
+
+    extraSessionInfoConfig: ExtraSessionInfoConfig;
 }
 ```
 
@@ -63,8 +76,14 @@ For web SDK, the user agent would be controlled by browser, for example:
 
 The parsed user agent information is provided in best-effort basis, and its
 information may not be backward compatible.
-Some parsed user agent information may not be available, and empty string
-would be used instead.
+
+
+## SDK Extra Information Collection
+Extra information collection is disabled by default. If enabled, the collected
+information would be sent in cookie `SKYGEAR_SESS_INFO` as JSON encoded object.
+
+Extra information collection configuration would be persisted across app
+restart.
 
 
 ## Web-hook Events
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -4,14 +4,11 @@
 ```typescript
 interface UserAgent {
     raw: string;
-    name: string;
-    version: string;
+    name: string; // browser name or app ID
+    version: string; // browser version or app version
     os: string;
     osVersion: string;
-    app: string;
-    appVersion: string
     device: string;
-    deviceID: string;
 }
 
 interface Session {
@@ -53,12 +50,12 @@ class AuthContainer {
 
 In general, for native SDKs, the user agent would have following format:
 ```
-[App ID]/[App Version]([Device]; [OS]; [Device ID]) [SDK Library Name]/[SDK version]
+[App ID]/[App Version]([Device]; [OS]) [SDK Library Name]/[SDK version]
 ```
 
 Examples:
-- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0; 46FF79AA-4E2E-475F-B58F-E7E37BAB771C) SKYKit/2.0.1`
-- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0; 32DDE92083E1AA54) io.skygear.skygear/2.2.0`
+- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0) SKYKit/2.0.1`
+- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0) io.skygear.skygear/2.2.0`
 
 The parsed user agent information is provided in best-effort basis, and its
 information may not be backward compatible.
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -6,22 +6,28 @@ interface Session {
     id: string;
     identityID: string;
     createdAt: Date;
-    lastAccessAt: Date;
+    lastAccessedAt: Date;
     createdByIP: string;
-    lastAccessIP: string;
+    lastAccessedIP: string;
     createdByUserAgent: string;
-    lastAccessUserAgent: string;
+    lastAccessedUserAgent: string;
     data: object;
 }
 
+// list all sessions of current user
 function listSessions(): Promise<Session[]>;
 
+// get specific session information for current user
+// for master-key, can use for any user
 function getSession(sessionID: string): Promise<Session>;
 
+// this is master-key only API
 function updateSession(sessionID: string, data: object): Promise<void>;
 
 function revokeOtherSessions(): Promise<void>;
 
+// revoking current session is disallowed
+// for master-key, can use for any session
 function revokeSession(sessionID: string): Promise<void>;
 
 class AuthContainer {
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -9,12 +9,8 @@ interface UserAgent {
     version: string; // browser version or app version
     os: string;
     osVersion: string;
-    device: string;
-}
-
-// Field value would be empty string if not available
-interface Device {
-    name: string; // device name
+    deviceName: string;
+    deviceModel: string;
 }
 
 interface Session {
@@ -25,14 +21,13 @@ interface Session {
     createdByIP: string;
     lastAccessedIP: string;
     userAgent: UserAgent;
-    device: Device?;
     name: string;
     data: object;
 }
 
 interface ExtraSessionInfoConfig {
-    enabled: boolean; // indicate whether extra information is enabled, default false
-    deviceName: string; // if empty, SDK would try to detect it if possible
+    deviceNameEnabled: boolean; // indicate SDK should collect device name information
+    deviceName?: string; // if falsy, SDK would try to detect it if possible
 }
 
 // list all sessions of current user
@@ -80,10 +75,9 @@ information may not be backward compatible.
 
 ## SDK Extra Information Collection
 Extra information collection is disabled by default. If enabled, the collected
-information would be sent in cookie `SKYGEAR_SESS_INFO` as JSON encoded object.
+information would be sent in header `X-Skygear-Extra-Info` as JSON encoded object.
 
-Extra information collection configuration would be persisted across app
-restart.
+Client SDKs should persist the extra information configuration locally.
 
 
 ## Web-hook Events
--- a/features/202-session-management/api.md
+++ b/features/202-session-management/api.md
@@ -16,6 +16,10 @@ interface Session {
 
 function listSessions(): Promise<Session[]>;
 
+function getSession(sessionID: string): Promise<Session>;
+
+function updateSession(sessionID: string, data: object): Promise<void>;
+
 function revokeOtherSessions(): Promise<void>;
 
 function revokeSession(sessionID: string): Promise<void>;
--- a/features/202-session-management/config.md
+++ b/features/202-session-management/config.md
@@ -9,20 +9,19 @@ clients:
       disabled: false
       api_key: XXX
       session_transport: cookie
+      access_token_lifetime: 1800
+      session_idle_timeout_enabled: true
+      session_idle_timeout: 300
+      same_site: lax
     - name: iOS App
       disabled: false
       api_key: YYY
       session_transport: header
-    - name: Android App
-      disabled: false
-      api_key: ZZZ
-      session_transport: header
-session:
-    access_token_lifetime: 1800
-    refresh_token_disabled: false
-    refresh_token_lifetime: 86400
-    session_idle_timeout_enabled: true
-    session_idle_timeout: 300
+      access_token_lifetime: 1800
+      refresh_token_disabled: false
+      refresh_token_lifetime: 86400
+      session_idle_timeout_enabled: false
+      session_idle_timeout: 300
 ```
 
 ## User Configuration
@@ -33,15 +32,15 @@ The value of key `clients` is a list of client configuration:
 - `api_key`: API key.
 - `session_transport`: The transport method of session tokens.
                        Can be `cookie` or `header`.
-
-The value of key `session` is a map of session configuration:
 - `access_token_lifetime`: The lifetime of access token in seconds, default
                            to 1800.
-- `refresh_token_disabled`: Indicate whether refresh token is disabled, default
+- `refresh_token_disabled`: (valid for `header` transport only)
+                            Indicate whether refresh token is disabled, default
                             to `false`. If `session_transport` is `cookie`,
                             refresh token is disabled and this configuration
                             has no effect.
-- `refresh_token_lifetime`: The maximum lifetime of refresh token in seconds,
+- `refresh_token_lifetime`: (valid for `header` transport only)
+                            The maximum lifetime of refresh token in seconds,
                             default to max(`access_token_lifetime`, 86400).
                             Must greater than or equal to `access_token_lifetime`.
 - `session_idle_timeout_enabled`: Indicate whether session idle timeout is
@@ -49,3 +48,10 @@ The value of key `session` is a map of session configuration:
 - `session_idle_timeout`: The session idle timeout in seconds,
                           default to min(`access_token_lifetime`, 300).
                           Must less than or equal to `access_token_lifetime`.
+- `same_site`: (valid for `cookie` transport only)
+               The `SameSite` property of cookie. Can be `lax`, `strict`, or
+               empty string. Default to `lax`.
+
+## Auth Gear standalone configuration
+- `INSECURE_COOKIE`: Indicate whether session cookie should not set the `Secure`
+                     flag. Default to `false`.
--- a/features/202-session-management/config.md
+++ b/features/202-session-management/config.md
@@ -50,7 +50,7 @@ The value of key `clients` is a list of client configuration:
                           Must less than or equal to `access_token_lifetime`.
 - `same_site`: (valid for `cookie` transport only)
                The `SameSite` property of cookie. Can be `lax`, `strict`, or
-               empty string. Default to `lax`.
+               `none`. Default to `lax`.
 
 ## Auth Gear standalone configuration
 - `INSECURE_COOKIE`: Indicate whether session cookie should not set the `Secure`
--- a/features/202-session-management/config.md
+++ b/features/202-session-management/config.md
@@ -1,0 +1,51 @@
+# Session Configuration
+
+## Sample configuration
+
+In user configuration:
+```yaml
+clients:
+    - name: Web App
+      disabled: false
+      api_key: XXX
+      session_transport: cookie
+    - name: iOS App
+      disabled: false
+      api_key: YYY
+      session_transport: header
+    - name: Android App
+      disabled: false
+      api_key: ZZZ
+      session_transport: header
+session:
+    access_token_lifetime: 1800
+    refresh_token_disabled: false
+    refresh_token_lifetime: 86400
+    session_idle_timeout_enabled: true
+    session_idle_timeout: 300
+```
+
+## User Configuration
+
+The value of key `clients` is a list of client configuration:
+- `name`: Name of client. Show in UI (e.g. portal)
+- `disabled`: Indicate whether the client is disabled.
+- `api_key`: API key.
+- `session_transport`: The transport method of session tokens.
+                       Can be `cookie` or `header`.
+
+The value of key `session` is a map of session configuration:
+- `access_token_lifetime`: The lifetime of access token in seconds, default
+                           to 1800.
+- `refresh_token_disabled`: Indicate whether refresh token is disabled, default
+                            to `false`. If `session_transport` is `cookie`,
+                            refresh token is disabled and this configuration
+                            has no effect.
+- `refresh_token_lifetime`: The maximum lifetime of refresh token in seconds,
+                            default to max(`access_token_lifetime`, 86400).
+                            Must greater than or equal to `access_token_lifetime`.
+- `session_idle_timeout_enabled`: Indicate whether session idle timeout is
+                                  enabled, default to `false`.
+- `session_idle_timeout`: The session idle timeout in seconds,
+                          default to min(`access_token_lifetime`, 300).
+                          Must less than or equal to `access_token_lifetime`.
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -15,7 +15,8 @@ Sessions of auth gear should be configurable and manageable.
 ## Stateful Session
 
 Sessions would be stateful (i.e. persisted into database), instead of being
-stateless.
+stateless. Sessions would be stored in a low-latency horizontally scalable
+store for scalability reasons.
 
 Each session would be associated with attributes:
 - session ID
@@ -149,6 +150,11 @@ specific client, and different attributes can be configured for the client:
 
 Session token transport must be specified at client creation. It cannot be
 changed after creation.
+
+
+# Future Works
+- Audit log should include session attributes for historical records (#340)
+- A country value should be provided, deriving from available information (#354)
 
 
 # Appendix
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -32,9 +32,8 @@ Each session would be associated with attributes:
 - name
 - custom attributes
 
-Session attributes can be manipulated through management APIs. Session ID would
-be included in web-hook event, so developer can read/write session attributes
-if needed.
+Session attributes can be manipulated through management APIs and web-hook
+handler.
 
 Each session is associated with a name, default to empty. User can update the
 session name using SDK. This name is shown in portal UI.
@@ -58,6 +57,10 @@ to be revoked.
 
 Update session API allows updating name of session. Custom attributes can also
 be updated using master key.
+
+When the name of session, or custom attributes are updated, a `session_update`
+event would be generated. Web-hook event context would include session
+attributes and mutations can be used to update them.
 
 
 ## Access / Refresh Token
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -105,6 +105,7 @@ If developer choose to enable session cookie:
 - Access token would not be returned from APIs.
 - Gateway would read access token from cookie only, ignoring `Authorization`
   header.
+- Some SDKs (e.g. mobile SDKs) would not be able to consume it.
 
 ### Security Consideration
 
@@ -123,12 +124,11 @@ If developer choose to enable session cookie:
 
 Optionally, developer can enable extra information collection at client side.
 If enabled, client SDK would collect extra information about user and send them
-to server in a cookie. This collection is on best-effort basis, and fields are
+to server in a header. This collection is on best-effort basis, and fields are
 not guaranteed to have values.
 
 This information would be associated with the session. Server would update
-the session with the information provided in cookie every request (erase if
-not present).
+the session with the information provided in header (if present) every request.
 
 Following information would be collected at the moment:
 - Device name
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -119,6 +119,21 @@ If developer choose to enable session cookie:
   tenant, the top-level domain must be included in Public Suffix List.
 
 
+## Extra Information Collection
+
+Optionally, developer can enable extra information collection at client side.
+If enabled, client SDK would collect extra information about user and send them
+to server in a cookie. This collection is on best-effort basis, and fields are
+not guaranteed to have values.
+
+This information would be associated with the session. Server would update
+the session with the information provided in cookie every request (erase if
+not present).
+
+Following information would be collected at the moment:
+- Device name
+
+
 ## API Key per Client
 
 Each app can support multiple clients, such as web client and mobile client.
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -1,0 +1,163 @@
+# Background
+
+Sessions of auth gear should be configurable and manageable.
+
+
+# Objectives
+
+- App Developer/User should be able to manage sessions.
+- App Developer should be able to configure session lifetime.
+- App Developer should be able to configure session token transport.
+
+
+# Proposed Design
+
+## Stateful Session
+
+Sessions would be stateful (i.e. persisted into database), instead of being
+stateless.
+
+Each session would be associated with attributes:
+- session ID
+- app ID
+- user ID
+- identity ID
+- expiration time
+- creation time
+- last access time
+- creation IP
+- last IP
+- creation user agent
+- last user agent
+- custom attributes
+
+Session attributes would be available in web-hook event context, and session
+management APIs.
+
+App developer can update custom attributes in session using web-hook event
+mutation.
+
+Gateway is responsible to update last access time, last IP, and last user agent
+of session after processing the request, regardless whether the request is
+successful.
+
+
+## Session Management
+
+APIs would be provided to manage sessions of a user.
+
+Session list API returns a list of valid session of a user. Each list entry
+contains the session ID and session attributes. Access token / refresh token of
+the session would not be included.
+
+Revoke sessions API revokes a specific session, or all sessions other than the
+requesting session. A `session_delete` event would be generated for all sessions
+to be revoked.
+
+
+## Access / Refresh Token
+
+Each session would have an access token, and optionally a refresh token. Both
+type of tokens are opaque: token formats are unspecified and developer should
+not attempt to interpret the content of tokens.
+
+Access token would always have a lifetime; access token would be treated as
+invalid after its expiry.
+
+Refresh token can be used to obtain a new access token. If the old access token
+is still valid, the old access token is invalidated; there would be at most one
+valid access token for a session at any time.
+
+Refresh token would always have a maximum lifetime.
+
+Optionally, a session idle timeout can be specified: refresh token (or access
+token if refresh token is disabled) must be used at least once before the
+timeout, otherwise the session would be expired.
+
+A session is invalidated if its identity is deleted, or its user is disabled.
+A session is expired if its refresh token is expired, or its access token is
+expired (if refresh token is disabled).
+If a session is invalidated, expired, revoked, or logout, its associated access
+token / refresh token would be treated as invalid.
+
+Client-side SDK should handle refreshing of expired access token automatically
+if refresh token is available; SDKs would not expose concept of access token /
+refresh token; however, developers should still handle situation where session
+is no longer valid due to various reasons.
+
+
+## Session Cookie
+
+Optionally, developer can choose to use cookies to transport session tokens, in
+order to facilitate server-side rendering.
+
+If developer choose to enable session cookie:
+- Refresh token would not be issued. (See [Appendix](#session-cookie-and-refresh-token))
+- Access token would be transported in session cookie.
+- Access token would not be returned from APIs.
+- Gateway would read access token from cookie only, ignoring `Authorization`
+  header.
+
+### Security Consideration
+
+- HTTPS is required.
+- Session cookie is HTTP-only; web app and web SDK would not have access to the
+  session tokens.
+- Session cookie is set with `SameSite=lax`; generally, cross-domain requests
+  would not contain the cookie.
+- Session cookie would not be shared across custom domains.
+- App developers are responsible to prevent CSRF in their services.
+- For untrusted multi-tenant hosting where sub-domain is allocated to each
+  tenant, the top-level domain must be included in Public Suffix List.
+
+
+## API Key per Client
+
+Each app can support multiple clients, such as web app and mobile app. It is
+common that web app uses cookies, while mobile app uses Authorization header.
+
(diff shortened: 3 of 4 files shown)


features/202-session-management/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Security Consideration
+
+- HTTPS is required.
+- Session cookie is HTTP-only; web app and web SDK would not have access to the
+  session tokens.
+- Session cookie is set with `SameSite=lax`; generally, cross-domain requests
+  would not contain the cookie.
+- Session cookie would not be shared across custom domains.
+- App developers are responsible to prevent CSRF in their services.
```
chpapa: I haven't think it through, but if people would submit http req to Skygear's Cloud Functions or Microservices end point directly, maybe it would be great if we can have basic CRSF prevention support (e.g pre-generated CSRF token in cookies, and validate it on gateway)

We don't need it in version 1, but good to have it in spec and open issue first?

features/202-session-management/config.md
```diff
@@ -0,0 +1,51 @@
+# Session Configuration
+
+## Sample configuration
```
chpapa: I think a better design is just to have multiple API keys, and each API keys can be used to request a cookie-based session / header-based session, which will be determined by the SDK so it is almost invisible to developer users.

features/202-session-management/config.md
```diff
@@ -0,0 +1,51 @@
...
+    refresh_token_disabled: false
+    refresh_token_lifetime: 86400
+    session_idle_timeout_enabled: true
+    session_idle_timeout: 300
+```
+
+## User Configuration
+
+The value of key `clients` is a list of client configuration:
+- `name`: Name of client. Show in UI (e.g. portal)
```
chpapa: I don't think the name of client shall be determined by API key and show in UI portal.

We should allow developers to use the same API key in different clients, and each client can set the name of current session.

The name in UI should be determined by client side because the use case of a "named session" is for end-user to identify where the session come from. The name would probably be something like "Ben Cheng's iPhone", or "Ben Cheng's 2nd Blackberry at work" instead of a general name for each token.

features/202-session-management/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+# Proposed Design
+
+## Stateful Session
+
+Sessions would be stateful (i.e. persisted into database), instead of being
+stateless.
+
+Each session would be associated with attributes:
```
chpapa: I agree with these session attributes, good enough for 90% of usage. 

However I think we need a database structure that use something like an append-only log, to log down all session creation, deletion and access attempts.

i.e. So if developer want to, they can provide a feature to show not just the creation IP and last IP, but also all other IP address used (or agent)

We do not need to have the APIs for the logs access yet in version 1 or this spec, but I think we want the DB structure here too. Use append-only log/DB coz scalability is an issue. This feature could be "fast insert and slow read/query" because session access is extremely common and need to be scalable while accessing session historical data is rare in normal usage.

features/202-session-management/design.md
```diff
@@ -0,0 +1,163 @@
...
+- creation time
+- last access time
+- creation IP
+- last IP
+- creation user agent
+- last user agent
+- custom attributes
+
+Session attributes would be available in web-hook event context, and session
+management APIs.
```
carmenlau: Will microservices be able to access the context from header too?

chpapa: oh I missed this sentence when I first read the draft.

Do we need to have the session attributes in context actually? Is there some usecases that made @kiootic think they should be?

features/202-session-management/design.md
```diff
@@ -0,0 +1,163 @@
...
+If developer choose to enable session cookie:
+- Refresh token would not be issued. (See [Appendix](#session-cookie-and-refresh-token))
+- Access token would be transported in session cookie.
+- Access token would not be returned from APIs.
+- Gateway would read access token from cookie only, ignoring `Authorization`
+  header.
+
+### Security Consideration
+
+- HTTPS is required.
```
louischan-oursky: I guess this is `secure; `. It should be configurable for development purpose and by default it is **enabled**.

features/202-session-management/design.md
```diff
@@ -0,0 +1,163 @@
...
+- Session cookie is HTTP-only; web app and web SDK would not have access to the
+  session tokens.
+- Session cookie is set with `SameSite=lax`; generally, cross-domain requests
```
louischan-oursky: `"" | "lax" | "strict"`. Default to `"lax"`

features/202-session-management/api.md
```diff
@@ -0,0 +1,26 @@
+# Session Management APIs
+
+## SDK API
+```typescript
+interface Session {
+    id: string;
+    identityID: string;
+    createdAt: Date;
+    lastAccessAt: Date;
```
louischan-oursky: `lastAccessedAt`

features/202-session-management/api.md
```diff
@@ -0,0 +1,26 @@
...
+    createdByIP: string;
+    lastAccessIP: string;
```
louischan-oursky: `lastAccessedIP`

features/202-session-management/api.md
```diff
@@ -0,0 +1,26 @@
...
+    createdByUserAgent: string;
+    lastAccessUserAgent: string;
```
louischan-oursky: `lastAccessedUserAgent`

features/202-session-management/config.md
```diff
@@ -0,0 +1,51 @@
...
+    - name: Android App
+      disabled: false
+      api_key: ZZZ
+      session_transport: header
+session:
+    access_token_lifetime: 1800
+    refresh_token_disabled: false
+    refresh_token_lifetime: 86400
+    session_idle_timeout_enabled: true
+    session_idle_timeout: 300
```
louischan-oursky: A more configurable design is to inline the session configuration into client configuration. Each client configuration can have its own session configuration. Notably when `session_transport: cookie`, `refresh_token_*` are errors.

features/202-session-management/config.md
louischan-oursky: So `name` is a developer-facing name. The display name of a session should be derived from the new session properties. For example, for our iOS SDK, we should get the model of the device and store it in the session properties. We need to come up with a list of properties we can possibly get from different platform and specify them.

features/202-session-management/use-cases.md
```diff
@@ -0,0 +1,36 @@
+# Use Cases
+
+This document describes non-obvious solution to use cases.
+
+## Revoke all sessions, including current session
+Developer can revoke all other sessions, then revoke current session.
+
+```typescript
+await revokeOtherSessions();
+await revokeSession(auth.currentSessionID);
```
louischan-oursky: I prefer we disallow to revoke current session. Instead, the developer should `logout`.

features/202-session-management/design.md
kiootic: I think supporting CSRF prevention would be out of scope for this spec. We can open an issue for it for now.

features/202-session-management/design.md
kiootic: I don't think so, it would bloat the request quite a lot. Quick testing shows that including session attributes as plain JSON object would make request size +70% (+20% gzipped). I think I'll include a get & update session attribute API instead and set just the session ID in header.

features/202-session-management/config.md
kiootic: First half of your sentence is same as proposed design: developer can configure properties (e.g. session token transport) for each client (i.e. API key). I'm not sure what the second half (determined by the SDK) means.

features/202-session-management/config.md
kiootic: As mentioned in design doc, 'Client' means client application, e.g. web app and mobile app. The name would be just an internal name shown in **developer** portal.

If developer want to name a session, they could set it as custom attributes of session.

features/202-session-management/design.md
kiootic: I think the feature you want is access logs, and it is out of scope for this feature.

features/202-session-management/design.md
chpapa: Depends on if this spec is about the session management API or session itself, if it is the session itself (it seems yes, the spec got access token and refresh token together), I think CSRF is within the spec as it is closely related.

But if that's too much works, please open a feature issue for it, as we are constrained by time anyway.

features/202-session-management/config.md
kiootic: > example, for our iOS SDK, we should get the model of the device and store it in the session properties.
This would be the user agent of user.

features/202-session-management/use-cases.md
kiootic: Revoking a session and logging out can have different meaning, and I see no technical reason to disallow it.

features/202-session-management/config.md
chpapa: I meant, each API keys can be used for requesting a cookie-based session OR a header-based session, not pre-determined by the user configuration.

The second half meant, either requesting a cookie or header-based session don't have to be explicitly set by developers (at least for common usages), as the SDK will request the right kind of session for its usage (Web use Cookies, Mobile use header)

features/202-session-management/config.md
louischan-oursky: > This would be the user agent of user.

What value do we set it to? Can we specify this?

chpapa: I don't know what is the usage of having a Client name in Developer Portal to name a session, why it exists at all?

It would make more senses if we rename `clients` to `api_keys`, aka, something like this:

```
api_keys:
    - name: Default API Key
       disabled: false
       key: XXX
    - name: Employee App API Key
       disabled: false
       key: XXX
```

(However, I feel like it is out of scope and very confusing, we're talking about sessions management here right? Why bother we support multiple API keys suddenly?)

For Client's session name, we need to support session name to display in Developer Portal and when a user needs to see it in their applications (say they have a page to list all sessions), both names need to be identical. It shouldn't be a custom attributes. That's at least useful for things like doing customer technical support, or regular security audit.

Regarding @louischan-oursky 's suggestion, that make sense, but that should just be a feature in Client SDK to show the "default session name", each session name still need to be changable by clients side.

features/202-session-management/design.md
chpapa: Yes and access log was considered the feature in session management, did you look at the wireframe of the ultimate portal we want to build?

features/202-session-management/design.md
chpapa: So we shouldn't have it in web-hook event context too?

features/202-session-management/config.md
chpapa: Make sense.

But I prefer to have each API key able to do both cookies and header based session.

features/202-session-management/use-cases.md
chpapa: Agree, revoke and logout is two things.

Do we need a short cut for revokeAllSessions?

features/202-session-management/config.md
kiootic: I've discussed this with the team previously, and (please correct me if I'm wrong @louischan-oursky @carmenlau) the conclusion is that we don't fully understanding the security implication of this yet. There are some issues raised:
- Behavior when session token present in both header and cookie
- Session portability between front-ends
- Refresh token handling for cookies

For me, it seems dangerous. Using cookie would require awareness of CSRF prevention. Using header would expose the session token to client-side JS. Developers should determine which transport is acceptable to them.

features/202-session-management/config.md
kiootic: Let's clarify the scope of 'session management'. My understanding is that it includes how session works and how to manipulate sessions.

I don't see much app supporting naming individual sessions. Most UI display attributes derived from user agent & IP, such as device type, app version, IP address, country, etc. Do you have a use case requiring native support of session names?

features/202-session-management/design.md
kiootic: I didn't look very clearly last time you showed it, do you have a link for the portal design? In my opinion, access log would be implemented as part of audit log feature.

Requiring DB update every request is not that bad, depending on what DB is used. Also, every request need to hit the DB anyway. This would be an implementation detail, but let's say Redis is used to store sessions.

features/202-session-management/design.md
kiootic: Web-hook events are generated far less frequently than every request. Also, developer need to opt-in to which events to generate.

features/202-session-management/use-cases.md
kiootic: Not much application do revoke all sessions including self, most are either revoke individually or all other sessions. So I think there's no need to support it natively.

features/202-session-management/design.md
kiootic: opened #351

features/202-session-management/use-cases.md
louischan-oursky: > Revoking a session and logging out can have different meaning, and I see no technical reason to disallow it.

I agree they are two things. What I want to emphasize is when the developer would want to `revokeSession(auth.currentSessionID)` instead of `logout()`. Even if revoking the current session is technically possible, it does not mean it is sensible.

> Do we need a short cut for revokeAllSessions?

That would log the user out. I think the user case is very rare.

features/202-session-management/config.md
chpapa: Please figure it out. I think requires users to pick api keys / understand all these before they start using skygear, usability-wise is unacceptable. I have yet to see any similar platform requires it when the user start using the platform for the first time.

The requirement is we only need to provide ONE API Key for users to use on both Web/Mobile platform, ideally we provide a sensible default for which transport to use on different platform.

I think we can disallow session portability between cookies/header, no refresh token for cookies, and always have header override cookies if both present?

If a strong argument is presented that it is unavoidable, let's have another issues to handle multiple api key and transports support, this is too confusing to mix up multiple api keys feature with session feature. Bottom line is we don't present two api keys by default to users

features/202-session-management/config.md
chpapa: I have used multiple apps support naming individual sessions. I can try to list them out with research but maybe you can do the research instead? (E.g. iClouds allow naming of each devices, which effectively is used as session name)

features/202-session-management/design.md
chpapa: It is in basecamp https://3.basecamp.com/3096882/buckets/11477446/uploads/1962515766

We are using postgresql now I think? Yea sorry to go into a bit implementation here but I just wish we don't duplicate effort in architecting and implementation things, so I thought we can either do it now or leave those "last_xxx" to implement with log altogether

features/202-session-management/design.md
chpapa: But is there a frequent use case webhook event needs session data to justify the cost involved?

features/202-session-management/config.md
kiootic: Most auth platforms (e.g. Auth0) require developer to setup App and Clients at the beginning, and different Client ID/Secret (i.e. API Key) is used for different front-ends.

We had concluded that session portability should not be provided and refresh token for cookie is not useful, as mentioned in design doc. However, there is no clear choice on priority of header & cookie. I could look into it and see if there're some significant risk on accepting both transport.

There would not be 2 API keys by default. Instead, developers are required to setup the clients using presets (web/mobile) or customize settings when creating app.

features/202-session-management/design.md
kiootic: Yes, but it is intended to switch to use Redis for sessions.

features/202-session-management/config.md
kiootic: Naming device is indeed a more common feature. However session are much more transient than device, so I don't see usefulness in naming sessions.

I'd expect device name to be more persistent than session name. For example, Device name should remains the same when I re-install the app. It may require using native API from mobile devices, and tracking across installation would be likely against OS policy.

Developer can implement session name if required using custom attributes. Do you have concern on using custom attributes for session name?

features/202-session-management/design.md
kiootic: After further consideration, web-hook handler would rather want just some of session attributes, e.g. check user agent/IP on sign up. I'll remove the session attributes (except session ID) from web-hook event context, and developer can get the attributes using API if required.

features/202-session-management/use-cases.md
kiootic: At the moment, I can't come up with a use case involving revoking current session. We could disallow revoking current session for now if it does not make sense.

features/202-session-management/design.md
kiootic: Implementing access log later on should not be too difficult/requiring a rewrite, considering Redis has stream type now.

features/202-session-management/config.md
chpapa: As explained above, because we will need to display the session name in Developer Portal, I don't think it is suitable to use custom attributes for that. It is common enough to be built in.

I haven't seen anyone implemented a "persistent device name" in session management feature actually. E.g. no matter it is Google or iCloud, I can freely rename my device name whatever, and I will see the device name in my session lists. In fact I can have multiple sessions with one devices (say an Android devices with multiple apps that connected with the same services) and have different names.

I don't think we have the concept of devices here yet? and I assume practically we don't really track "Device name", but "session name" only.

Device tracking is usually more related with things like ad-tracking etc, which have privacy concern and not really related with this scope. The device name mentioned above is just to make a sensible default for session name, not for persistent tracking purpose.

features/202-session-management/config.md
chpapa: However, most Serverless platform also has a default API key and endpoint that can be used in "getting started situation".

Thanks for looking into the risk of accepting both transport -- if it indeed have problem, what about these changes?

1. We can cover multiple API keys in this spec, however, we adjust our Web JS SDK that it is smart enough to know whether the API key should be token based or cookies based (any idea how to do it?); So that by default we can still simply give 1 API key and users can use it in whatever clients.
2. Some settings as mentioned (forgot which comment, seems disappeared? Or maybe just I was on mobile) should be per-key setting like expiration

features/202-session-management/design.md
chpapa: Hmmm, okay, but let's try to avoid touching PostgreSQL DB for each request? I know currently it is but it was accepted as an "MVP implementation trade-off" and will need to be fixed some date -- let's not purposefully add another need to be fixed things in future.

features/202-session-management/design.md
chpapa: noted, actually if you think there are good reasons I'm okay if we just inject the whole session in it... just want to make sure it is considered carefully (coz it was not explained)

features/202-session-management/use-cases.md
chpapa: Good.

features/202-session-management/config.md
kiootic: If developer use Skygear as a Serverless platform, they can just deploy the services and get started; API key is not required. However, if they start using Skygear as a Auth platform, it is reasonable to expect developer to create API keys for each consuming client.

I've discussed this with the team:

- developer must choose which transport to use: they need to aware of the risks associated with each transport (e.g. CSRF for cookie, token accessible to scripts for header).
- single transport must be used: accepting/sending token in cookie may lead to security problem if services are not CSRF-aware.
- transport to use should be configured at server-side: client should not be able to request cookie if server would not accept it.

Therefore, single session token transport should be configure for each client at server-side. If we would like to simplify it, we can drop multiple API key support for now and provide a reasonable default transport (header).

features/202-session-management/config.md
kiootic: Personally I do not found session name a feature common enough - I have not yet uses app that allows assigning name to individual sessions. For developer portal, I expect it to show a full list of custom attributes for each session, so putting session name there if needed should be usable.

For iCloud, it uses device name, which is persistent across session, and in iOS this is available as `UIDevice.name`. However, there seems no consistent way to get device name on Android, and impossible on web platform. Since we cannot provide a sensible default for session name, I don't think we should support this natively.

features/202-session-management/config.md
```diff
@@ -33,19 +32,26 @@ The value of key `clients` is a list of client configuration:
...
                             default to max(`access_token_lifetime`, 86400).
                             Must greater than or equal to `access_token_lifetime`.
 - `session_idle_timeout_enabled`: Indicate whether session idle timeout is
                                   enabled, default to `false`.
 - `session_idle_timeout`: The session idle timeout in seconds,
                           default to min(`access_token_lifetime`, 300).
                           Must less than or equal to `access_token_lifetime`.
+- `same_site`: (valid for `cookie` transport only)
+               The `SameSite` property of cookie. Can be `lax`, `strict`, or
+               empty string. Default to `lax`.
```
louischan-oursky: Due to golang zero value convention, we cannot use empty string to mean turning this off. Let 's introduce `disabled`

features/202-session-management/config.md
chpapa: > If developer use Skygear as a Serverless platform, they can just deploy the services and get started; API key is not required.

Don't get it, if they use the Client SDK, they will need an API key right? Did I missed anything???

Ic, I think we can support multiple API keys, but let's make sure on the UX and getting start experienced it is streamlined.

As I mentioned in last message, if we go this route, can we have our client side SDK able to know weather the API key is cookie-based or token-based without extra configuration? E.g. in JS SDK, if a user copy/paste an API key into it, it will automatically use the right way to get the session.

kiootic: If the developer just want use Skygear as a hosting platform, they can choose to ignore auth gear and client SDK and just call service end-point directly. This does not require API key.

The client SDK should able to infer server configured session transport through server response, so developer just need to set the API key.

features/202-session-management/config.md
chpapa: Despite your personal preference, this feature is common enough. However I think this should be discussed in another part of the spec, will create another comment there. Thanks.

features/202-session-management/api.md
```diff
@@ -2,15 +2,28 @@
...
     id: string;
     identityID: string;
     createdAt: Date;
     lastAccessedAt: Date;
     createdByIP: string;
     lastAccessedIP: string;
-    createdByUserAgent: string;
-    lastAccessedUserAgent: string;
+    createdByUserAgent: UserAgent;
+    lastAccessedUserAgent: UserAgent;
```
chpapa: is it "normal" for a session to be used by different UserAgent? I know it is technically possible, but if that happens should that be treated as an "exception" (so maybe just log it down?) or it is "normal" thing that we will have `lastAccessedUserAgent`?

features/202-session-management/api.md
```diff
@@ -34,3 +48,19 @@ class AuthContainer {
     get currentSessionID(): string | null;
 }
 ```
+
+## SDK User Agent
```
chpapa: Give the example of when it is from Browser?

features/202-session-management/api.md
```diff
@@ -34,3 +48,19 @@ class AuthContainer {
...
+
+In general, for native SDKs, the user agent would have following format:
+```
+[App ID]/[App Version]([Device]; [OS]; [Device ID]) [SDK Library Name]/[SDK version]
+```
+
+Examples:
+- `io.skygear.test.ios/1.0.1 (iPhone11,8; iOS 12.0; 46FF79AA-4E2E-475F-B58F-E7E37BAB771C) SKYKit/2.0.1`
+- `io.skygear.test.android/1.3.0 (Samsung GT-S5830L; Android 9.0; 32DDE92083E1AA54) io.skygear.skygear/2.2.0`
```
chpapa: Is it really needed? Only browser have a "raw" agent string right? For other SDKs maybe they can just pass structured information to the server to fill in each data without a "raw" one?

features/202-session-management/api.md
```diff
@@ -0,0 +1,66 @@
+# Session Management APIs
+
+## SDK API
+```typescript
+interface UserAgent {
+    raw: string;
```
chpapa: only web have a "raw user agent"?

features/202-session-management/api.md
```diff
@@ -0,0 +1,66 @@
...
+    name: string;
+    version: string;
+    os: string;
+    osVersion: string;
+    app: string;
```
chpapa: would it be used for browsers when it is from web? Or there should be separate attributes for web browsers?

features/202-session-management/config.md
chpapa: > If the developer just want use Skygear as a hosting platform, they can choose to ignore auth gear and client SDK and just call service end-point directly. This does not require API key.

ic, I know, but it is not relevant to the problems we were discussing, maybe next time can try to avoid raising irrelevant comments which will confuse the main topic.

As discussed yesterday, we would use header token based session by default, and Web Client SDK can support both type of API token automatically.

features/202-session-management/config.md
```diff
@@ -0,0 +1,57 @@
+# Session Configuration
+
+## Sample configuration
+
+In user configuration:
+```yaml
+clients:
```
chpapa: Evil idea: `session_statefulness: stateless` XD

features/202-session-management/api.md
kiootic: I don't think a changed raw user agent string is exceptional case, e.g. updating OS / browser. Including a `lastAccessedUserAgent` would not have observable overhead, so I think it is acceptable.

features/202-session-management/api.md
kiootic: We should follow common convention of user agent string format (https://tools.ietf.org/html/rfc7231#section-5.5.3). I don't oppose sending an extra header containing structured data, but it would be redundant.

features/202-session-management/api.md
kiootic: I think exposing native SDK raw user agent string is acceptable?

features/202-session-management/config.md
kiootic: Stateless session is not desirable under current design:
- Each request would need to read user information from store (maybe PgSQL/Redis), so using stateless session does not avoid DB hit.
- A list of revoked session ID need to be maintained, so stateless session would not be really stateless.
- We had decided to drop client-side token introspection support for now (i.e. JWT would not be used).

features/202-session-management/api.md
chpapa: Does iOS / Android SDK's user agent string is meaningful at all? I guess not?

features/202-session-management/api.md
chpapa: Acceptable but still feel like something weird / not useful..?

kiootic: It would be more consistent: all transient attributes (e.g. timestamp, IP, user agent) would have value at creation and at last access.

features/202-session-management/config.md
chpapa: I don't agree it is not desirable... actually I think some how in future we would do it...

>     * Each request would need to read user information from store (maybe PgSQL/Redis), so using stateless session does not avoid DB hit.

Yes but IO is IO, one less IO would have shorter latency. Plus if we design to be stateless maybe we can optimize it to a point where user information are in the stateless sessions (so larger network payload in exchange for lower IO overhead and easier to scale)



>     * A list of revoked session ID need to be maintained, so stateless session would not be really stateless.

We can drop support of terminate sessions for stateless sessions



>     * We had decided to drop client-side token introspection support for now (i.e. JWT would not be used).

"for now" :)

features/202-session-management/api.md
kiootic: It is meaningful, considering the parsed information is derived from the raw user agent string. Indeed it is unlikely that developer would use this directly, but exposing this should have no problem.

features/202-session-management/api.md
chpapa: I think the use of RFC7231's user agent here is strange, it defines the user agent in HTTP protocol, but for Skygear HTTPS is just the underlying protocol we used for APIs (So technically one day we could have a different RPC protocol), and than we use HTTPS's user agent as an attributes in our session data...

features/202-session-management/config.md
kiootic: As mentioned, it is undesirable 'under current design'. Of course we can revisit it in future, but the current design does not support truly stateless session.

features/202-session-management/api.md
chpapa: @kiootic I think you resolved this comment accidentally?

features/202-session-management/api.md
chpapa: Oh I guess a difference between us is I don't think we should use the user-agent string for all information. User agent string is just ONE source of information about the sessions' devices, a structured information could be passed from client SDK if it have additional information....

features/202-session-management/api.md
kiootic: Surely if we use other protocols (grpc?), we could send this information as a structured data. Since we just support HTTP at the moment, I think it's better to conform to the common convention of HTTP for now.

features/202-session-management/api.md
kiootic: I had updated the API doc to unify this: the name would mean the browser name or app ID, and version would mean browser version or app version.

features/202-session-management/api.md
chpapa: fair enough.

features/202-session-management/config.md
chpapa: So I think we can create a feature issue about it? And mention it is likely we can add a `session_statefulness` option in this configuration?

chpapa: BTW, @kiootic when I said an "evil idea", obviously I don't expect we would have it in current design.

features/202-session-management/api.md
chpapa: However for now if you "overload" HTTP's user agent string of the network call between iOS / Android Client to the server, with a custom-made user agent string we made for user agent, it might leads to, say what if, for debug or statistic or whatever purpose, some developers think the raw user agent string our iOS's SDK HTTP library advertise itself as is useful? It is not available anymore as we already "overload" it with a user agent string of our own.

I don't think we are "conform to the common convention of HTTP for now", it is more like "a higher layer protocol overload a lower layer protocol's attributes for a special purpose"

features/202-session-management/api.md
kiootic: I had discussed with the team on how much information should be included in session:
- To comply with GDPR, we should collect the minimum amount of information by default: preferably just the user agent & IP, since it can be justified for security purpose.
- I consider device name/ID is a personal identifier (e.g. John's iPhone), therefore we should not collect this by default.
- I'm not sure what you mean by 'additional information', but personally I think this is in grey area of user tracking and analytics.

chpapa: * Please look at the other conversation thread on what I meant when I say I don't think we should overload the user agent string. 
* IP address is personal identifier too, why IP address but not device name/ID? (However I do agree we can have a feature switch for device name/ID, but I guess it is definitely helpful for users to identify strange session quickly)

features/202-session-management/api.md
kiootic: For your suggested case (`raw user agent string our iOS's SDK HTTP library advertise itself as is useful`), developer should not do this, since the HTTP library SDK used is not a public API and is subject to change.

If we used something like JSON for user agent, it would be indeed 'overloading a lower layer protocol's attributes'. The server is already assuming standard HTTP semantics, so I don't think it is our intention to do this?

chpapa: > For your suggested case (`raw user agent string our iOS's SDK HTTP library advertise itself as is useful`), developer should not do this, since the HTTP library SDK used is not a public API and is subject to change.

ya so I was saying maybe debugging or statistical purpose, or even the developer is writing its own clients (say Flutter client?) to connect with Skygear API



> If we used something like JSON for user agent, it would be indeed 'overloading a lower layer protocol's attributes'.

I have never said you should use a JSON for the HTTP user agent field. What about a custom header? 

BTW, I found that it is very confusing to call it user agent here, there are two user agent meaning: User Agent as in HTTP, or User Agent as in Sessions, maybe just make things clearer we should not call it User Agent in our sessions? Let's just use another name to avoid this confusion in future discussion / when people look at user agent in Skygear and thought it is 100% identical to the user agent in HTTP.

What about something like this:
```
interface Device {
    name: string; // browser name or app ID
    version: string; // browser version or app version
    os: string;
    osVersion: string;
    deviceType: string;
    http_user_agent: string;
}
```

features/202-session-management/api.md
kiootic: As mentioned, user agent & IP can be justified for security purpose, and most server is already logging this information. However, device name contains potentially sensitive information (e.g. 'John's iPhone' would expose real name of user), and device ID is mostly useless for security purpose (cannot co-relate sessions on same device practically). It is much harder to justify the use of device name/ID.

chpapa: The justification is it is definitely helpful for users to identify strange session quickly

features/202-session-management/api.md
kiootic: I don't see why developer would use user agent for debugging purpose? For statistical purpose, the proposed user agent string can be used. If developers would want to write SDK for other platform, they should confirm to the suggested user agent format so the server could parse meaningful information.

I don't agree there are two meanings of user agent: the user agent is the browser for web platform, and the app for native platform. User agent just means the application which the user used to interact with the server. The proposed user agent string formation reflect this clearly and conform to HTTP semantics.

Naming the user agent object as 'Device' would be misleading: it is mostly impossible to identify a device on modern platform.

features/202-session-management/api.md
```diff
@@ -0,0 +1,100 @@
...
+    identityID: string;
+    createdAt: Date;
+    lastAccessedAt: Date;
+    createdByIP: string;
+    lastAccessedIP: string;
+    createdByUserAgent: UserAgent;
+    lastAccessedUserAgent: UserAgent;
+    name: string;
+    data: object;
+}
```
chpapa: Consider the most common use case of rendering a Session to User's interface, usually it make more sense to display Location instead of IP addresses.

So our session data should include a list of Locations that currently used by the sessions / a few last locations used historically? 

So I think we should add an attributed call something like `lastAccessedLocation` in session? And for now we get the location from IP address.

And we should remember (open an issue?) to add related APIs for people to queries a list of locations when we implement the logging of session data feature (Which feature issue it is?)

features/202-session-management/api.md
kiootic: > As mentioned the justification is it is definitely helpful for users to identify strange session quickly

If it is used to identify potential account intrusion automatically, or displayed in an end-user accessible UI, then likely it can be counted as 'legitimate purpose' under GDPR. However, if we do not have such automatic mechanism, and the UI is not implemented by developers, I think it would be a violation. (correct me if I'm wrong)

features/202-session-management/api.md
chpapa: > I don't see why developer would use user agent for debugging purpose?

I don't know, maybe because they want to know which version of SDK is being used and sadly the only way they can identify it is different SDK's give different user agent string?

Again, what are the benefits of overloading HTTP user agent with our client's provided user agent information? 

> User agent just means the application which the user used to interact with the server.

Let's just use another name other than user agent, because it is confusing when most people use the term user agent, a developer would immediately think of HTTP user agent, but it is not in our case, we have mixed it up here already.

> Naming the user agent object as 'Device' would be misleading: it is mostly impossible to identify a device on modern platform.

This is just device information, calling it `Device` doesn't mean we must be able to uniquely identify a device. When you call it `User Agent` it also doesn't meant we can uniquely identify a user agent.

features/202-session-management/api.md
kiootic: I'd prefer not to include it for now, since there are some points to consider:
- Format of location (just country / more precise location)
- Geo IP database customization
- Localization
- Privacy & compliance

features/202-session-management/api.md
```diff
@@ -0,0 +1,100 @@
...
+// list all sessions of current user
+function listSessions(): Promise<Session[]>;
+
+// get specific session information for current user
+// for master-key, can use for any user
+function getSession(sessionID: string): Promise<Session>;
+
+// update specific session name for current user
+// for master-key, can use for any user and also update custom attributes
+function updateSession(sessionID: string, name: string | undefined, data?: object): Promise<void>;
```
chpapa: Second thought -- for use case that need to display a custom name in the portal, should it be a session name or device name???

Let's say the case we discussed yesterday, a Skygear APP is to developed with custom Android Devices each name with an ID. The name should be stored in Client Side, and supplied as Devices' information right? hmmm

features/202-session-management/api.md
chpapa: Strictly speaking for GDPR, it doesn't matter if it is "legitimate" enough or not in these kinds of cases, what matter is the app need to explain and let user opt-in/opt-out if they want to do it or not.

But how granular the opt-in or out need to have yet to be defined, need more court case.

So technically to support GDPR, we need a feature switch for developers to turn this feature on/off for each user. Which we can create an issue and handle later if we want to stirctly follows and respect GDPR.

features/202-session-management/api.md
kiootic: > I don't know, maybe because they want to know which version of SDK is being used and sadly the only way they can identify it is different SDK's give different user agent string?

The proposed user agent includes the SDK name and version.

> This overloading have so much implication, as mentioned above, what if we have additional information that want to pass in the session's user agent/device attributes? In mobile SDK you can put it in, what about browser?

The user agent would contains strictly the information about the user agent. For browser, considering this: if the user spoof user agent using browser extension, do we still sniff the correct information in JS and send it to server? I would suggest using only the information available in user agent header for browser.

> Let's just use another name other than user agent, because it is confusing when most people use the term user agent, a developer would immediately think of HTTP user agent, but it is not in our case, we have mixed it up here already.

As mentioned, our case is indeed the HTTP user agent.

> This is just device information, calling it Device doesn't mean we must be able to uniquely identify a device. When you call it User Agent it also doesn't meant we can uniquely identify a user agent.

Fair enough. Still some fields would need to be renamed. (e.g. name -> user agent name, version -> user agent version)

features/202-session-management/api.md
chpapa: But if we don't have it now, we almost defeated the MVP purpose of this feature issue.... :( I won't consider a session list that don't show sessions country good enough to let me check if I have unwanted session....

features/202-session-management/api.md
kiootic: The name displayed in portal would be the session name.

In that case, the developer should mention that the device name would be collected in privacy policy. Then, upon session creation/startup, use `updateSession` API to update the name of session to desired value.

features/202-session-management/api.md
chpapa: > The user agent would contains strictly the information about the user agent. For browser, considering this: if the user spoof user agent using browser extension, do we still sniff the correct information in JS and send it to server? I would suggest using only the information available in user agent header for browser.

it still doesn't cater for if we need to pass data other than those available in user agent.. say on the other thread we might have misled and need a custom name from each device for session data. Why we want to stop ourselves and make it really troubles to do it in future?

> As mentioned, our case is indeed the HTTP user agent.

As mentioned it is not -_- that is two different concept. I don't think we want to limit Skygear now to ONLY support HTTP as our transport protocol, I don't see any benefits we tie the two concepts together.



> Fair enough. Still some fields would need to be renamed. (e.g. name -> user agent name, version -> user agent version)

I don't think so? We can still just call it name and version.

chpapa: > The user agent would contains strictly the information about the user agent. For browser, considering this: if the user spoof user agent using browser extension, do we still sniff the correct information in JS and send it to server? I would suggest using only the information available in user agent header for browser.

BTW, that's why I said http user agent should only be ONE sources of information for Skygear's UserAgent/Device information.

features/202-session-management/api.md
chpapa: So I'm asking it seems better to display the device name instead of session name in Developer Portal?

So if magically someone "stole" a session, and copy it to other devices, one mistake it they may have forgot to supply the same device name and a wrong session is spotted.

features/202-session-management/api.md
kiootic: I would consider these points are essential to including location information, and changing it after release would lead to backward incompatibility. We can open another issue to address it before official release?

For now, since we provide IP, app developer can solve and implement these points themselves, and our management portal can also implemented it ourselves.

features/202-session-management/api.md
kiootic: Device name cannot be obtained reliably, the method to get it is inconsistent on Android platform, and impossible on web platform. It is not useful for a majority of user.

features/202-session-management/api.md
kiootic: If we handle it later, would it means the first official release would not be GDPR compliant?

I'm not a GDPR specialist, but I read that there are 6 basis which processing personal data is allowed. Opt-in is the 'consent' basis, and 'legitimate interest' is another basis, which security features would fall under, so consent is not required. Personally I find it hard to justify using device name/ID as security feature.

chpapa: @kiootic legally speaking a lot of things is a grey area for GDPR now. E.g. GDPR requires "granular" opt-in/out, how granular is an issue. Google thinks it is okay to make it like a few major areas to opt-in, Apple think it is okay to show one opt-in screen, and show an icon whenever privacy data is being collected... a lot of privacy-specialist think you need a long list for people to opt-in/out, some people try to file a suit to Google....

We will only know after the court makes the ruling.

I think for us we should use our best effort to enable support of GDPR compliant software, so for this feature, I would support if we add an feature issue that let people opt-in/out of session control. (probably too difficult to do it now and not common enough)

Similarly "Legitimate" is even more undefined, but generally speaking, it is legitimate interest if it is used for the benefit for that company. In this case, it is not relevant -- we are not even processing the data for the purposes of any benefits -- it is for a vital interest of the user -- to help them identify a potentially problematic session.

features/202-session-management/api.md
chpapa: Summary of my opinions on how to do it:

1. I think we should use HTTP user-agent header as ONE of the sources of information. If we have additional information about the session's end device's information, we should use another way to send it to the server (e.g. a custom header)
   1. So for Mobile SDK, since many information won't be available in HTTP user-agent header, additional information should go to the custom header
1. In the API / data structure, we should store the HTTP user-agent as raw data, but also save the additional information or information extracted for the user-agent string.
1. Try to avoid the confusion of terms user-agent between HTTP user-agent (the HTTP client library) and Skygear's user-agent (the devices that's using the session)

features/202-session-management/api.md
chpapa: Sure. Please open another feature issue than. Thanks.

features/202-session-management/api.md
chpapa: I'm talking about a device name that's definable by users here. So instead of a session name, we have a device name that could either be filled in automatically (our client SDK use the best effort to get it, or just the device type), or a device name that can be configured at client side just for display purpose.

features/202-session-management/api.md
louischan-oursky: > I think we should use HTTP user-agent header as ONE of the sources of information...

Agree. For web, we use HTTP user-agent header. For native application, we use structured data the SDK collect from the device. If we choose to set the HTTP user-agent  in native application, then the value must conform to the HTTP RFC.

> In the API / data structure, we should store the HTTP user-agent as raw data, but also save the additional information or information extracted for the user-agent string.

Agree

louischan-oursky: However, if we need to keep lastAccessedUserAgent up-to-date, the value should be transmitted via HTTP header. Add custom header(s) to represent structured data seems weird.

In case the client is not sending conforming HTTP user-agent (e.g. non-official flutter client), many information of `UserAgent` would be missing. So all fields should be made optional.

chpapa: Architecture-wise, we can do the following to safe-guard future implementation:

1. User Agent is one of the sources, especially for web SDK most information should be from it.
2. We have a custom header for additional information, which will override with the information we extract from http user agent header. 

So different client SDK implementation can do it with the "best way" for each client.



> However, if we need to keep lastAccessedUserAgent up-to-date, the value should be transmitted via HTTP header. Add custom header(s) to represent structured data seems weird.

Why? What is the werid part? @louischan-oursky  



> In case the client is not sending conforming HTTP user-agent (e.g. non-official flutter client), many information of `UserAgent` would be missing. So all fields should be made optional.

It is optional anyway, we don't even have anyway to "requires" client connecting to the server have a User Agent or any header... (unless we want it to be that strict to reject requests without a proper header?)

louischan-oursky: > Why? What is the werid part? @louischan-oursky

Use a proprietary header while there is `user-agent`.

> We have a custom header for additional information, which will override with the information we extract from http user agent header.

Terry has proposed a format of `user-agent` of native application SDK. I guess we can have two identical headers `user-agent` and `x-skygear-user-agent` where the former can be overridden by the developer.

chpapa: > Use a proprietary header while there is `user-agent`.

@louischan-oursky I think my concern and different with you and @kiootic, as explained with the reasons above, is I don't think we should use the HTTP `user-agent` for our clients SDK to inject the information of sessions' device information. Instead, we should just treat the HTTP `user-agent` as one source of information, and inject additional information to a custom header (preferably structurally)

Let me put it into another perspective, imagine in a case where the client apps need to survive in an environment where a WAF filter all abnormal HTTP traffic, where the detection of abnormal includes some requirements on the HTTP User-Agent header. Isn't it weird for some information that should not appear in normal HTTP User-Agent string included only becoz we have magically injected some Skygear specific session device information in it?

> Terry has proposed a format of `user-agent` of native application SDK. I guess we can have two identical headers `user-agent` and `x-skygear-user-agent` where the former can be overridden by the developer.

So my suggestions is, instead we just put our session information into a custom header called something like `x-skygear-session-information` for the reason above.

The difference here should be minor, and I'm not 100% certain we will run into troubles with overriding HTTP user-agent in future.

But if we know there might be problem of `over-ridding` user-agent, why risk it? What is the downside of the alternative approach?

louischan-oursky: I am fine with the approach proposed by Ben.

features/202-session-management/api.md
kiootic: > We will only know after the court makes the ruling.

However, it would be irresponsible to provide a feature enabled by default, and say it's developer's responsibility to check the ruling and see if need to ask for consent.

> I think for us we should use our best effort to enable support of GDPR compliant software, so for this feature, I would support if we add an feature issue that let people opt-in/out of session control. (probably too difficult to do it now and not common enough)

Yes, supporting per-user opt-in/out would not fit our current schedule. Therefore, we should  only enable features that we can justify the usage of personal data by default (i.e. IP and user agent as previously mentioned).

> Similarly "Legitimate" is even more undefined, but generally speaking, it is legitimate interest if it is used for the benefit for that company. In this case, it is not relevant -- we are not even processing the data for the purposes of any benefits

It a legitimate interest of developer to use user's personal data [for IT security](https://gdpr.eu/Recital-49-Network-and-information-security-as-overriding-legitimate-interest/), as permitted by GDPR.

> it is for a vital interest of the user -- to help them identify a potentially problematic session.

In context of GDPR, vital interest is [essential to life of the user](https://gdpr.eu/Recital-46-Vital-interests-of-the-data-subject/), which obviously is not applicable in our situation.

I'll suggest reading some summary on GDPR to understand more about it (personally used resources at [ICO](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/lawful-basis-for-processing/)). I found it quite clear to understand.

features/202-session-management/config.md
kiootic: opened #353 to track.

features/202-session-management/api.md
kiootic: opened #354 to track

features/202-session-management/api.md
kiootic: > I'm talking about a device name that's definable by users here.

I don't understand; session name is already configurable by users?

> a device name that can be configured at client side just for display purpose.

So this device name is a client-side concept only, which SDK would automatically fill in as session name?

chpapa: > I don't understand; session name is already configurable by users?

Yes, but I meant it seems the concept of a user configurable name, is more appropriate to be modelled as a client side configurable name per device, instead of a server side configurable name per session.

The different should be small though, coz unless users do something tricky, the session token is not suppose to be "transfer" to another device.. Got it? and thought?

features/202-session-management/api.md
chpapa: > However, it would be irresponsible to provide a feature enabled by default, and say it's developer's responsibility to check the ruling and see if need to ask for consent.

Common law just doesn't work that way. Until there is a court ruling even lawyer can't tell you what exactly the legal requirements is -- especially in tech-related law where legislator always have no clue lol.

I agree we can provide a best practices approach, but than it also relates with cost. For example we both agree support per-user opt-in will not fit our schedule.

I'm losing you here, what are we discussing?

kiootic: > I'm losing you here, what are we discussing?

We are discussing whether including device name/ID by default would be compliant to GDPR. As we discussed offline earlier, we could make including device name opt-in, since it would be beneficial to layman user for identifying devices. However, device ID is not need for now, so it is not included in session info.

chpapa: > In context of GDPR, vital interest is [essential to life of the user](https://gdpr.eu/Recital-46-Vital-interests-of-the-data-subject/), which obviously is not applicable in our situation.

BTW, I didn't use vital interest in legal terms here. So it is beyond the point of our discussion..

chpapa: In conclusion for offline discussion with @kiootic , unless there are more logical arguments:

1. we believe device name is a good information sources to help layman users to identify compromised sessions
1. device name is not accurate and can only be gather as a best effort
1. device name feature could be turned off or disable by default and turned on.

features/202-session-management/api.md
kiootic: Summarizing my opinions:
- In response to WAF: personally speaking, I'd say unknown header is much more abnormal than unknown user agent header value.
- According to RFC 7231, a user agent header should be included to indicated the user agent. Most HTTP library has already set user agent header to identify themselves. Therefore, regardless whether we store session info in it, client SDK should fill in user agent header correctly.
- In server-side rendered web app, a custom header cannot be sent with every request. A better approach would be a JS accessible cookie with a specified name, containing a JSON formatted extra device info. This feature should require user consent. (i.e. opt-in by developer by setting client side option).

chpapa: >     * In response to WAF: personally speaking, I'd say unknown header is much more abnormal than unknown user agent header value.

Not really in the age of XMLHTTPRequest all over the places...



>     * According to RFC 7231, a user agent header should be included to indicated the user agent. Most HTTP library has already set user agent header to identify themselves. Therefore, regardless whether we store session info in it, client SDK should fill in user agent header correctly.

Exactly -- HTTP library set user agent header to **identify themselves** as a HTTP client, not as a Skygear's Client...

>     * In server-side rendered web app, a custom header cannot be sent with every request. A better approach would be a JS accessible cookie with a specified name, containing a JSON formatted extra device info. This feature should require user consent. (i.e. opt-in by developer by setting client side option).

hmmm, but then for token-based sessions, we will have a header? So we will support both header and tokens even for client information? I think it is a bit over-complicated and maybe we could just drop sending extra Client's information for server-side rendered web app, feel like a okay trade-off for complexity (because for server-side rendered app, the client must be a Web Browser which provide enough information by its HTTP user-agent header already)

kiootic: > Exactly -- HTTP library set user agent header to identify themselves as a HTTP client, not as a Skygear's Client...

It seems you're think Skygear is something like a protocol implemented on top of HTTP protocol? If at the moment, we're using HTTP as the transport for API calls, the app (and the SDK) would be the user agent, and should be identified as such. For example, when gRPC is used with HTTP2 transport, it still recommend client libraries to fill in correct user agent header.

> hmmm, but then for token-based sessions, we will have a header? So we will support both header and tokens even for client information?

No, the extra info is put in a cookie regardless session token transport: the nature of this token is unrelated to authentication. It is expected that client SDK would have a cookie jar ready, in case server is configured to use cookie as session token transport, so using cookie for extra info would not be more complex.

chpapa: I think requires Client SDK must have a cookie jar ready even for token based session (which I believe is the default) is an additional technical complexity which doesn't worth for the feature of supplying session devices information, please don't do it.

features/202-session-management/api.md
```diff
@@ -11,19 +12,29 @@ interface UserAgent {
...
-    createdByUserAgent: UserAgent;
-    lastAccessedUserAgent: UserAgent;
+    userAgent: UserAgent;
+    device: Device?;
     name: string;
     data: object;
 }
 
+interface ExtraSessionInfoConfig {
+    enabled: boolean; // indicate whether extra information is enabled, default false
```
chpapa: Is it a server-side config or client-side config? (btw, I always have this question when I read design spec recently)

I don't think the enabled is needed for server side config, so I hope it is purely an API in the SDK.

The field of `deviceName` could be empty and optional

features/202-session-management/api.md
```diff
@@ -11,19 +12,29 @@ interface UserAgent {
...
     lastAccessedIP: string;
-    createdByUserAgent: UserAgent;
-    lastAccessedUserAgent: UserAgent;
+    userAgent: UserAgent;
+    device: Device?;
     name: string;
     data: object;
 }
 
+interface ExtraSessionInfoConfig {
```
chpapa: Actually I think having an `ExtraSessionInfoConfig` is very strange. Why does DeviceName an `ExtraSessionInfo` but not other fields? Why we need such a distinguishment?

features/202-session-management/api.md
```diff
@@ -63,8 +76,14 @@ For web SDK, the user agent would be controlled by browser, for example:
...
 information may not be backward compatible.
-Some parsed user agent information may not be available, and empty string
-would be used instead.
+
+
+## SDK Extra Information Collection
+Extra information collection is disabled by default. If enabled, the collected
+information would be sent in cookie `SKYGEAR_SESS_INFO` as JSON encoded object.
+
+Extra information collection configuration would be persisted across app
```
chpapa: I think it should be a client side choice to supply what information to server? Why mention "configuration would be persisted across app restart"?

features/202-session-management/use-cases.md
```diff
@@ -40,3 +40,10 @@ async function updateDeviceID(sessionID: string, deviceID: string): Promise<void
     await skygear.auth.updateSession(sessionID, { deviceID });
 }
 ```
+
+# Session name
+To display a session name in session management UI, it is recommended to
```
chpapa: UI/UX-wise, I think in all cases we should display device model & OS version, IP address etc.

But we display device name / session name if present as a more prominent information, and downplay other information in those cases.

features/202-session-management/api.md
chpapa: Moreover, if we somehow collect even more sensitive information for each session for security reason, the opt-in / opt-out would be per information pieces. So it also doesn't make sense to have a "global switch" to enabled or disable `ExtraSessionInfoConfig`

features/202-session-management/api.md
kiootic: It is client side API. Server side config are in `config.md`.

features/202-session-management/api.md
chpapa: TBH, if I am forced to choose between accept overriding all sessions information in HTTP user-agent, or add a requirement of using cookies, I would definitely goes for reusing HTTP user-agent for the session information purpose XD

chpapa: > It seems you're think Skygear is something like a protocol implemented on top of HTTP protocol? If at the moment, we're using HTTP as the transport for API calls, the app (and the SDK) would be the user agent, and should be identified as such. For example, when gRPC is used with HTTP2 transport, it still recommend client libraries to fill in correct user agent header.

Yes, as mentioned above, we're just playing safe here -- since I don't see any real downside of not overriding user-agent and use a custom header, why not...

But anyway if you really really insist and we don't have further reasons, I'm okay to just accept it and use http user-agent, chance is 99% we won't run into any trouble anyway and it's not worth the time for such a long back-and-forth. I also hate we suddenly come up with requirements that force client SDK to have a cookie jar which is even more unacceptable.

features/202-session-management/api.md
kiootic: I use browser user agent as the baseline; mobile SDKs are expected to send information contained in browser user agent in user agent header. Extra information are put in the cookie and configure here.

features/202-session-management/api.md
kiootic: Yes it is client side choice. Developer can, for example, update the device name once and no need to set it every time app starts.

features/202-session-management/use-cases.md
kiootic: It is a recommendation on how to display a meaningful session name, developer should consider the actual UI design themselves.

features/202-session-management/design.md
chpapa: @kiootic to resolve this comment and make this design more complete, can you add the note here about:

1. The session/device management features won't be completed without an audit log of session data, so the session attributes will be stored in the log (add the feature issue number) in future and consolidated into this API. The current version of API is considered an MVP versions to only allow access to "latest session information" of each user.
1. Mention the session would be persisted in a low-latency horizontally scalable store for scalability reasons (can mentioned redis if it is decided)
1. Mention country is missing which is too important to ignore.

features/202-session-management/api.md
kiootic: We can have a granular toggle for individual information pieces, and a global 'kill switch' to disable all extra info collection (i.e. `enabled`).

features/202-session-management/api.md
```diff
@@ -0,0 +1,119 @@
...
+    raw: string;
+    name: string; // browser name or app ID
+    version: string; // browser version or app version
+    os: string;
+    osVersion: string;
+    device: string;
+}
+
+// Field value would be empty string if not available
+interface Device {
```
chpapa: If we don't foresee anything more than a name for each device, shall we just use a simple string instead to keep the SDK API simple?

features/202-session-management/api.md
```diff
@@ -0,0 +1,119 @@
...
+
+interface Session {
+    id: string;
+    identityID: string;
+    createdAt: Date;
+    lastAccessedAt: Date;
+    createdByIP: string;
+    lastAccessedIP: string;
+    userAgent: UserAgent;
+    device: Device?;
```
chpapa: device is part of UserAgent?

kiootic: I'll rename it to `deviceModel` for clarity.

features/202-session-management/api.md
chpapa: ic. Maybe say "Client is responsible for persisting the extra information configuration on its own" which is more clear? (App restart could meant the mobile app / web app or skygear app which is server side)

features/202-session-management/api.md
kiootic: If we use a simple string, it would be hard to expand it in future. Some potential information to include:
- GPS location
- device ID
- custom data

features/202-session-management/use-cases.md
chpapa: Ya of coz, but since we write a recommendation in our spec, I think we should give a sensible best practices / default?

Even with the extra information I think device model & OS version, IP address (country) etc should still be displayed.

features/202-session-management/api.md
kiootic: I'll write 'Client SDKs should persist the extra information configuration locally', is that okay for you?

features/202-session-management/use-cases.md
kiootic: I agree a session entry should contains these information. However, this recommendation concerns about session name, which displaying IP address would be a bad UX.

features/202-session-management/api.md
chpapa: 1. Please, I really hate the idea of requiring cookies just for session information even for token-based header. It adds a very big hassle for client SDK implementation just for a piece of information...
2. I think the baseline of browser user agent's information is vague and an unnatural concept to follow for developer users. What are the downside of sticking with a simple Session Data structure?
3. I think having two levels of feature switch here is too complicated. We won't have a lot of switch anyway, keep it simple? (The only one other than device name I can foresee is more accurate location e.g. GPS which obviously need to be opt-in.)

features/202-session-management/api.md
chpapa: Hmm, in that case, what is the different between user agent and device? Why not just combine both information into one struct?

chpapa: This struct looks like it fits perfectly?
```
interface Device {
    http_useragent: string;
    name: string;
    version: string
    os: string
    osVersion: string
    deviceModel: string
    deviceName: string
    gps: Coordinates
}
```
}

features/202-session-management/api.md
chpapa: Perfect.

features/202-session-management/use-cases.md
chpapa: Well we could replace IP address with country when we have it.

Actually the point is the current wordings imply we should not display device Model OS etc if user supplied session name / device name is present.)

features/202-session-management/api.md
kiootic: 1. Okay, then mobile SDKs would not support cookie as session token transport.
2. It is not expected for normal developer to understand the format of user agent. It is generated by us and consumed by us, also for intermediate consumers like analytics and firewalls.
3. Okay, I'll use individual switches for each information pieces.

features/202-session-management/api.md
```diff
@@ -0,0 +1,107 @@
...
+// list all sessions of current user
+function listSessions(): Promise<Session[]>;
+
+// get specific session information for current user
+// for master-key, can use for any user
+function getSession(sessionID: string): Promise<Session>;
+
+// update specific session name for current user
+// for master-key, can use for any user and also update custom attributes
+function updateSession(sessionID: string, patch: { name?: string; data?: JSONObject }): Promise<void>;
```
chpapa: @kiootic As discussed before, not sure if you remember -- it seems better to have "deviceName" customizable instead of "sessionName", it seems more useful for the use case, if someone need to be able to identify each devices, and prevent sessions being used on the wrong devices.

So please help to:

1. Remove `updateSession` it seems not useful (we can still keep it if we think there are use case for users to store `data`, but I don't have any in my mind yet)
2. Remove `name` in the `Session` Interface
3. Provide a client side SDK api to update `deviceName` manually.

Or maybe we should scrap the whole feature of a custom name for the initial release first.

chpapa: BTW, for custom attributes of sessions, would it be better to design it as:

1. Just remove `updateSession` API
2. Attributes in `X-Skygear-Extra-Info` header will be stored in the session custom data (we need to add a strict size limit for the extra session custom data to avoid spam, and also for now just keep the latest version's value of each key? Until we have the audit trail feature)

kiootic: - Currently, it supports both device name and session name. Device name can be set in code or discovered automatically by SDK, while Session name can be updated using API and persisted at server side.
- The main purpose of custom attributes is to store session specific trusted data (i.e. can be updated only with master key).
- The size limit of `X-Skygear-Extra-Info` header is already implemented. I consider it as an implementation details so I didn't write it into the spec.

features/202-session-management/design.md
```diff
@@ -0,0 +1,215 @@
...
+If developer choose to enable session cookie:
+- Refresh token would not be issued. (See [Appendix](#session-cookie-and-refresh-token))
+- Access token would be transported in session cookie.
+- Access token would not be returned from APIs.
+- Gateway would read access token from cookie only, ignoring `Authorization`
+  header. If a token is present in both header and cookie, it would be treated
+  as if no token is present.
+- The session cookie would be cleared when:
+    - access token will be returned when header transport is used; or
+    - token is present in both header and cookie; or
```
chpapa: Would it be better to present an error instead of fail silently by clear session cookie in this case?

kiootic: - We do not know how to present the error in a correct format: it may be JSON for API, and HTML for rendered pages.
- This case can occurred only due to configuration changes, which is considered an edge case.

features/202-session-management/api.md
chpapa: * Ic. How the device name is set in Client side SDK? Seems not covered in spec?
* What are the use cases of custom attributes? I'm thinking is it better to remove Session Name and Session data, if we don't have a use case. Avoid uncertain API interface for future troubles if we don't know what it is for.

features/202-session-management/design.md
chpapa: What about just an HTTP Error without content? Sounds like could be 400. As you said it is edge case, so a HTTP Error with a simple text seems sufficient to me.

Or it is intended to fail silently, so when a configuration changes, next request will go through because the cookies are cleared?

chpapa: (If it is the later case I think we should include the "Why" in the spec, it is an important piece of imformation why we intend to silently handle an error.

features/202-session-management/api.md
kiootic: - It is set in `auth.extraSessionInfoOptions.deviceName`.
- I was thinking about traditional session data (e.g. `$_SESSION`). I don't think there are use case that strictly requires this custom attributes, so I could remove it for now.

features/202-session-management/design.md
kiootic: Strictly speaking, it is not an error or failure: it is just treated as session not found. If session is not found (e.g. revoked), the session cookie is cleared.

chpapa: But it is not expected, right? In that case, I think we can treat it as an error.

Again, or unless it is intended to fail silently? In that case, can we write down the reason in spec?

features/202-session-management/api.md
chpapa: Noted, let's remove session data and `updateSession` until we got a real use cases (or know what we are handling lol)

Usability-wise, would it better if we change the option `collectDeviceName` into a function that try to automatically fill in `deviceName`? It would be less confusing in case someone manually set `deviceName`, when it will be overrided if `collectDeviceName` is true

features/202-session-management/api.md
```diff
@@ -21,32 +21,12 @@ interface Session {
...
     lastAccessedIP: string;
     userAgent: UserAgent;
-    name: string;
-    data: object;
 }
 
 interface ExtraSessionInfoOptions {
-    collectDeviceName: boolean; // indicate SDK should collect device name information
-    deviceName?: string; // if falsy, SDK would try to detect it if possible
+    deviceName?: string;
```
chpapa: @kiootic can you add in the spec, client side should provide helper function to auto-gen the deviceName?

PR #355, opened 2019-09-05
Support preconfigured environment in microservice

carmenlau:


features/305-microservice/microservice.md
```diff
@@ -14,16 +16,50 @@ deployments:
...
+    command: "npm start"
+```
+
+*`port`, `command`, `environment_version` are optional.*
+
+When the user run `skycli app deploy`, the following steps are taken:
+
+1. Assert `./backend/service` to be a directory.
+1. If `running_environment` is specified, `skycli` will require the controller
+server to get with `environment_version` to get the `Dockerfile` and default `port`.
```
kiootic: I don't think a default port would be useful? Developer need to set the port in code/config to setup the server, so the default would be rarely used.

features/305-microservice/microservice.md
```diff
@@ -83,3 +119,11 @@ In the future, it is possible to have configurable replicas or even auto horizon
 ## Routing to the microservice
 
 The existing routing mechanism for cloud code is enhanced to support routing to microservices.
+
+## Preconfigured environment
+
+To avoid maintain too many minor versions, only major versions will be
+provided. For example, if user set `running_environment` to `nodejs` and
+`environment_version` to `10`, docker base image `node:10` will be used. So the
```
kiootic: So for Python, we only offer two variants for 2.x and 3.x? I think the offered environment version should depends on the runtime/tooling, since not all projects follow semver.

carmenlau: Yes, you are right. For the python case, we may want to offer 2.7, 3.5, 3.6, 3.7 and 3.8? And for node maybe 8, 10 and 12. 

In the current design, every `running_environment` and `environment_version` will have 1 dockerfile. So we can decide what versions we want to support for each language.

kiootic: > For the python case, we may want to offer 2.7, 3.5, 3.6, 3.7 and 3.8? And for node maybe 8, 10 and 12.

That sounds reasonable to me. If developers want to pin a specific version (e.g. node 12.9.1), they can write their own Dockerfile.

> In the current design, every running_environment and environment_version will have 1 dockerfile. So we can decide what versions we want to support for each language.

Would we support a docker-like image spec, instead of two separate fields? e.g. `runtime: python:2.7`, `runtime: node:10`, etc.

carmenlau: > That sounds reasonable to me. If developers want to pin a specific version (e.g. node 12.9.1), they can write their own Dockerfile.

🙌🏻

> Would we support a docker-like image spec, instead of two separate fields? e.g. runtime: python:2.7, runtime: node:10, etc.

If we use docker-like image spec, will user think they can use any docker images in docker hub? Actually, I think the word `running_environment` is too long, I used it because we used it in cloud function. Given that we may want to change cloud function, maybe we don't have to use the same word. I love the word `runtime`, how about using `runtime` and `runtime_version`?

carmenlau:
Some updates:

- Removed default port to keep the flow simple
- Instead of downloading docker file based on environment, download template in `tarball` format instead, which give us more flexibility to define an environment.
- Renamed `running_environment` and `environment_version` to `runtime` and `runtime_version`.

Feel free to let me know for any comments!:)

features/305-microservice/microservice.md
```diff
@@ -14,16 +16,47 @@ deployments:
...
+    command: "npm start"
+```
+
+*`command`, `runtime_version` are optional.*
+
+When the user run `skycli app deploy`, the following steps are taken:
+
+1. Assert `./backend/service` to be a directory.
+1. If `runtime` is specified, `skycli` will request the controller with `runtime` and `runtime_version` to download the template in `tarball` format.
+1. Unzip the template and merge to the user code. Look for `<context>/.dockerignore`. If it is not present, the entire context is archived. Otherwise, use the same mechanism as `.dockerignore` to ignore files.
```
kiootic: I think the merge process need to be elaborate more. For example, what if the both the template and user code has `.dockerignore`?

features/305-microservice/microservice.md
```diff
@@ -83,3 +116,9 @@ In the future, it is possible to have configurable replicas or even auto horizon
 ## Routing to the microservice
 
 The existing routing mechanism for cloud code is enhanced to support routing to microservices.
+
+## Preconfigured environment
+
+Every `runtime` and `runtime_version` has its own template. The template should contain a `Dockerfile`. If the template requires scripts or other files, we should give special name to folders or files, to avoid conflict with the user source code files.
```
kiootic: I think we should reserve and specify a directory (e.g. `.skygear`) to store the scripts/dependencies in template.

features/305-microservice/microservice.md
louischan-oursky: For compiled language like Golang, the word runtime may not be appropriate. Travis uses the word `language`.

carmenlau: After discussion, we will use the term `template`. e.g. `template: nodejs:10`

carmenlau:
Updated! Please take a look @louischan-oursky @kiootic

features/305-microservice/microservice.md
```diff
@@ -132,3 +132,30 @@ Folder structure of template
...
+### Nodejs example
+
+Dockerfile
+
+```Dockerfile
+FROM node:12
+
+WORKDIR /usr/src/app
+
+COPY package*.json ./
```
louischan-oursky: We should support both yarn and npm

features/305-microservice/microservice.md
```diff
@@ -132,3 +132,30 @@ Folder structure of template
...
+COPY . .
+
+CMD [ "npm", "start" ]
+```
+
+.dockerignore
+
+```
+node_modules
+npm-debug.log
```
louischan-oursky: Maybe we should follow https://github.com/github/gitignore/blob/master/Node.gitignore

But it is gitignore file so it cannot be directly used as dockerignore.

carmenlau:
@louischan-oursky Updated! Thanks for comments:)

features/305-microservice/microservice.md
```diff
@@ -14,16 +16,48 @@ deployments:
...
+    context: ./backend/nodejs
+    template: nodejs:10
+    command: ["npm", "start"]
+```
+
+*`command` is optional.*
+
+When the user run `skycli app deploy`, the following steps are taken:
+
+1. Assert `./backend/service` to be a directory.
```
chpapa: `./backend/nodejs`?

features/305-microservice/microservice.md
```diff
@@ -14,16 +16,48 @@ deployments:
...
+
+
+### Microservice with preconfigured environment (e.g. nodejs)
+
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
+    path: /api/
+    port: 3000
```
chpapa: I'm a bit confused if the docker template is provided by us when the template is present, what is the point of specifying port?

features/305-microservice/microservice.md
```diff
@@ -14,16 +16,48 @@ deployments:
...
+    context: ./backend/nodejs
+    template: nodejs:10
+    command: ["npm", "start"]
```
chpapa: similarly, the template is by us... the command should be pre-defined by the template?

features/305-microservice/microservice.md
```diff
@@ -14,16 +16,48 @@ deployments:
...
+### Microservice with preconfigured environment (e.g. nodejs)
+
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
+    path: /api/
+    port: 3000
+    context: ./backend/nodejs
+    template: nodejs:10
```
chpapa: for readability, I think `template` should go after `type`.

features/305-microservice/microservice.md
carmenlau: User can start the web server at any port which is specified in their code, so we need user to provide the port number in skygear.yaml

features/305-microservice/microservice.md
carmenlau: `command` is optional. user can still override it to have more flexibility. The default is `npm start`

features/305-microservice/microservice.md
carmenlau: sure

features/305-microservice/microservice.md
chpapa: ah ic, sorry for my stupidity

PR #359, opened 2019-09-24
Support "organizations" and "common" special tenant

louischan-oursky:


features/332-sso/api.md
```diff
@@ -326,11 +326,23 @@ oauth:
...
     id: azure-ad-1
+    # When "type" is "azureadv2"
+    # The developer must specify "tenant".
+    #
+    # The value typically should be the tenant ID of the Azure AD.
+    #
+    # When the OAuth client is configured to be "Any Azure AD directory - Multitenant",
+    # then the value must be the special value "organizations".
+    #
+    # When the OAuth client is configured to be "Any Azure AD directory - Multitenant) and personal Microsoft accounts (e.g. Skype, Xbox",
```
kiootic: The parenthesis seems misplaced.

features/332-sso/api.md
```diff
@@ -326,11 +326,23 @@ oauth:
...
+    # then the value must be the special value "common".
+    #
+    # If the developer wants to disallow some AD directories or disallow some kinds of Microsoft accounts,
```
kiootic: I remembered we discussed before on putting the list of allowed tenants in configuration. Do we choose not to do it?

louischan-oursky: It is because when the account is from Microsoft personal account services likes Xbox and Skype, the tenant ID is the special value `9188040d-6c67-4c5b-b112-36a304b66dad` as stated in the documentation of Azure ID token. In case of Microsoft personal account, the developer may want to have very complex logic to determine whether the account is allowed to become a new skygear user.

Therefore, I decided to delegate all work to the developer by recommending them to use hook. To implement allowed tenants, they need to maintain a whitelist of tenant IDs and check `tid` in the ID token claims.

PR #366, opened 2019-10-21
Support microservice with pre-built docker image and update secret 

carmenlau:
- Update microservice spec to support per-built docker image
- Update environment variable and secrets in microservice spec
- Update secret with type

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -380,12 +380,6 @@ aws_access_secret       2019-01-31T15:00:00+08:00
 
 Create new secret in app
 
-### skycli secret rename
-
-Update secret name
-
-`skycli secret rename [SECRET_NAME] [NEW_SECRET_NAME]`
-
 ### skycli secret remove
 
 Remove secret in app
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -376,9 +376,17 @@ aws_access_secret       2019-01-31T15:00:00+08:00
 
 ### skycli secret create
 
-`skycli secret create [SECRET_NAME] [SECRET_VALUE]`
+`skycli secret create [SECRET_NAME] [SECRET_VALUE] --type=[SECRET_TYPE] --file==[FILE_NAME]`
 
 Create new secret in app
+
+#### Flags
+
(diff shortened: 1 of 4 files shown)


features/305-microservice/microservice.md
```diff
@@ -61,6 +82,20 @@ When the user run `skycli app deploy`, the following steps are taken:
...
+```yaml
+deployments:
+  my-awesome-service:
+    type: http-service
+    ....
+    environment:
+    - name: "DEMO_ENV"
+      value: "value of environment"
+      name: "MONGODB_URL"
+      secret: "<secret_name>"
```
chpapa: An idea, would it be better if have a shorthand when user want to use a secret name directly as an env var name too? Just to save a few keystrokes, like:

`secret2env: "<secret_name>"` or a `secret` without a `name`

is identical to

```
name: "<secret_name>"
secret: "<secret_name>"
```

louischan-oursky: `secret` without a `name` is unambiguous.

- `ENV_NAME: "plain value"`
- `ENV_NAME: "<secret_name>"`

is too ambiguous

features/270-next-skycli/commands.md
```diff
@@ -376,10 +376,18 @@ aws_access_secret       2019-01-31T15:00:00+08:00
...
 
 Create new secret in app
 
+#### Flags
+
+- `--type` Secret type, used to facilitate programmatic handling of secret data. Available types: `opaque`, `dockerconfigjson`. Default is `opaque`.
+    - `opaque`: Secret used in environment variable.
+    - `dockerconfigjson`: Docker config used as image pull secret.
+
+- `--file` Secret value from file.
```
louischan-oursky: Is `--file` specific to `--type=dockerconfigjson` or is it applicable to `--type=opaque` too, just like k8s?

carmenlau: I think file should be independent to type, both type can use file.

features/305-microservice/microservice.md
carmenlau: I think we can do `secret` without a `name`, but I will still prefer user provide the `name`  maybe in the guide so user will know they can specify the env name in `skygear.yaml`.

louischan-oursky:
@kiootic Any comments?

Issue #368, opened 2019-10-28
Review tenant config and set principles 


tensiuyan:
## Description

**Add 2-3 sentence description of your feature**

## Portal Design

**Remove this section if the feature have no Portal**

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

carmenlau:
@louischan-oursky @kiootic @chpapa For the tenant config change that we discussed in the product meeting, I drafted some points and added the example config. Please kindly take a look and let me know your thoughts. I will add this to the spec when we have consensus!

## Should use array or map to define dynamic items in user config?

- Map with identifier as key
    - Pros:
        - Clearer for referencing (e.g. configure user verification need to refs items in login id keys)
        - Prevent duplicate key by config structure
    - Cons:
        - The config looks inconsistent. For items that we don't have identifier, we use array. (e.g. In id in sso provider is optional. If it is not provided, we use type as the id. sso providers is array now).
        - Cannot preserve item order
    - Existing config that use map
        - docker compose config defines volumes and services
        - aws cloudformation defines resources
- Array
    - Pros:
        - Can represent all items (with or without identifier)
        - Can preserve order
    - Cons:
        - Need validation to prevent duplication
    - Existing config that use array
        - k8s resources spec

I prefer using array to prevent inconsistent.

## Should we have `enabled` config

We should only have the `enabled` for enabling feature. Config file should not be a place that store the config data temporarily. We should not add `enabled` or `disabled` flag to item, instead user should delete the item if they don't need it. In most of the common platform (gcp, aws.. e.g), we don't often see disable an item option in their portal. User has responsibility to keep the copy of the config, to make it simple, ui should only support adding or removing items.

## Config updates

- Remove `disabled` in `clients`
- Change all map to array. (e.g. clients, auth.login_id_keys, user_verification.login_id_keys)
- Rename `id` in sso provider to `name` as it is optional?

## Portal remarks

Portal should handle editing config that affects multiple sections. For example, remove login id key should also remove the user verification config if the id is enabled in user verification section. Maybe adding warning in ui, before deleting a login id?

## Config example

Based on the proposal, here are the config for preview.

[...]

PR #371, opened 2019-10-31
Add user config conventions

carmenlau:
connect #368

### DIFF
--- a/features/324-config/conventions.md
+++ b/features/324-config/conventions.md
@@ -1,0 +1,56 @@
+# Configuration Conventions
+
+## Background
+
+User config is the user-facing configuration that allow user to enable and provide configurations to Skygear provided gears. This document is intended to provide conventions to ensure the configuration is clear, easy to understand, flexible for new features and consistent.
+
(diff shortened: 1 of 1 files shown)


features/324-config/conventions.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Background
+
+User config is the user-facing configuration that allow user to enable and provide configurations to Skygear provided gears. This document is intended to provide conventions to ensure the configuration is clear, easy to understand, flexible for new features and consistent.
+
+## Conventions
+
+### Prefer using list of items over map
+
+To defines multiple items (e.g. login id keys, sso providers) in config, don't use map. A list of items containing the key field should be used instead.
```
louischan-oursky: Is the field name `key` unconditional for all cases? Maybe in some cases, we may want to use other terms, say `name`, `id`, etc.

carmenlau: I thought the key field is added only when necessary. Rephrase the sentence and see if it looks better.

PR #372, opened 2019-11-04
Glossary

### LINKED ISSUE #363: Define terminology
## Description

Define a list of terminology, so that documentation and UI would be consistent.

- Skygear Cluster: the Skygear hosting platform.
- Skygear Portal: the web management interface of Skygear Cluster.
- Skygear App: a collection of resources on Skygear Cluster, can omit 'Skygear' prefix if context if clear.
- Gear: officially-supported multi-tenant-aware services.
- Auth Gear
- Developer: the developer of a Skygear App.
- User: the user of a Skygear App.
- Deployment: a unit of executable code belong to an app.
- Secret: confidential key value pair belonging to an app.
- Collaborator: developer with management access to app.
- User Config: configuration to gears.
- skygear.yaml: configuration to deployment, including web-hooks.
- Scaffold: the action of initializing an empty directory using template.
- Scaffolding Template:
- API Key
- Master Key

Terms NOT to use:
- Controller/Platform: should use Skygear Cluster.
- Project: should use Skygear App.
- Skygear Auth: should use Auth Gear.

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

kiootic:
resolve #363

### DIFF
--- a/features/363-glossary/glossary.md
+++ b/features/363-glossary/glossary.md
@@ -17,6 +17,7 @@ Glossory
 - **Skygear Cluster**: the Skygear services (Gears, Gateway, and Controller), hosted on Kubernetes.
 - **Skygear Controller**: the management interface of Skygear Cluster.
 - **Skygear Portal**: the web management interface of Skygear Cluster.
+- **skycli**: the command-line management interface of Skygear Cluster.
 - **Skygear App**: a collection of microservices on Skygear Cluster, can omit 'Skygear' prefix if context if clear.
 - **Skygear Gateway**: the proxy which incoming traffic to Gears and Microservices must go through.
 - **Gear**: officially-supported multi-tenant-aware services.
--- a/features/363-glossary/glossary.md
+++ b/features/363-glossary/glossary.md
@@ -1,0 +1,48 @@
+Glossary
+========
+
+In this document, we define a list of terms to be used in communication.
+
+Objective
+---------
+
+In various documents, we will use different terms to refer to concepts in Skygear:
+- Documentation
+- Discussions
+- Emails/Announcements
+
+It would be clearer to use widely agreed terms to refer to specific concepts.
+
+Glossory
+--------
+- **Skygear Cluster**: the Skygear services (Gears, Gateway, and Controller), hosted on Kubernetes.
+- **Skygear Controller**: the management interface of Skygear Cluster. *Should not be used in public documents*.
+- **Skygear Portal**: the web management interface of Skygear Cluster.
+- **Skygear App**: a collection of microservices on Skygear Cluster, can omit 'Skygear' prefix if context if clear.
+- **Skygear Gateway**: the proxy which incoming traffic to Gears and Microservices must go through.
+- **Gear**: officially-supported multi-tenant-aware services.
+    - **Auth Gear**
+    - **Asset Gear**
+- **Deployment**: a unit of executable code belong to an app.
+    - **Microservices**
+- **Developer**: the developer of a Skygear App.
+- **User**: the user of a Skygear App.
+- **Secret**: confidential key value pair belonging to an app.
+- **Collaborator**: developer with management access to app.
+- **User Config**: configuration to gears.
+- **skygear.yaml**: configuration to deployment, including web-hooks.
+- **Scaffold**: the action of initializing an empty directory using template.
+- **Scaffolding Template**:
+- **Access Key**:
+    - **API Key**
+    - **Master Key**
+
+
+Misused Terms
+-------------
+These terms should not be used. Instead, clearer terms from above list should
+be used.
+
+**Skygear Platform**: should use Skygear Cluster.
+**Project**: should use Skygear App.
+**Skygear Auth**: should use Auth Gear.
--- a/features/363-glossary/glossary.md
+++ b/features/363-glossary/glossary.md
@@ -8,15 +8,14 @@ Objective
 
 In various documents, we will use different terms to refer to concepts in Skygear:
 - Documentation
-- Discussions
-- Emails/Announcements
+- Technical discussions
 
 It would be clearer to use widely agreed terms to refer to specific concepts.
 
 Glossory
 --------
 - **Skygear Cluster**: the Skygear services (Gears, Gateway, and Controller), hosted on Kubernetes.
-- **Skygear Controller**: the management interface of Skygear Cluster. *Should not be used in public documents*.
+- **Skygear Controller**: the management interface of Skygear Cluster.
 - **Skygear Portal**: the web management interface of Skygear Cluster.
 - **Skygear App**: a collection of microservices on Skygear Cluster, can omit 'Skygear' prefix if context if clear.
 - **Skygear Gateway**: the proxy which incoming traffic to Gears and Microservices must go through.
@@ -29,7 +28,7 @@ Glossory
 - **User**: the user of a Skygear App.
 - **Secret**: confidential key value pair belonging to an app.
 - **Collaborator**: developer with management access to app.
-- **User Config**: configuration to gears.
+- **App Config**: configuration to gears.
 - **skygear.yaml**: configuration to deployment, including web-hooks.
 - **Scaffold**: the action of initializing an empty directory using template.
 - **Scaffolding Template**:
@@ -40,9 +39,15 @@ Glossory
 
 Misused Terms
 -------------
-These terms should not be used. Instead, clearer terms from above list should
-be used.
+These terms should not be used. Instead, terms from above list should be used.
 
 **Skygear Platform**: should use Skygear Cluster.
 **Project**: should use Skygear App.
-**Skygear Auth**: should use Auth Gear.
+
+Exceptions
+----------
+
+The defined terms in this document only applies to technical documents. For
+other purposes, such as marketing materials, other terms may be used depending
+on the situation. If any doubts, the guiding principle is to be consistent on
+the term usage in related documents.


features/363-glossary/glossary.md
```diff
@@ -0,0 +1,48 @@
...
+
+In this document, we define a list of terms to be used in communication.
+
+Objective
+---------
+
+In various documents, we will use different terms to refer to concepts in Skygear:
+- Documentation
+- Discussions
+- Emails/Announcements
```
chpapa: I can forasee some conflict if the terms here are expected to be in Emails / Annoucements or other marketing communications.

e.g.:

* If Skygear Auth is wrong but Auth Gear is right -- marketing-wise we would want to make sure the term Skygear is on it...
* Skygear Platform are currently used in most marketing communications instead of Skygear Cluster

So do we want to align marketing terms or not?

kiootic: In my opinion, if the marketing materials targets at technical staff (e.g. developers), they'll most likely also read the documentation, and using multiple terms referring to the same thing would be confusing.

Your example (platform vs cluster) is exactly the motivation for this glossary when I was drafting the Getting Started guide. Is there any important difference between "Skygear Platform" & "Skygear Cluster"? When should we use one term over another? This document intends to clarify it.

This is just a very rough draft. I just opened this PR so we can start discussing about it, as discussed in product meeting, considering we'll need to get started on documentations very soon, and consistent writing style and terminology is very important.

chpapa: Noted, I will re-comment this spec assuming the terms will be used on marketing side too.

features/363-glossary/glossary.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+Misused Terms
+-------------
+These terms should not be used. Instead, clearer terms from above list should
+be used.
+
+**Skygear Platform**: should use Skygear Cluster.
+**Project**: should use Skygear App.
+**Skygear Auth**: should use Auth Gear.
```
chpapa: I think we need to consider the following: We will describe the major features of Skygear for new comers, especially on marketing / landing pages etc.

In those cases, we will probably want to use the following terms:
* Skygear Microservices
* Skygear Auth
* Skygear Assets

and maybe in future...
* Skygear Cloud Functions
* Skygear Add-ons (e.g. An integrated Helm3 to allow quick and simple deployment of PGSQL, Redis, etc)

I think it would conflict with some terms suggested here. We would probably not want to call them just "Auth Gear" in marketing. Instead, we will sometimes just refer to "Auth, Microservices" etc, or when branding is needed, we call it "Skygear Auth, Skygear Microservices".

We can either:
1. Just use those marketing branded terms in documentations and other places, adopt as standard terms, or...
2. Treat those started with "Skygear X" a categorization of features.
3. Call them "Microservices Gear, Auth Gear, Asset Gear" and "Skygear Microservices Gear, Skygear Auth Gear", etc... but seems a bit too long, and "Microservices Gear" is not exactly correct in technical definition.

Thoughts?

kiootic:
I think there are some points to discuss:
1. Do we want a separate set of terms to be used in marketing material? If so, how can we ensure the reader associates two terms with same concept? (e.g. visitor may have a hard time to learn about 'Skygear Auth' if only 'Auth Gear' appears in documentation)
2. How to cater for less technically oriented readers? (e.g. a novice developer may not understand technical concepts of 'cluster', 'gateway', 'microservices', etc.)
3. Do we want to mention the term 'gear' in documentation/marketing material? If so, what is the definition of 'gear'?

chpapa:
@kiootic 

1. I think your arguments for trying to sync between marketing materials and documentations are good. The major problems are some high-level features like Skygear Microservices/Auth/Assets which conflict with the terms to describe the exact components.
2. I think the best way to handle it is DO NOT introduce the concept of "gateway" to beginners. For Cluster I think it is kind of unavoidable? And for microservices too.. coz that's like the basic of using Skygear.
3. This is tricky. Originally technically we want to define gears as microservices in Skygear that's provided by default, multi-tenant aware, and useful for the context of app development.

1 and 3 are kind of referring to the same thing as I commented above... How do we handle the conflcit between **Skygear Auth** vs **Auth Gear**

Maybe we just accept the situation? We will need to call Skygear Auth / Microservices etc as a high level features in our marketing page, and we will explain to developers, we have auth gear that provide the auth features... (plus... conceptually, Skygear Auth should include more than auth gear, such as client SDKs or other portal / features that's related to auth)

kiootic:
1. It seems that you intended that `Skygear Auth` refers to the 'auth' feature as a whole, and `Auth Gear` refers to the specific component that is part of the implementation of this feature. If so, I suppose we **do not** mention `Auth Gear` publicly (in documentation/marketing materials), since developers should not need to know about the underlying implementation details.

2. My personal opinion is: For 'novice developers', they'll not be understand what is microservices; and if they already understand microservices, the term 'gateway' should be familiar with them. What levels of developer should our documentation targets at? Do we expect the reader to know Docker, JWT, k8s, or other concepts?

louischan-oursky:
I added Ten and Elliot as reviewers

chpapa:
@kiootic 

>     2\. My personal opinion is: For 'novice developers', they'll not be understand what is microservices; and if they already understand microservices, the term 'gateway' should be familiar with them. What levels of developer should our documentation targets at? Do we expect the reader to know Docker, JWT, k8s, or other concepts?

I think here are the level of concept we expect the user of Skygear should know when they use it...

- They should have a brief understanding of what is microservices but not the details or best practices of how to do certain things. You're right maybe Gateway is a concept/terms we can use in our documentation when needed; I think what I should have said instead of "DO NOT introduce the concept of "gateway" to beginners", should be avoid the details of how gateway works, we can tell them here is the Skygear Gateway and it did ABCDEFG for you, but they don't need to customize it most of the time.
- They do not need to know exactly how Docker works / how to use Docker to use Skygear; so a beginner could just use Skygear, put sources code in the directory, and rely on `skycli deploy` to deploy the code in Docker template we provided. Advanced users can define their own Dockerfile or deploy with pre-built Docker image.
- They do not need to know exactly how k8s works / how to use k8s to use Skygear, similar to Docker. Most users of skygear probably won't touch the underlying k8s directly, only if they know and want to they will do it themselves.

So go back to the original question of:

> How to cater for less technically oriented readers? (e.g. a novice developer may not understand technical concepts of 'cluster', 'gateway', 'microservices', etc.)

I think the answer should be:
* The most novice devleoper we can cater to, are those who have heard of "Docker" "k8s" and "microservices", but they don't know the best practices, or how exactly things work, they can still use Skygear to do develop an app, they just know Skygear help handled Docker / k8s to them... they know there is a gateway, but they probably don't understand all features Skygear provided via the gateway, or which gateway software we used.

features/363-glossary/glossary.md
```diff
@@ -0,0 +1,48 @@
...
+- **Gear**: officially-supported multi-tenant-aware services.
+    - **Auth Gear**
+    - **Asset Gear**
+- **Deployment**: a unit of executable code belong to an app.
+    - **Microservices**
+- **Developer**: the developer of a Skygear App.
+- **User**: the user of a Skygear App.
+- **Secret**: confidential key value pair belonging to an app.
+- **Collaborator**: developer with management access to app.
+- **User Config**: configuration to gears.
```
tensiuyan: rename user config

features/363-glossary/glossary.md
```diff
@@ -0,0 +1,48 @@
...
+
+Glossory
+--------
+- **Skygear Cluster**: the Skygear services (Gears, Gateway, and Controller), hosted on Kubernetes.
+- **Skygear Controller**: the management interface of Skygear Cluster. *Should not be used in public documents*.
+- **Skygear Portal**: the web management interface of Skygear Cluster.
+- **Skygear App**: a collection of microservices on Skygear Cluster, can omit 'Skygear' prefix if context if clear.
+- **Skygear Gateway**: the proxy which incoming traffic to Gears and Microservices must go through.
+- **Gear**: officially-supported multi-tenant-aware services.
+    - **Auth Gear**
```
tensiuyan: only have to take care documentation lang. Marketing terms can differ from doc terms

features/363-glossary/glossary.md
```diff
@@ -0,0 +1,53 @@
...
+    - **Asset Gear**
+- **Deployment**: a unit of executable code belong to an app.
+    - **Microservices**
+- **Developer**: the developer of a Skygear App.
+- **User**: the user of a Skygear App.
+- **Secret**: confidential key value pair belonging to an app.
+- **Collaborator**: developer with management access to app.
+- **App Config**: configuration to gears.
```
carmenlau: Currently we call the config **user config**, I opened an issue to update the config name to **app config**. Since this is another breaking change which has higher priority. https://github.com/SkygearIO/features/issues/402

carmenlau:
@chpapa @louischan-oursky @elly0t Please take a look when you have time? Mark *approve* if it looks good to you, and then I will merge it. Thanks:)

features/363-glossary/glossary.md
```diff
@@ -0,0 +1,53 @@
...
+- Documentation
+- Technical discussions
+
+It would be clearer to use widely agreed terms to refer to specific concepts.
+
+Glossory
+--------
+- **Skygear Cluster**: the Skygear services (Gears, Gateway, and Controller), hosted on Kubernetes.
+- **Skygear Controller**: the management interface of Skygear Cluster.
+- **Skygear Portal**: the web management interface of Skygear Cluster.
```
louischan-oursky: Not sure if we should also add `skycli` to this list. If we do, `: the command line interface of Skygear Cluster`

kiootic: Added skycli to glossary.

PR #373, opened 2019-11-05
Update request verification specification

### LINKED ISSUE #300: Verify requests from auth gear and gateway in CF & microservices
# Description

There are 2 parts we need verification.
1. CF need to verify if the request are from gateway, so the injected context (current auth user) can be trusted.
2. webhook (should be CF) need to verify if the request are from auth gear.

We (@carmenlau, @Steven-Chan)  had a short discussion on this, and here is the note. We may need to draft a detailed spec for this.

- the program that wrap user function (i.e. fission environment / openfaas watchdog) is responsible for verifying gateway request
- if the user function is deployed independently, the responsibility of verifying gateway request go to the developer, so we need to tell them how we sign the request
- SDK is responsible for verifying auth gear request
- secrets for verifying gateway and auth gear request are defined per app
- we introduce predefined cloud code secret (vs user defined secrets using skycli api and used in cloud code secret config)
- the secret for verifying gateway request is a predefined cloud code secret
- if cloud code config specify that a function is a auth gear web hook, the secret for that is also a predefined cloud code secret

## Scenario

**Put list of Scenario how developers will use this API**

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

kiootic:
fix #300

### DIFF
--- a/features/300-request-verification/request-verification.md
+++ b/features/300-request-verification/request-verification.md
@@ -2,94 +2,57 @@
 
 ## Motivation
 
-The gateway injects HTTP headers to tell the current authenticated user.
+Different Skygear components may communicate with other components:
+- Gateway <-> Gears/Microservices for incoming HTTP traffic.
+- Auth Gear <-> Microservices/External services for web-hook events.
 
-The auth gear sends webhooks.
+We need ways to verify the authenticity of these communications.
 
-We need ways to verify the HTTP headers and the body are from the gateway
-and the auth gear.
+To verify authenticity of communication, we need to verify two property:
+- the identity of receiver (i.e. microservices)
+- the identity of sender (i.e. gateway/auth gear)
 
-## Special Headers
+The industry standard of securing and authenticating communication is to use
+mTLS. However, for simplicity, Skygear would use a simpler design to acheive
+this objective. This document describe Skygear's design of verification.
 
-`x-skygear-headers-signature`
+## Gateway communication verification
 
-This header contains the signature for headers verification.
+Gateway communication occurs between Gateway and Microservices/Gears.
 
-`x-skygear-body-signature`
+Microservices/Gears are designed to host on Kubernetes. Therefore, for
+communication between Gateway and Microservices/Gears, we use Network Policy to
+ensure that only Gateway may interact with Microservices/Gears.
 
-This header contains the signature for body verification.
+## Web-hook communication verification
 
-## Headers signature generation
+Web-hook communication occurs between Auth Gear and Microservices/external
+services.
 
-1. Derive a byte sequence from all Skygear headers excluding the special headers.
-1. Calculate the uppercase hex digest HMAC-SHA256 of the byte sequence with a shared secret.
-1. Pass this digest as `x-skygear-headers-signature`
+To verify the identity of receiver, we require that the web-hook handler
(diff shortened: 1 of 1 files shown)


features/300-request-verification/request-verification.md
```diff
@@ -2,91 +2,44 @@
...
+- the identity of receiver (i.e. microservices)
+- the identity of sender (i.e. gateway/auth gear)
 
-## Special Headers
+The industry standard of securing and authenticating communication is to use
+mTLS. However, for simplicity, Skygear would use a simpler design to acheive
+this objective. This document describe Skygear's design of verification.
 
-`x-skygear-headers-signature`
+## Web-hook communication verification
```
chpapa: Web-hook could be outside of Skygear Cluster right?

If yes, readability-wise, this paragraph is a bit misleading.

Coz the last paragraph is talking about Communication between Gateway / Auth Gear, this paragraph is about web-hook which could be microservices or other HTTPS endpoint outside of cluster.

Should instead:
1. discuss the communication between gateway and auth gear first (move the last paragraph to here?)
2. make it clear, we have finished discussing gateway / authgear, now we discuss web-hook which is another thing?

kiootic: I intended that the heading levels should clarify the relationship between sections: there are only two second-level sections: Motivations, Web-hook communication verification, and Gateway communication verification.

I do think it is a bit confusing after a long section on web-hook, I'll move the gateway section before web-hook and clarify it further.

features/300-request-verification/request-verification.md
```diff
@@ -2,94 +2,57 @@
...
-lines = [n + ':' + v for n, v in headers]
-content = '\r\n'.join(lines)
-bytes_ = content.encode('utf-8')
-
-
-assert bytes_ == b'x-skygear-auth-disabled:false\r\nx-skygear-auth-userid:a\r\nx-skygear-auth-verified:true'
+The shared secret is set in user configuration per web-hook handler. Developer
+must to set them in configuration to use web-hook feature.
 
+To access the shared secret in microservices, it is recommended to use the
```
louischan-oursky: This brings up a problem. Currently we have no way to expose values (e.g. master key and this shared secret) in tenant config to microservice via environment variable.

kiootic: I intended that the developer would create a secret with the shared secret value, and sync it manually for now.

features/300-request-verification/request-verification.md
chpapa: > I intended that the heading levels should clarify the relationship between sections: there are only two second-level sections: Motivations, Web-hook communication verification, and Gateway communication verification.

Heading levels is not the problem here. Problem is usually people assume the first paragraph would be something like Introduction, so originally it didn't cover Webhook could be external services, it only mentions the microservices, gateway and auth gear, it could be quite misleading. 

The new force pushed one is better, thx.

features/300-request-verification/request-verification.md
chpapa: So shall we create a feature issue for this?

kiootic: Opened #379 for tracking this feature.

PR #381, opened 2019-11-19
Improve skycli for CI

carmenlau:


features/270-next-skycli/commands.md
```diff
@@ -35,6 +35,13 @@ and password interactively.
 #### Flags
 
 - `--email=EMAIL` Login as cluster user, ask for password interactively.
+- `--password-stdin` Read password from stdin.
+
+#### Example
+
+```
+$ cat ~/password.txt | skycli auth login --email user@example.com --password-stdin
```
louischan-oursky: Add one more example if the password is stored in environment variable.

```sh
$ printf "$MY_PASSWORD" | skycli auth login --email user@example.com --password-stdin
# It is very important NOT to use echo because it adds newline at the end of the output.
# echo -n is also not portable.
```

carmenlau: For better ux, skycli will trim the last newline from the input like docker cli

carmenlau: Added example:)

Issue #339, opened 2019-07-04
Improve uniqueness checking for LoginID


chpapa:
# Description

Improve uniqueness checking for LoginID with common security issues:

- [x] Default case-insensitive for email 
- [x] Default case-insensitive for loginID (could be disabled)
- [x] Normalize non-ASCII username using NFKC as [recommended here](http://www.unicode.org/reports/tr36/#Recommendations_General)
- [x] Avoid confusing homoglyphs username based on some libraries like [confusable_homoglyphs](https://confusable-homoglyphs.readthedocs.io/en/latest/index.html)
- [x] Options to disallow non-ASCII username
- [x] Normalization of email address / phone number
- [x] Options to ignore `.` and  email address after `+` for uniquness checking
- [x] Disallow loginID that is confusing ([Djagon List](https://github.com/ubernostrum/django-registration/blob/1d7d0f01a24b916977016c1d66823a5e4a33f2a0/registration/validators.py))

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

chpapa:
@kiootic If something listed above are easy, feel free to add it first :) cc: @louischan-oursky

louischan-oursky:
I discovered a related RFC

- [PRECIS Framework: Preparation, Enforcement, and Comparison of
           Internationalized Strings in Application Protocols](https://tools.ietf.org/html/rfc8264)
- [Preparation, Enforcement, and Comparison of Internationalized Strings
                  Representing Usernames and Passwords](https://tools.ietf.org/html/rfc8265)

carmenlau:
Would like to bring this up, currently signup user with same email in different cases will create multiple users.

Maybe we need to implement the point 1 first (Default case-insensitive for loginID). If the cases in display is not that important, maybe we can just store the lower case of login id when case-insensitive is enabled.

carmenlau:
Had a short discussion with @kiootic. Since email should be case insensitive by nature, rather than using a case-insensitive setting for loginID, that would be better to do the normalization by login id type.

Email should be case-insensitive, we will store the normalized value separately. All the user facing login id data will be the original input data (e.g. login id in identify when calling `/me` api). Normalized value will be used only for login or user query.

Let me know if it looks good for you. If so, we will do the case-insensitive version for email first. @louischan-oursky @chpapa

chpapa:
[...]

PR #386, opened 2019-12-03
Define login id types validation and normalization

carmenlau:
connect #339

### DIFF
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -51,7 +51,9 @@ user_config:
 
 #### Validation
 
-- Disallow username with reserved keywords in [Django list](https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127) (Configurable, default ON)
+- Disallow username with reserved keywords listed below (Configurable, default ON)
+  - https://github.com/marteinn/The-Big-Username-Blacklist/blob/master/list_raw.txt
+  - https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127
 - Disallow username with user defined list (Configurable, default empty list)
 - Disallow non-ASCII username (Configurable, default OFF)
 - Disallow username with confusing homoglyphs (Compulsory)
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -1,0 +1,94 @@
+# Login ID Type
+
+## Background
+
+Skygear Auth supports login with login id and password. This document defines the types of login id.
+
+## Validation and Normalization
+
+Validation and normalization are defined base on login id type. When user signup with login ids, Skygear Auth will validate the login ids and reject the signup if it won't pass. After passing the validation, Skygear Auth will store the normalized login id for login.
+
+When user login with login id, login id key is optional. If login id key is provided, Skygear Auth will normalize the input by type, query the principal and perform the login. If login id key is not provided, Skygear Auth will walk through all login id keys, normalize the input and query the principal. If multiple principals are found, the login process will be stopped and error will be returned.
+
+## Types
+
+This section defines the login id types and its validation and normalization rules.
+
+- [Email](#Email)
+- [Username](#Username)
+- [Phone](#Phone)
+- [Raw](#Raw)
+
+### Email
+
+#### Validation
+
+- [RFC 5322 address](https://tools.ietf.org/html/rfc5322#section-3.4.1) (Compulsory)
+
+#### Normalization
+
+- Change domain part to lower case (Compulsory)
+- Change local part to lower case (Configurable, default ON)
+- Remove words after `+` sign in the local part (Configurable, default OFF)
+- Remove all `.` sign in the local part (Configurable, default OFF)
+
+#### Configuration
+```yaml
+user_config:
+  auth:
+    login_id:
+      email:
+        case_sensitive: false
+        ignore_local_part_after_plus_sign: false
+        ignore_dot: false
+    login_id_keys:
+      - type: email
+        key: email
+```
+
+### Username
+
+#### Validation
+
+- Disallow username with reserved keywords in [Django list](https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127) (Configurable, default ON)
+- Disallow username with user defined list (Configurable, default empty list)
+- Disallow non-ASCII username (Configurable, default OFF)
+- Disallow username with confusing homoglyphs (Configurable, default ON)
+
+#### Normalization
+
+- Case insensitive (Configurable, default ON)
+- Perform NFKC (Configurable, default ON)
+
+#### Configuration
+```yaml
+user_config:
+  auth:
+    login_id:
+      username:
+        block_reserved_keywords: true
+        excluded_keywords:
+          - skygear
+          - skygeario
+        non_ascii: false
+        block_homoglyphs: true
+        case_sensitive: false
+        nfkc: true
+    login_id_keys:
+      - type: username
+        key: username
+```
+
+### Phone
+
+#### Validation
+
+- Ensure phone number in E.164 format (Compulsory)
+
+#### Normalization
+
+Only E.164 format phone is accepted, no normalization for phone login id.
+
+### Raw
+
+No validation and normalization for raw login id.
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -27,8 +27,9 @@ This section defines the login id types and its validation and normalization rul
 
 #### Normalization
 
-- Change domain part to lower case (Compulsory)
-- Change local part to lower case (Configurable, default ON)
+- Case fold and encode the domain part to punycode (Compulsory)
+- Case sensitive local part (Configurable, default OFF, case fold the value for case insensitive)
+- Perform NFKC to the local part (Compulsory)
 - Remove words after `+` sign in the local part (Configurable, default OFF)
 - Remove all `.` sign in the local part (Configurable, default OFF)
 
@@ -57,7 +58,7 @@ user_config:
 
 #### Normalization
 
-- Case insensitive (Configurable, default ON)
+- Case sensitive (Configurable, default OFF, case fold the value for case insensitive)
 - Perform NFKC (Compulsory)
 
 #### Configuration
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -53,12 +53,12 @@ user_config:
 - Disallow username with reserved keywords in [Django list](https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127) (Configurable, default ON)
 - Disallow username with user defined list (Configurable, default empty list)
 - Disallow non-ASCII username (Configurable, default OFF)
-- Disallow username with confusing homoglyphs (Configurable, default ON)
+- Disallow username with confusing homoglyphs (Compulsory)
 
 #### Normalization
 
 - Case insensitive (Configurable, default ON)
-- Perform NFKC (Configurable, default ON)
+- Perform NFKC (Compulsory)
 
 #### Configuration
 ```yaml
@@ -71,9 +71,7 @@ user_config:
           - skygear
           - skygeario
         non_ascii: false
-        block_homoglyphs: true
         case_sensitive: false
-        nfkc: true
     login_id_keys:
       - type: username
         key: username
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -37,7 +37,7 @@ This section defines the login id types and its validation and normalization rul
 ```yaml
 user_config:
   auth:
-    login_id:
+    login_id_handling:
       email:
         case_sensitive: false
         ignore_local_part_after_plus_sign: false
@@ -67,7 +67,7 @@ user_config:
 ```yaml
 user_config:
   auth:
-    login_id:
+    login_id_handling:
       username:
         block_reserved_keywords: true
         excluded_keywords:
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -73,7 +73,7 @@ user_config:
         excluded_keywords:
           - skygear
           - skygeario
-        non_ascii: false
+        ascii_only: false
         case_sensitive: false
     login_id_keys:
       - type: username
--- a/features/323-login-id/type.md
+++ b/features/323-login-id/type.md
@@ -4,11 +4,15 @@
 
 Skygear Auth supports login with login id and password. This document defines the types of login id.
 
-## Validation and Normalization
+## Validation, normalization and unique key checking
 
 Validation and normalization are defined base on login id type. When user signup with login ids, Skygear Auth will validate the login ids and reject the signup if it won't pass. After passing the validation, Skygear Auth will store the normalized login id for login.
 
 When user login with login id, login id key is optional. If login id key is provided, Skygear Auth will normalize the input by type, query the principal and perform the login. If login id key is not provided, Skygear Auth will walk through all login id keys, normalize the input and query the principal. If multiple principals are found, the login process will be stopped and error will be returned.
+
+For some login id types, different login id may represent same identify. E.g. Unicode email and punycode-encoded email. In order to ensure the uniqueness of login id, unique key will be generated per type.
+
+Original login id will be stored in db as `original_login_id` for internal reference only.
 
 ## Types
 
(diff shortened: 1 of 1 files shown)


features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+    login_id_keys:
+      - type: email
+        key: email
+```
+
+### Username
+
+#### Validation
+
+- Disallow username with reserved keywords in [Django list](https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127) (Configurable, default ON)
```
louischan-oursky: I just googled and found [this list](https://github.com/marteinn/The-Big-Username-Blacklist/blob/master/list_raw.txt). Not sure if we should use this too but the django list is quite short.

features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+- Disallow username with user defined list (Configurable, default empty list)
+- Disallow non-ASCII username (Configurable, default OFF)
+- Disallow username with confusing homoglyphs (Configurable, default ON)
+
+#### Normalization
+
+- Case insensitive (Configurable, default ON)
```
louischan-oursky: I guess this means [Unicode case folding](https://godoc.org/golang.org/x/text/cases#Fold), right?

features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+- [Email](#Email)
+- [Username](#Username)
+- [Phone](#Phone)
+- [Raw](#Raw)
+
+### Email
+
+#### Validation
+
+- [RFC 5322 address](https://tools.ietf.org/html/rfc5322#section-3.4.1) (Compulsory)
```
louischan-oursky: So both the local part and the domain part are ASCII only, right? If it is not the case, then we may also need NFKC and Unicode case folding in normalization.

kiootic: Some additional information: https://en.wikipedia.org/wiki/International_email

features/323-login-id/type.md
chpapa: Sounds good. maybe can merge...

features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+- Change domain part to lower case (Compulsory)
+- Change local part to lower case (Configurable, default ON)
+- Remove words after `+` sign in the local part (Configurable, default OFF)
+- Remove all `.` sign in the local part (Configurable, default OFF)
+
+#### Configuration
+```yaml
+user_config:
+  auth:
+    login_id:
```
kiootic: I think using `login_id` as the config key is confusing; maybe call it `login_id_handling`?

features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+### Username
+
+#### Validation
+
+- Disallow username with reserved keywords in [Django list](https://github.com/ubernostrum/django-registration/blob/31478a8acbf705a654565105c791f1ec4cdbf581/src/django_registration/validators.py#L127) (Configurable, default ON)
+- Disallow username with user defined list (Configurable, default empty list)
+- Disallow non-ASCII username (Configurable, default OFF)
+- Disallow username with confusing homoglyphs (Configurable, default ON)
```
kiootic: I think we don't need this option, allowing confusing homoglyphs is almost certainly insecure.

features/323-login-id/type.md
chpapa: I don't think local part nor the domain part of an email address should be ASCII only, it needs to support unicode...

ICANN have a concept called [Universal Acceptance](https://uasg.tech/wp-content/uploads/documents/UASG004-en-digital.pdf), which is to make sure all software support email address and domain name correctly....

To quickly grab the concept, can look at their [list of test cases](https://uasg.tech/wp-content/uploads/documents/UASG004-en-digital.pdf) of domain name and email address a system should support.



ICANN is promoting a concept called Universal Acceptance.

features/323-login-id/type.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Normalization
+
+- Case insensitive (Configurable, default ON)
+- Perform NFKC (Configurable, default ON)
```
kiootic: I also don't think this option is needed, I consider unicode normalization a requirement of correct handling of unicode strings.

features/323-login-id/type.md
louischan-oursky: login_id_handling is too verbose IMO

features/323-login-id/type.md
louischan-oursky: Agree.

features/323-login-id/type.md
carmenlau: Ok, let us change to perform NFKC to all email address, always case fold domain part and case fold the local part if case sensitive setting is Off. And local part default case sensitive setting is Off.

features/323-login-id/type.md
carmenlau: 💯

features/323-login-id/type.md
carmenlau: Sure!

features/323-login-id/type.md
kiootic: > always case fold domain part

We need to encode the domain properly: https://godoc.org/golang.org/x/net/idna

carmenlau: Do you mean conversing the domain to punycode? @louischan-oursky mentioned this too, but we are not sure should we store the unicode version, and the email service should be able to handle the unicode domain?

kiootic: At least we should ensure that a punycode-encoded email should be treated as duplication of raw email.

carmenlau: ummm... maybe we should convert it. We will want to avoid user signup 2 accounts that actually the same email, one is unicode version and another is ascii code version

features/323-login-id/type.md
carmenlau: Cannot think of any better name which is shorter, I changed it to `login_id_handling`. See if we have any other ideas? Otherwise we will use this one.

carmenlau:
Please take a look again when have time. Thanks:)

features/323-login-id/type.md
louischan-oursky: What about `login_id_format`?

features/323-login-id/type.md
chpapa: @kiootic What do you meant by encode? If you meant encode to punycode, I think we would need some special handling.. it is OKAY to match login ID with punycode, but we would probably want to display it as "normal" most of the time...

Alternatively, is it good if we just use the [ValidateForRegistration()](https://godoc.org/golang.org/x/net/idna#ValidateForRegistration) to validate the email address part? and don't do any "normalization" on the domain part?

BTW, If we want to be really good at displaying the domain name securely, can reference [how Firefox handle it](https://wiki.mozilla.org/IDN_Display_Algorithm) -- but I think this is probably not worth the time now and should be another feature issue...

chpapa: @kiootic @carmenlau a tricky thing is if we store domain part in punycode, we probably need to make it easy for users to display the unicode version of the loginID.

Any idea how to do it? Or we left it for developers? @@

Another thing is, if we convert to punycode, do we still need to call `ValidateForRegistration()` to validate the email address part?

BTW, If we want to be really good at displaying the domain name securely, can reference [how Firefox handle it](https://wiki.mozilla.org/IDN_Display_Algorithm) -- but I think this is probably not worth the time now and should be another feature issue (create a feature issue for it? @carmenlau )

features/323-login-id/type.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Email
+
+#### Validation
+
+- [RFC 5322 address](https://tools.ietf.org/html/rfc5322#section-3.4.1) (Compulsory)
+
+#### Normalization
+
+- Case fold and encode the domain part to punycode (Compulsory)
```
chpapa: Does the punycode conversion do the case fold already?

carmenlau:
How about we store the punycode version, but always convert it back to unicode when display (api response)?

But the result will become, if user use a punycode version email to signup, he will see the unicode email.

Thought?

features/323-login-id/type.md
carmenlau: I think no. Punycode conversion only convert it from unicode to ascii, so the case is not converted.

louischan-oursky:
> How about we store the punycode version, but always convert it back to unicode when display (api response)?

Acceptable to me.

chpapa:
> > How about we store the punycode version, but always convert it back to unicode when display (api response)?
> 
> Acceptable to me.

It all got messed up now lol, seems whatever simple thing it is once we get serious enough it is crazily troubles.. I'm reading some devel thread on PostgreSQL and people argue that since different folding can convert to the same Punycode, you need to save the original version to make sure display what was originally intended...

Or we magically help them save it in metadata...

kiootic:
I think developer are not supposed to display the login ID directly anyways, since we perform potentially destructive (i.e. non-reversible) normalization (e.g. case folding, striping `+` part in email, etc.). We just need to provide some way to let developers obtain the original login ID on signup to save it themselves.

chpapa:
I agree with the [Tripartite Identity Pattern](http://habitatchronicles.com/2008/10/the-tripartite-identity-pattern/) which login ID is not suppose to be display publicly...

But on the other hand, I also guess a lot of users would have that expectation, and feel really suprised the User ID got ripped off, as they probably expect the login ID when first sign up would be the Public ID.

How do we "let developers obtain the original login ID on sign up to save it themselves" now? After auth hook?

kiootic:
> How do we "let developers obtain the original login ID on sign up to save it themselves" now? After auth hook?

I expect `before_user_create` should return the original login ID somewhere. I suppose we'll need to add a field somewhere in the payload for this.

chpapa:
> I expect `before_user_create` should return the original login ID somewhere. I suppose we'll need to add a field somewhere in the payload for this.

My first guess / thought too, but it feel like too subtle / difficult to use (users need to create an auth hook just for storing the initial login ID?) anyway guess it needs more thought...

I think we don't want to drag this issue at deadline, let's approve it if nothing serious, but create a follow up feature issue of "Make passing initial login ID as Public ID at sign up more usable", in that we can mention a few idea:

1. Use `before_user_create` hook
2. Save it in metadata automatically
3. Consider adopting Firefox's algo to display the domain name

@carmenlau can help do that?

chpapa:
@carmenlau sorry I have a last min suggestion...

I just chat with some info security friends, they have a different POV: the login ID should be unique -- retaining the original information of the users. Hence if the user signed up with a specific form of Unicode domain, they need to sign in with that exact form next time.

Checking uniqueness should be done separately.

So if we take that into account, the suggestions are:

1. In this spec / features, we should store the login ID in it's original form.
2. Change the new feature spec to: a. help display unicode domain name in a way similar with Firefox to help avoid phishing; b. Check uniqueness for domain name using both IDNA 2008 and 2003 (yes there are two different specs and different tld use different encoding!!!)

kiootic:
@chpapa 

Sorry I don't understand how your suggestion works. Do you mean:
- normalize input login ID to check duplication at sign up (checking uniqueness should be done separately)
- save raw login ID at sign up (store the login ID in it's original form)
- don't normalize login ID at login? ('sign in with that exact form next time')

carmenlau:
I just try different format of username in gmail and have some finding for the normalization rules. This is how gmail handle the user login:
1. case sensitivity and dot: No matter what case I use or I add dots to the username randomly, I can still login my account. For display, it always show me lower case email, and keep the dots in the same position when I signup.
1. plus sign: I cannot use plus sign in both signup and login

Suggestion of handling in auth gear:
1. We will still keep normalized value in login id and add one more column in db for the unique key.
1. Keep normalized value in both signup and login.
1. The db table will have 2 unique constraint, one for login id and another for unique key column.
1. Besides validation and normalization, add one more step to compute the unique key.
1. Assume we don't handle the display in this stage, we may change the email login id like this

    #### Validation

    - [RFC 5322 address](https://tools.ietf.org/html/rfc5322#section-3.4.1) (Compulsory)
    - Disallow `+` sign in the local part (Configurable, default OFF) **Moved from normalization to validation**

    #### Normalization

    - Case fold domain part (Compulsory)
    - Case sensitive local part (Configurable, default OFF, case fold the value for case insensitive)
    - Perform NFKC to the local part (Compulsory)
    - Remove all `.` sign in the local part (Configurable, default OFF)

    #### Unique key

    - Use the normalized value and encode the domain part to punycode with IDNA 2008. (If we want to use IDNA 2003 too, the login id will have 2 unique key, still thinking how to handle, any ideas?)

1. For the other type of login id, value of login id and unique key will be the same.

Remarks for point 4, that means 
1. we will store the unicode form of email in login id and user can only use the same form when login.
1. all dot will be removed, so user will not see the dot in their login id. which is different from how gmail display.

Questions:

1. How to handle 2 unique keys if we use both IDNA 2008 and 2003?
1. Maybe we should store the initial login id (without any processing) too, but we will not use it anywhere. Since we have not handle app may change their setting, we don't have migration to recompute all ids, but we may still want to keep the initial value, in case we want to do it in future.

kiootic:
> How to handle 2 unique keys if we use both IDNA 2008 and 2003?

It seems officially, [IDNA 2008 is intended to replace IDNA 2003](https://www.icann.org/resources/pages/idn-guidelines-2011-09-02-en). Therefore supporting just IDNA 2008 should be enough.

> Maybe we should store the initial login id (without any processing) too, but we will not use it anywhere. Since we have not handle app may change their setting, we don't have migration to recompute all ids, but we may still want to keep the initial value, in case we want to do it in future.

If we store the original login ID in database, I think we can also expose it in API so developers do not need to write hooks to consume them.

chpapa:
Thanks @carmenlau , you basically captured what I meant.

> It seems officially, [IDNA 2008 is intended to replace IDNA 2003](https://www.icann.org/resources/pages/idn-guidelines-2011-09-02-en). Therefore supporting just IDNA 2008 should be enough.

Yea problem is many TLDs seems haven't implemented IDNA 2008. However, I think this is a really edge cases, and I don't think we should spend more resources on this, let's assume **we only take care of IDNA 2008**.

>     2\. all dot will be removed, so user will not see the dot in their login id. which is different from how gmail display.

>     2\. Maybe we should store the initial login id (without any processing) too, but we will not use it anywhere. Since we have not handle app may change their setting, we don't have migration to recompute all ids, but we may still want to keep the initial value, in case we want to do it in future.

I think the two points above are related. The issue is if someone signup with `b.e.n@oursky.com`, his "normalized email login ID" should be `ben@oursky.com` while his public ID, are expected to be `b.e.n@oursky.com`

So we end up have 3 ideas to handle it now:
1. As @kiootic suggested, store the DB and expose API
2. expose it in a hook, let user handle it
3. store it in metadata

Given the nature of metadata, maybe we can simply assume there is a key like `display_id` and automatically set it to the first login ID signed up's original string? Would it be easier? As we have thought that we will have different default key in metadata to display in our user management dashboard in portal in the future... I think Display ID / Public ID is something naturally fits into metadata. We also don't need to have another API

chpapa:
BTW, actually in my original suggestion, I even thought "creating another index to check uniqueness based on IDNA 2008" is not a must to implement now... lol

carmenlau:
For the initial value, I prefer to store it in db, since the value may not for display only, but also for migration if the config is changed in future. And user can change the value in meta data, so storing in meta data is not preferred. I think we can decide how to expose it to user later in the new feature issue. I prefer to store the initial data now, we will lost the data if we don't store it from day one.

I think we can add the unique key now, if we have conclusion of the handling. 

Let me try summarise, 

1. Keep storing normalized value in login id, both login and signup will do the normalization. (Same as the current spec)
1. Add initial_login_id column to store the initial login id, but we will not access it in this stage
1. Add unique_key column to store the punycode encoded of normalized email (IDNA 2008). For the other types, unique_key will be the same as the login id.
1. The db table will have 2 unique constraint, one for login id and another for unique key column
1. Change `+` sign email handling from normalization to validation.

If it looks good, I will update the spec and feature issue accordingly. Thanks all:)

features/323-login-id/type.md
carmenlau: I think it is more than just formatting? e.g. we reject username from list of keywords

louischan-oursky: My only concern is I haven't seen people using the term `handling` in the configuration name.

kiootic: Maybe use `login_id_types`?

chpapa:
> and user can change the value in meta data, so storing in meta data is not preferred. I think we can decide how to expose it to user later in the new feature issue.

Actually, for Public ID / Display ID, it is intended for users to change the value, so I think that's why metadata is a good place to do so.

It sounds like now the `initial_login_id` is just for our internal reference only. BTW instead I think name it `unnormalized_login_id` or `original_login_id` is clearer?

carmenlau:
> Actually, for Public ID / Display ID, it is intended for users to change the value, so I think that's why metadata is a good place to do so.

If we allow user change the public id through meta data, we need to clearly define it is for display only for have another meaning. For example, will it change the login id too if user change the public id in meta data? We can discuss and decide it later. 🙈

> It sounds like now the initial_login_id is just for our internal reference only. BTW instead I think name it unnormalized_login_id or original_login_id is clearer?

Yes, in this stage it is for internal reference only. Let's use `original_login_id`!

carmenlau:
Updated again!

chpapa:
> If we allow user change the public id through meta data, we need to clearly define it is for display only for have another meaning. For example, will it change the login id too if user change the public id in meta data? We can discuss and decide it later. 🙈

I thought if it is in metadata, it should be obvious that's just one of the value we "pre-defined" and won't change anything and not relate with loginID at all. (Well maybe I have a bad assumption myself this time here...)

In fact I think it is quite strange if we handle public ID separately in another set of APIs.

louischan-oursky:
User can have multiple emails as login IDs. How does the public ID work with multiple login IDs?

chpapa:
> User can have multiple emails as login IDs. How does the public ID work with multiple login IDs?

Only initialize the display ID (maybe call it display_name?) once with whatever loginID the user sign up with.

carmenlau:
If the current spec looks good, let us merge this first? And continue the discussion on #388 ?

PR #389, opened 2019-12-09
Update username ascii only

carmenlau:


carmenlau:
Some updates for the username validations:
1. Discussed with @louischan-oursky, we agree that we should use stricter rule as the default. So we changed ASCII only username default ON
1. Add `Ensure valid for PRECIS IdentifierClass profile  (Compulsory)`
1. In most of the cases, user will want to do substring search instead of exact match for the user defined `excluded_keywords`. I renamed the keys as below, as they match the list differently.
    - `block_reserved_usernames`: block username by exact match skygear defined list
    - `excluded_keywords`: substring search by using user defined list

louischan-oursky:
cc @chpapa

chpapa:
>     1. Discussed with @louischan-oursky, we agree that we should use stricter rule as the default. So we changed ASCII only username default ON

Why?

louischan-oursky:
> Why?

Popular internet services such as Facebook, Twitter and Instagram do not allow Chinese, Japanese, etc characters in their username. So I think by default username is ASCII only is least surprising.

chpapa:
> > 
> 
> Popular internet services such as Facebook, Twitter and Instagram do not allow Chinese, Japanese, etc characters in their username. So I think by default username is ASCII only is least surprising.

Noted thx

PR #394, opened 2019-12-30
Sign in with Apple

louischan-oursky:


features/332-sso/api.md
```diff
@@ -80,6 +78,48 @@ skygear.auth.getLoginRedirectResult().then(user => {
...
+function linkApple(providerID: string, callbackURL: string): Promise<User>;
+```
+
+### Example
+
+```typescript
+function LoginScreen() {
+  // For a React Native app, Sign in with Apple must provide two experiences.
+  //
+  // On iOS, the developer must use loginApple which uses AuthenticationServices
```
kiootic: It seems there are more cases here:
- iOS >= 13: use AS
- Android / iOS < 13: use standard OIDC flow

Is it possible to unify these three cases into single call to `loginApple`?

features/332-sso/sign-in-with-apple-prerequisite.md
```diff
@@ -0,0 +1,96 @@
+# How to set up Sign in with Apple as a developer
```
kiootic: I'm not sure if we should include these instructions in this repo. These instructions does not seems specific to Skygear (which can be easily found by searching).

features/332-sso/api.md
louischan-oursky: The provider ID is different. (`apple-app` vs `apple-web`). I expect the developer to use [react-native-device-info](https://github.com/react-native-community/react-native-device-info) to check the version themselves.

features/332-sso/sign-in-with-apple-prerequisite.md
louischan-oursky: All contents are searchable, but scattered in different places. The skygear specific instruction is the UserConfiguration example.

features/332-sso/api.md
kiootic: Is there any reason that the provider IDs have to be different? Would there be problems if developer would want to link the principals of the same Apple ID signed up using `apple-app` & `apple-web` together?

louischan-oursky: The opaque user ID we obtain from Apple is scoped under the team. So as long as the team remains the same, different App IDs and different Services IDs still get the same opaque user ID.

louischan-oursky: I have updated the example to use react-native-device-info

PR #395, opened 2019-12-30
Make login ID key required in login ID APIs

kiootic:


features/329-identity-apis/api.md
```diff
@@ -52,8 +52,9 @@ type Identity = PasswordIdentity | OAuthIdentity | CustomTokenIdentity;
 async function listIdentities(): Promise<Identity[]>;
 
 async function addLoginID(loginID: { [key: string]: string }, realm?: string): Promise<void>;
-async function removeLoginID(loginID: string, realm?: string): Promise<void>;
-async function updateLoginID(oldLoginID: string, newLoginID: { [key: string]: string }): Promise<User>;
+// Login ID key is required to ensure no ambiguity would exist due to normalization.
+async function removeLoginID(loginID: { [key: string]: string }: string, realm?: string): Promise<void>;
```
louischan-oursky: Syntax error near `}: string`

features/329-identity-apis/design.md
```diff
@@ -132,3 +132,11 @@ Conceptually, identities are immutable. Developer should instead:
...
 2. Ensure the new identity is in desired state (e.g. verification)
 3. Delete old identity
+
+## Adding login ID without existing login IDs
+If user without existing login IDs tried to add a new login ID (e.g. a SSO
+user), a password identity would be created, but the password would not be
+known. In this case, the password identity would be created without password
+(i.e. user cannot login using this identity). User should change the password
+using re-authentication through SSO to enable logging in with the password
+identity. Note that reauthentication must be enabled for this to work.
```
carmenlau: Maybe we should also mention the scenario that `Misconfiguration to required login ID amount may cause inability to add login ID`?

refs: https://github.com/SkygearIO/skygear-server/pull/1178#discussion_r362147743

louischan-oursky: After some thoughts, the minimum number of login ID is actually quite hard to uphold. For example, it is completely fine for the user to sign up as a new user via SSO. In this case, the user does not have any login IDs. If the developer sets both the minimum number of `email` and `phone` to be 1, the constraint cannot be upheld in `addLoginID` because only one login ID can be added at a time. It turns out that the constraint is only checked during `signup`. Maybe we should remove the minimum and let user to check this in `before_user_create` hook?

kiootic: @chpapa what do you think about removing the configuration of minimum amount of login ID?

chpapa: I don't see anything particular problem with it.

features/329-identity-apis/design.md
```diff
@@ -132,3 +132,19 @@ Conceptually, identities are immutable. Developer should instead:
...
+If user without existing login IDs tried to add a new login ID (e.g. a SSO
+user), a password identity would be created, but the password would not be
+known.
+
+In this case, the password identity would be created without password (i.e.
+user cannot login using this identity). User should change the password using
+re-authentication through SSO to enable logging in with the password identity.
+Note that reauthentication must be enabled for this to work.
+
+Also, it is possible to misconfigure login ID amount, so that login ID cannot
```
louischan-oursky: I think we can remove this paragraph since we decided to remove the minimum config.

features/329-identity-apis/api.md
```diff
@@ -51,7 +51,7 @@ type Identity = PasswordIdentity | OAuthIdentity | CustomTokenIdentity;
 
 async function listIdentities(): Promise<Identity[]>;
 
-async function addLoginID(loginID: { [key: string]: string }, realm?: string): Promise<void>;
+async function addLoginID(loginID: { [key: string]: string }[], realm?: string): Promise<void>;
```
louischan-oursky: It appears much more natural to me with `{ "a": "b", "c": "d" }` instead of `[{ "a": "b"}, { "c": "d" }]`

kiootic: It is possible to have multiple login IDs with the same key, so list of login IDs is used instead.

Issue #246, opened 2018-08-07
Custom domain to skygear endpoint


carmenlau:
# Description

Support custom domain for application.

# Portal Design

**Remove this section if the feature have no Portal**

# API Design

**Remove this section if the feature have no API**

# Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #396, opened 2019-12-30
Custom domain support

carmenlau:
connect #246 

Added the custom domain spec and updated the skycli spec for managing domains commands.

### DIFF
--- a/features/246-custom-domain/index.md
+++ b/features/246-custom-domain/index.md
@@ -74,9 +74,25 @@ Allow developer to setup custom domain for their application, basically followin
 ## Features Details
 
 - `A` record is used for setting up DNS record, so both apex domain and subdomain can be supported.
-- Developers only need to verify domain once for different subdomains in the same domain within an app.
-  - If developer add multiple subdomains in the same domain at the same time, the verification `TXT` record value should be the same.
-  - If there are verified domain records in the app, developer doesn't need to add the `TXT` for new subdomains in the same domain.
+- After developer verified a domain, Skygear Cluster will verify the root domain by checking `TXT` record. Once the root domain is verified, the root domain will be added to that specific app. Developer will not need to add `TXT` record again for the new subdomains in that app, only `A` record will be checked.
+
+## Scenario
+
+### Scenario 1: Using same root domain in 2 different apps
+
+1. Developer adds `api.example.com` to `myapp1`, he will be requested to update the DNS with `A` record for `api.example.com` and `TXT` record for `_skygear.example.com`. After verification, `myapp1` will have
+    1. Custom domain: `api.example.com`. (Can be viewed through `skycli domain list` command and portal)
+    1. Root domain: `example.com`. (Store in backend only)
+1. Developer adds `myapp1.example.com` to `myapp1`, only `A` record for `myapp1.example.com` will be checked. Since `myapp1` has root domain `example.com` verified.
+1. Developer adds `myapp2.example.com` to `myapp2`, he will be requested to update the DNS with `A` record for `myapp2.example.com` and `TXT` record for `_skygear.example.com`. Since `myapp2` doesn't not verified root domain `example.com`. For developer who want to verify same root domain in multiple apps, they can append the values into the `TXT` record list.
+1. Developer delete all domains in `myapp1`, the root domain `example.com` will also be removed. He will need to verify the domain again.
+
+### Scenario 2: Two apps use the same custom domain
+
+1. Developer adds `api.example.com` to `myapp1`.
+1. Developer adds `api.example.com` to `myapp2`.
+1. Developer completes verification in `myapp1`.
+1. Since `api.example.com` is verified and used in `myapp1`, verifying `api.example.com` in `myapp2` will be rejected.
 
 ## APIs
 
--- a/features/246-custom-domain/index.md
+++ b/features/246-custom-domain/index.md
@@ -7,6 +7,7 @@ Allow developer to setup custom domain for their application, basically followin
 - Allow developer to setup custom domain.
 - Let's Encrypt certificate will be provided by default, certificate will be renew automatically.
 - Developer can configure to use custom SSL certificates for their domain. They will need to update the certificates before it expires in this case.
+- Support configure custom domain to serve the app or redirect to another domain.
 
 ## Use Cases
 
@@ -15,16 +16,16 @@ Allow developer to setup custom domain for their application, basically followin
 1. Add custom domain
 
     ```sh
-    $ skycli domain add myapp.example.com --app=myapp
-    Added domain myapp.example.com successfully!
+    $ skycli domain add www.example.com --app=myapp
+    Added domain www.example.com successfully!
 
     Add following DNS records in your DNS provider.
 
     TYPE      HOST                     VALUE
     TXT       _skygear.example.com     5636486ffc5a4dfebf4a13f480bd9a95
-    A         myapp.example.com          <ingress controller lb ip>
+    A         www.example.com          <ingress controller lb ip>
 
-    After updating DNS records, run `skycli domain verify myapp.example.com` to verify domain.
+    After updating DNS records, run `skycli domain verify www.example.com` to verify domain.
     ```
 
 1. Developer update the DNS records through their DNS provider.
@@ -32,8 +33,8 @@ Allow developer to setup custom domain for their application, basically followin
 1. Ask Skygear Cluster to verify the domain
 
     ```sh
-    $ skycli domain verify myapp.example.com --app=myapp
-    Success! You can now access your app through myapp.example.com.
+    $ skycli domain verify www.example.com --app=myapp
+    Success! You can now access your app through www.example.com.
     Your site may show a security certificate warning until the certificate has been provisioned.
     ```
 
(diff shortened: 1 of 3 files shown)


features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+    ```sh
+    $ skycli domain add myapp.example.com --app=myapp
+    Added domain myapp.example.com successfully!
+
+    Add following DNS records in your DNS provider.
+
+    TYPE      HOST                     VALUE
+    TXT       _skygear.example.com     5636486ffc5a4dfebf4a13f480bd9a95
+    A         api.example.com          <ingress controller lb ip>
```
louischan-oursky: `api.example.com` -> `myapp.example.com`

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+    DOMAIN              VERIFIED         CUSTOM_CERT        SSL_CERT_EXPIRY               CREATED_AT
+    myapp.myapp.com     true             true               2020-11-26 20:00:00 +08:00    2019-11-26 18:00:00 +08:00
+    ```
+
+## Features Details
+
+- `A` record is used for setting up DNS record, so both apex domain and subdomain can be supported.
+- Developers only need to verify domain once for different subdomains in the same domain within an app.
+  - If developer add multiple subdomains in the same domain at the same time, the verification `TXT` record value should be the same.
+  - If there are verified domain records in the app, developer doesn't need to add the `TXT` for new subdomains in the same domain.
```
louischan-oursky: Can we support the following case?

`myapp1` set `myapp1.example.com` as domain.
`myapp2` set `myapp2.example.com` as domain.

I guess TXT record can have multiple lines so the `_skygear.example.com` may have 2 lines to contain the verification codes of the 2 apps?

carmenlau: Yes, if 2 apps use the same domain, they need to do the verification separation. The domains are belongs to app, not user. So the verification only valid within the same app.

For your case `myapp1` and `myapp2`, yes, they should append the verification code in `_skygear.example.com` TXT record.

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+1. Add custom domain
+
+    ```sh
+    $ skycli domain add myapp.example.com --app=myapp
+    Added domain myapp.example.com successfully!
+
+    Add following DNS records in your DNS provider.
+
+    TYPE      HOST                     VALUE
+    TXT       _skygear.example.com     5636486ffc5a4dfebf4a13f480bd9a95
```
kiootic: Not sure If I got it correct, but shouldn't it be `_skygear.myapp.example.com`?

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+    A         myapp.example.com          <ingress controller lb ip>
```
kiootic: It seems many providers use CNAME, is there any reason we use A & TXT instead?

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. (Optional) Setup custom certificate
+
+    ```sh
+    # Create tls secret
+    $ skycli secret create myapp-tls --type=tls --cert=path/to/tls.cert --key=path/to/tls.key --app=myapp
+    Success! Created secret myapp-tls
+
+    # Update domain using custom certificates
+    $ skycli domain update myapp.example.com --tls-secret=myapp-tls --app=myapp
```
kiootic: I suppose after running `skycli domain update`, I can run `skycli secret delete` on previous secret without impact on the operation? We may want to note that the developer need to rotate the certificate manually.

Do we have a way to complete erase a secret from the cluster? Our current implementation keeps the secrets forever.

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+    TYPE      HOST                     VALUE
+    TXT       _skygear.example.com     5636486ffc5a4dfebf4a13f480bd9a95
+    A         myapp.example.com          <ingress controller lb ip>
+
+    After updating DNS records, run `skycli domain verify myapp.example.com` to verify domain.
+    ```
+
+1. Developer update the DNS records through their DNS provider.
+
+1. Call Skygear cloud to verify the domain
```
kiootic: Let's use `Ask Skygear Cluster to verify the domain` instead.

features/246-custom-domain/index.md
```diff
@@ -0,0 +1,70 @@
...
+    # Create tls secret
+    $ skycli secret create myapp-tls --type=tls --cert=path/to/tls.cert --key=path/to/tls.key --app=myapp
+    Success! Created secret myapp-tls
+
+    # Update domain using custom certificates
+    $ skycli domain update myapp.example.com --tls-secret=myapp-tls --app=myapp
+    Success! Updated domain myapp.example.com
+
+    # The domain is now using the custom certificate
+    $ skycli domain ls --app=myapp
```
kiootic: It is `ls` or `list`? The API documentation said list.

features/270-next-skycli/api.md
```diff
@@ -64,10 +64,10 @@ skycli [group] [verb/action] [parameters]
     - [skycli app invoke-function [FUNCTION_NAME] --payload [PAYLOAD_JSON]](commands.md#skycli-app-invoke-functions)
 
 - skycli domain
-    - [skycli domain add [CUSTOM_DOMAIN]](commands.md#skycli-domain-create)
+    - [skycli domain add [CUSTOM_DOMAIN]](commands.md#skycli-domain-add)
```
kiootic: Is there no way to delete the domain, except removing DNS record and let it naturally expire?

features/270-next-skycli/commands.md
```diff
@@ -340,29 +339,42 @@ instructions again.
...
+  A         api.example.com          <ingress controller lb ip>
 
-`skycli domain set-alias api.example.com myapp.[HASH].skygear.io`
+```
+
+### skycli domain update
+
+`skycli domain update [CUSTOM_DOMAIN] --tls-secret=[SECRET_NAME] --use-letsencrypt`
+
+Update domain tls certificates and configure redirect, user can only provide either `tls-secret` or `use-letsencrypt`.
```
kiootic: I think you mean 'must provide either`.

features/270-next-skycli/commands.md
```diff
@@ -340,29 +339,42 @@ instructions again.
...
+
+#### Flags
+
+- `--tls-secret=[SECRET_NAME]` Custom certificate secret name.
+- `--use-letsencrypt` Configure using let's encrypt certs for the given domain.
+- `--disable-redirect` Disable domain redirect.
+- `--redirect-domain` Configure domain redirect, 307 redirect will be performed.
```
kiootic: It seems this flag is not documented?

features/246-custom-domain/index.md
carmenlau: I did some research on some services that provide custom domain features:

### Firebase

When I added `a.example.com`, I need to add `TXT` record to `example.com` to verify the ownership. After I passed the verification, I can use any subdomains (e.g. `b.example.com`, `c.b.example.com`) to all of my apps. It seems that the verification is belong to user, if the subdomain is added by another collaborator, the `TXT` record verification step appear again.

### ZEIT

When I added `a.example.com`, I need to add `TXT` record to `_now.example.com` to verify the ownership. After adding `a.example.com`, we cannot use `example.com` in another account. User can add different subdomains in the same account.

So for us, I think that would be clearer for we to define the domain is belong to app. We used the TXT record verify developer own the domain. After developer verified the domain once, any new subdomains add to that app doesn't need to enter the TXT record again. So the host of the TXT record need to be under TLD+1. For example, if the TLD+1 is example.com, we will test _skygear.example.com. 

We can parse the domain with [EffectiveTLDPlusOne](https://godoc.org/golang.org/x/net/publicsuffix#EffectiveTLDPlusOne) function.

Thought? If it looks good, I can add a section to explain this.

features/246-custom-domain/index.md
carmenlau: Since we will want to support both apex domain and subdomains. I have thought of using ANAME or ALIAS for apex domain, but not all providers support this. At least route 53 doesn't support this.

features/246-custom-domain/index.md
carmenlau: > I suppose after running skycli domain update, I can run skycli secret delete on previous secret without impact on the operation

yes, since we have not delete the k8s secret actually, so  `skycli secret delete` will not accept the existing domain (same as deployment).

> Do we have a way to complete erase a secret from the cluster? Our current implementation keeps the secrets forever.

We don't have this yet, let me open another issue to discuss this.

features/246-custom-domain/index.md
kiootic: Am I correct that the verification process proves that the developer owns the root domain, and wants to assign the root domain to a specific app exclusively?

Do we support this situation?
- `forum.example.com` -> custom domain for `example-forum` app
- `blog.example.com` -> custom domain for `example-blog` app

features/270-next-skycli/commands.md
carmenlau: For the case developer want to update redirect only, they don't need to provide tls-secret or use-letsencrypt. Let me add back disable-redirect and redirect-domain to the command.

features/270-next-skycli/api.md
carmenlau: Let me add the domain remove command :)

features/246-custom-domain/index.md
carmenlau: > Am I correct that the verification process proves that the developer owns the root domain, and wants to assign the root domain to a specific app exclusively?

yes

> Do we support this situation?
> 
> - forum.example.com -> custom domain for example-forum app
> - blog.example.com -> custom domain for example-blog app

Yes, but developers will need to add 2 different TXT records, as they are 2 different apps.

kiootic: Oh I see this has been discussed above. So the root domain assignment is not exclusive.

features/246-custom-domain/index.md
kiootic: I think at least we need a way for developer to erase certification private key for custom domain. If  developer wants to migrate the domain away, we need to ensure we don't keep the key materials longer than necessary.

For simplicity, we may have a cron job to remove unreferenced TLS secrets for now?

carmenlau: I think no only the TLS secret, we can use the job to remove the unreferenced secrets for all types too :)

carmenlau:
@louischan-oursky @kiootic I added a section ced950d to clarify some scenarios (e.g. 2 apps use same root domain) which include what we have discussed, see if it looks good.

features/270-next-skycli/commands.md
carmenlau: Updated and added use case in efed586

Issue #384, opened 2019-11-27
Generate k8s auth token for each apps namespace with default permissions for the tenant


chpapa:
## Description

Some features such as logging, could be easily solved if we can generate an auth token for each app's k8s.

This is a work around for features that's required by developers, but we don't have resources to do it well yet. For example it could be difficult to write a nice users interface (both web and cli) for logging, it might be easier if we just give direct k8s access for users to do it before we can develop the nice UI.

Two todo:

1. run `skycli` to get the k8s token and automatically add it to their local k8s config, show the `use-context` command
2. add an instruction of 1 to portal.skygear.dev under Quick Starts

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
  - [x] https://github.com/oursky/skygear-controller/issues/235
- Client Issues
  - [x] https://github.com/SkygearIO/skycli/issues/192
- Guides Issues
  - [x] https://github.com/SkygearIO/guides/issues/269

PR #398, opened 2019-12-31
k8s access

louischan-oursky:
refs #384

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -281,29 +281,6 @@ static-index        static           /
 static-asset        static           /static
 ```
 
-### skycli app logs
-
-`skycli app logs [DEPLOYMENT_ITEM_NAME]`
-
-Show logs of specific cloud code item (e.g. function, http-handler or http-service).
-
-### skycli app invoke-function
-
-`skycli app invoke-function [FUNCTION_NAME] --payload [PAYLOAD_JSON]`
-
-Invoke cloud function.
-
-#### Example
-```
-$skycli app invoke-function helloworld --payload {"string": "value", "int": 1}
-{"result":"OK"}
-```
-
-#### Flags
-
-- `--payload` Invoke function with payload
-- `--access-key` Skygear auth access token
-
 ## skycli domain
 
 #### Overview
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -264,7 +264,7 @@ $ skycli app update-user-config -f ./user-config.yaml
 Fetch the credentials to access k8s resources with kubectl.
 The credentials are written to the machine's kubeconfig.
 
-The name of the context is `skygear-<app-name>`.
+The name of the context is `skygear-<cluster-name>-<app-name>`.
 
 This command requires the presence of kubectl in PATH.
 
@@ -272,7 +272,7 @@ This command requires the presence of kubectl in PATH.
 
 ```sh
 $ skycli --app myapp app get-k8s-credentials
-Run kubectl config use-context skygear-myapp to switch to the context of this app.
+Run kubectl config use-context skygear-skygear-production-myapp to switch to the context of this app.
 ```
 
 ### skycli app delete-k8s-credentials
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -257,6 +257,32 @@ $ skycli app view-user-config > ./user-config.yaml
 $ skycli app update-user-config -f ./user-config.yaml
 ```
 
+### skycli app get-credentials
+
+`skycli app get-credentials`
+
+Fetch the credentials to access k8s resources with kubectl.
+The credentials are written to the machine's kubeconfig.
+
+The name of the context is `skygear-<app-name>`.
+
+This command requires the presence of kubectl in PATH.
+
+#### Example
+
+```sh
+$ skycli --app myapp app get-credentials
+Run kubectl config use-context skygear-myapp to switch to the context of this app.
+```
+
+### skycli app delete-credentials
+
+`skycli app delete-credentials`
+
+Delete the credentials remotely if exists. Then delete the config from the machine's kubeconfig.
+
+This command requires the presence of kubectl in PATH.
+
 ### skycli app deploy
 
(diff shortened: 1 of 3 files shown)


features/270-next-skycli/commands.md
```diff
@@ -257,6 +257,32 @@ $ skycli app view-user-config > ./user-config.yaml
 $ skycli app update-user-config -f ./user-config.yaml
 ```
 
+### skycli app get-credentials
+
+`skycli app get-credentials`
```
kiootic: I think the subcommand name is unclear: what are the credentials? Something like `get-k8s-credentials` would be clearer.

features/384-k8s-access/index.md
```diff
@@ -0,0 +1,61 @@
...
+- [Authentication webhook](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication)
+- [Authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy)
+
+The most common one is Service Account.
+It is supported natively thus supported by all providers.
+Therefore it is the chosen one.
+
+# Approach
+
+We define a ClusterRole to allow read access to essential resources.
```
kiootic: I think ClusterRole allows access to all resources in the cluster? i.e. pods in other apps can be accessed too.

features/384-k8s-access/index.md
```diff
@@ -0,0 +1,61 @@
...
+We introduce a command to skycli to allow the developer to get credentials.
+skycli uses `kubectl config` to write the credentials into kubeconfig.
+
+Behind the scene, a service account is created per developer per app if not exists.
+A role binding is created to bind the above role to the service account.
+
+# Caveats
+
+The token of a service account is a JWT token that never expire.
+k8s does not support expiring JWT token.
```
kiootic: Does our k8s cluster supports [TokenRequest API](https://jpweber.io/blog/a-look-at-tokenrequest-api/)?

Also, [proxying requests to API server](https://github.com/jetstack/kube-oidc-proxy) may be a viable approach.

features/384-k8s-access/index.md
louischan-oursky: No. The resolution is still bound by the RoleBinding. Since RoleBinding belongs to a namespace. The access is scoped within that namespace. The benefit of using ClusterRole is to avoid creating one role per namespace.

features/384-k8s-access/index.md
```diff
@@ -0,0 +1,61 @@
...
+- apiGroups: ["extensions", "apps"]
+  resources: ["deployments"]
+  verbs: ["get", "watch", "list"]
+```
+
+We introduce a command to skycli to allow the developer to get credentials.
+skycli uses `kubectl config` to write the credentials into kubeconfig.
+
+Behind the scene, a service account is created per developer per app if not exists.
+A role binding is created to bind the above role to the service account.
```
kiootic: Let's use proper noun 'RoleBinding' instead, so that reader would not think any kind of role binding is usable.

features/384-k8s-access/index.md
louischan-oursky: > Does our k8s cluster supports TokenRequest API?

Just googled and it seems that GKE does not support it.

> Also, proxying requests to API server may be a viable approach.

This requires the Auth Gear to become a OIDC provider (so that token can be refreshed).

features/384-k8s-access/index.md
```diff
@@ -0,0 +1,61 @@
...
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["extensions", "apps"]
+  resources: ["deployments"]
+  verbs: ["get", "watch", "list"]
+```
+
+We introduce a command to skycli to allow the developer to get credentials.
+skycli uses `kubectl config` to write the credentials into kubeconfig.
+
+Behind the scene, a service account is created per developer per app if not exists.
```
kiootic: We may want to note that when would the ServiceAccount be deleted:
- explicitly requested using skycli (`delete-credentials`)
- developer is removed from collaborators

features/384-k8s-access/index.md
louischan-oursky: The new approach now uses TokenRequest.

features/270-next-skycli/commands.md
```diff
@@ -257,6 +257,32 @@ $ skycli app view-user-config > ./user-config.yaml
...
 ```
 
+### skycli app get-k8s-credentials
+
+`skycli app get-k8s-credentials`
+
+Fetch the credentials to access k8s resources with kubectl.
+The credentials are written to the machine's kubeconfig.
+
+The name of the context is `skygear-<app-name>`.
```
carmenlau: Since we have different clusters (e.g. staging or production...etc), maybe we should update the context name so that developer can distinguish which cluster they are using?

features/384-k8s-access/index.md
```diff
@@ -0,0 +1,93 @@
...
+when the developer is removed from the collaborators.
+
+# Authorization
+
+We define a ClusterRole to allow read access to essential resources.
+Note that we must never allow access to secrets because
+the token of a Service Account is stored as secret.
+Allowing access to secret means
+different users in the same app can know others' Service Account.
+This role allows read access to Pods, Logs of Pods and Deployments.
```
kiootic: @louischan-oursky @carmenlau do you think it is useful (and safe) for developers to have read access to other generated k8s resources? for example: ingress, service, job.

louischan-oursky: ## Ingress

The developer will know the secret name holding the TLS cert.

## Service

The developer will know the cluster IP of the service.

## Job

The push access token is passed via arguments. The developer has access to pod so they also have access to the pod created by the JobController. The developer will know the push access token by inspecting the pod. The consequence is that it is possible to push image by using the push access token. We have support for running external images. So even the developer abuse the push access token, it is effectively running external images hosted in our registry.

carmenlau: So I think in the current stage, we will give less permissions in the first version. And we can give read access to more different resources type if needed in future.

carmenlau:
If we don't have any further comments, I will merge this spec :)

Issue #375, opened 2019-11-06
Static Website Hosting


chpapa:
## Description

Support hosting a directory of static website on Skygear Cloud directly with CDN automatically configurated.

We have designed part of the static asset support in [CF spec]( https://github.com/SkygearIO/features/blob/master/features/268-next-cf). When we start designing the static asset spec, we can take it as reference, create a new spec and remove the irrelevant parts.

## Portal Design

* List all deployment of static website
* Configurate Custom Domain name for each deployment

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
    - SkygearIO/skygear-server#1201
- Client Issues
- Guides Issues

hidden-dimensions:
When we do it, will it also support react url routing?

chpapa:
> When we do it, will it also support react url routing?

Not entirely sure what that's? Mind explain more? @hidden-dimensions

chpapa:
@louischan-oursky is it closed accidentally? We haven't implemented it nor have the guides updated yet?

louischan-oursky:
Sorry, I merged the feature spec PR and this issue got accidentally closed because the PR description has a `fix` keyword.

kiootic:
For reference, on GCP:

First 5 forwarding rules (i.e. load balancer): $0.025 / hour
Each additional forwarding rule: $0.01 / hour

Load balancer ingress: $0.008 / GB
CDN cache lookup: $0.0075 / 10000 requests
Non-cached content egress: $0.08 - $0.12 / GB
CDN cached content egress: $0.02 - $0.20 / GB
CDN cache fill: $0.04 - $0.15 / GB

chpapa:
@kiootic any explaination of how that pricing applies to us for additional cost for each custom domain? (Ignore the bandwidth first)

kiootic:
If ignoring ingress/egress traffic, the only cost is the forwarding rules. Therefore, each custom domain would cost ~ $0.00066 per hour (0.01 / 15).

Also for reference:
SSL certificates per proxy (i.e. load balancer): 15 (cannot be increased)
Forwarding rules (i.e. load balancer): 45

PR #401, opened 2020-01-06
Update skygear.yaml to use list instead of hash

louischan-oursky:
Update the format skygear.yaml according to [our conventions](https://github.com/SkygearIO/features/blob/master/features/324-config/conventions.md)

### DIFF
--- a/features/268-next-cf/api.md
+++ b/features/268-next-cf/api.md
@@ -229,20 +229,20 @@ Example:
 
 ```yaml
 deployments:
-  function1:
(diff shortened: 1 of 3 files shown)


carmenlau:
@kiootic 
For `permission`: I updated the cloud function issue description, add **Revisit permission config in CF spec** section in (https://github.com/SkygearIO/features/issues/268).
For `type: static`: Update the static asset issue description (https://github.com/SkygearIO/features/issues/375), and we should address it there

Issue #402, opened 2020-01-06
Update tenant config structure to rename user config to app config


carmenlau:
## Description

Currently we have tenant config with app config (inviable to user) and user config (user facing). Per our discussion, we will rename user config to app config, and move all attributes of app config to the root level of tenant config.

TODO:
- Update tenant config struct
- Migrate the existing apps to new tenant config?
- Update controller api path (portal, skycli, controller need to be updated)
- Rename `user-config` commands to `app-config` in skycli

## Portal Design

**Remove this section if the feature have no Portal**

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
  - [x] https://github.com/SkygearIO/skygear-server/issues/1190
  - [x] https://github.com/SkygearIO/skygear-server/issues/1199
  - [x] https://github.com/oursky/skygear-controller/issues/229
  - [x] https://github.com/oursky/skygear-controller/issues/228
- Client Issues
  - [x] https://github.com/SkygearIO/skycli/issues/186
  - [x] https://github.com/SkygearIO/skycli/issues/184
  - [x] https://github.com/oursky/skygear-next-portal/issues/159
  - [x] https://github.com/oursky/skygear-next-portal/issues/164
- Guides Issues
  - [x] https://github.com/SkygearIO/guides/issues/268

louischan-oursky:
Known issues

- [ ] The create-app API will be changed. `skycli app create` will crash with `TypeError: Cannot convert undefined or null to object` but the app is created successfully. Existing users must update their skycli.

PR #403, opened 2020-01-06
Rename UserConfiguration to AppConfiguration

louischan-oursky:
refs #402


louischan-oursky:
@carmenlau I updated the command name to avoid mentioning the term `app` twice.

PR #404, opened 2020-01-07
Add user verification feature specification

### LINKED ISSUE #336: Specify user verification feature
# Description

Specify user verification feature in skygear next auth gear.

* If user removed a verified login ID, the login ID verification state is removed (so if it is added again, it won't be verified)
* If user removed a verified login ID, the verified status of user will be recomputed.
* Should a linked SSO account implies the email of that SSO account is verified? No
* Add APIs to manually unverify a login ID (phone, email)
* Add APIs to put a user into a "verified whitelist"

# API Design

**Remove this section if the feature have no API**

# Open Questions

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

kiootic:
[Rendered document](https://github.com/kiootic/skygear-features/blob/336-user-verification/features/336-user-verification/design.md)

ref #336

### DIFF
--- a/features/285-design-webhooks/events.md
+++ b/features/285-design-webhooks/events.md
@@ -58,16 +58,12 @@ information about the event:
 For BEFORE events about a user, following fields can be mutated:
 - `metadata`: user metadata
 - `verify_info`: verify info of user
-- `is_verified`: verified status of user
 - `is_disabled`: disbled status of user
 
 **NOTE**
 - The mutations would not affect authorization of current operation. e.g.
   disabling user through mutations would not fail an operation requiring
   non-disabled users.
-- `is_verified` is normally derived from `verify_info`. If it is mutated to be
-   different than the derived value, it would be reset to the derived value
-   when `verify_info` is updated.
 
 
 ## Event Types
--- a/features/285-design-webhooks/events.md
+++ b/features/285-design-webhooks/events.md
@@ -58,6 +58,7 @@ information about the event:
 For BEFORE events about a user, following fields can be mutated:
 - `metadata`: user metadata
 - `verify_info`: verify info of user
+- `is_manually_verified`: manually verified status of user
 - `is_disabled`: disbled status of user
 
 **NOTE**
--- a/features/336-user-verification/design.md
+++ b/features/336-user-verification/design.md
@@ -1,0 +1,122 @@
+# User Verification
+
+## Background
+
+User can sign-up using arbitrary login IDs (e.g. email, phone). Developer may
+want to know whether the login ID associated with the user is indeed controlled
+by the user.
+
+## Objectives
+
+- Developers should be able to request verification of a supported login ID of
+  user.
+- Developers should be able to trust a verified login ID is controlled by
+  the associated user.
+- Developers should be able to know whether a specific login ID of user is
+  verified.
+- Developers should be able to manually mark a user as verified.
+- Developers should be able to manually mark a login ID as verified/unverified.
+
+## Design
+
+### Verification State
+
+**Verification State** are states associated with each user, indicating the
+verification status of the user. Conceptually, it has this shape:
+```typescript
+interface VerificationState {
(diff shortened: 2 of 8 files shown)


features/336-user-verification/design.md
```diff
@@ -0,0 +1,122 @@
...
+        RETURN (L_a intersects L_v) is not empty
+```
+
+## Considerations
+
+### Verification Link in Email
+
+It is reported that some email provider may follows the links in the email for
+various purpose (e.g. anti-virus/phishing, prefetching). The built-in
+verification page is triggered with GET method for simplicity, which may cause
```
louischan-oursky: Should the built-in verification page be a form instead?

features/336-user-verification/design.md
```diff
@@ -0,0 +1,122 @@
...
+- explicitly requested using API.
+
+When the verification flow is initiated, a Verification Code would be generated,
+and delivered as described above. The format of Verification Code and expiration
+time can be configured by developers.
+
+The delivered message can be configured using templates. The default template
+includes a link to built-in verification page to verify the user directly.
+Alternatively, developer can update the template to direct user to their app,
+and verify the user using API with the Verification Code.
```
louischan-oursky: I remember we had a discussion on whether authentication is required when the user submits the verification code. Did we have a conclusion?

kiootic: I think we will replace the built-in UIs with auth UI in future? I'm specifying only the changes on API level, and left UI unchanged.

louischan-oursky: We have `/verify_code` and `/verify_code_form`. The first one requires authentication and is intended to be consumed by the SDK. The latter one does not require authentication. When we have auth UI, we are going to remove `/verify_code_form`, right?

kiootic: Yes, I suppose so.

features/285-design-webhooks/events.md
```diff
@@ -58,16 +58,12 @@ information about the event:
 For BEFORE events about a user, following fields can be mutated:
 - `metadata`: user metadata
 - `verify_info`: verify info of user
-- `is_verified`: verified status of user
```
chpapa: Why remove this attribute and how is it related with #336 ?

features/285-design-webhooks/events.md
```diff
@@ -58,16 +58,12 @@ information about the event:
...
 - `is_disabled`: disbled status of user
 
 **NOTE**
 - The mutations would not affect authorization of current operation. e.g.
   disabling user through mutations would not fail an operation requiring
   non-disabled users.
-- `is_verified` is normally derived from `verify_info`. If it is mutated to be
```
chpapa: What about if we keep this, but add a boolean like `is_manually_verified` to `verify_info`? Than we can make the verification white list not a black magic?

If it is a bad idea, maybe just change this description to something like: `is_verified` is normally derived from `verify_info` and `is_manually_verifed` to keep the explaination (assuming it is relevant in context, I only looked at the track changes, not the whole doc)

features/285-design-webhooks/events.md
kiootic: `is_verified` is a computed field. It would make no sense to modify it.

I suppose I'll allow `is_manually_verified` flag to be mutated instead.

features/285-design-webhooks/events.md
kiootic: All keys in `verify_info` map are login IDs; moving `is_manually_verified` into `verify_info` would be very confusing.

This paragraph is describing the behavior of mutating `is_verified`. Since we do not allow mutating `is_verified` flag directly anymore, this paragraph is not relevant.

features/336-user-verification/design.md
carmenlau: I think this description is valid for email verification only. For phone verification we send code only and we assume application will verify user by api. We may want to mention it here?

PR #405, opened 2020-01-13
Specify API versioning in Skygear Controller

louischan-oursky:
This PR mainly captured what the team has discussed and agreed on API versioning in Skygear Controller.

### DIFF
--- a/features/402-controller-api-versioning/index.md
+++ b/features/402-controller-api-versioning/index.md
@@ -1,0 +1,71 @@
+# API Versioning in Skygear Controller
+
+Skygear Controller adopts a simple versioning scheme to handle breaking changes and new feature changes.
+
+The versioning scheme is applied globally on Skygear Controller rather than individual endpoint or resource.
+
+## Format of API Version
+
+API Version starts with `v` and is followed by two integers separated by a dot.
+
+For example,
+
+```
+v2.3
+```
+
+where 2 is the major version and 3 is the minor version.
+
+## Major Version
+
+The first integer in API Version indicates the major version.
+The major version must be incremented when any breaking changes are introduced.
+
+## Minor Version
+
+The second integer in API Version indicates the minor version.
+The minor version must be incremented when any new features are introduced.
+
+When the major version is incremented, the minor version is reset to 0.
+
+## Check Compatible Version
+
+Skygear Controller compares the given API Version against its own version to
+determine whether the versions are compatible.
+The following steps illustrate the algorithm.
+
+1. Check if the given API Version is in [correct format](#format-of-api-version). If not, it is incompatible.
+2. Check if the given major version is equal to own major version. If not, it is incompatible.
+3. Check if the given minor is less than or equal to own minor version. If not, it is incompatible.
+
+## API Version in endpoint
+
+The REST API of Skygear Controller has the API Version specified in the first path component.
+
+For example,
+
+```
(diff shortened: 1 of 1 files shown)


features/402-controller-api-versioning/index.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Major Version
+
+The first integer in API Version indicates the major version.
+The major version must be incremented when any breaking changes are introduced.
+
+## Minor Version
+
+The second integer in API Version indicates the minor version.
+The minor version must be incremented when any new features are introduced.
```
kiootic: It should be 'backward-compatible configuration changes'.

features/402-controller-api-versioning/index.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Skygear Portal and API Version
+
+The API version of Skygear Portal should always be in sync with that of Skygear Controller.
+
+## API Version in AppConfiguration
+
+AppConfiguration has a field `apiVersion` to indicate the API Version.
+Skygear Controller always returns AppConfiguration in its own API Version.
+Skygear Controller must allow updating [Compatible](#check-compatible-version) AppConfiguration without information loss.
```
carmenlau: I think we should mention how gateway handle incompatible app configuration version? 
- If the given major version is less than own major version. Gateway will convert the config on the fly.
- If this given major version is greater than the major version, this case is weird. I think we can just reject it.

features/402-controller-api-versioning/index.md
```diff
@@ -0,0 +1,82 @@
...
+The major version of TenantConfiguration in the storage may be less than the own major version of Skygear Gateway.
+In this case, Skygear Gateway must migrate on the fly.
+
+It is an error if the major version of TenantConfiguration in the storage is greater than the own major version of Skygear Gateway.
+This implies it is possible to downgrade Skygear Cluster at minor version level though some features will become unavailable.
+Downgrading Skygear Cluster at major version level is unsupported.
+
+## API Version in skygear.yaml
+
+skygear.yaml has a field `api_version` to indicate the API Version.
```
chpapa: Since for Skygear Components, we have major.minor.patch, instead of major.minor, would it be confusing? Or doesn't matter? Or not discussed?

louischan-oursky: A patch version should contain bug fix only and no new features. So the patch version part does not matter here.

chpapa: @louischan-oursky I am aware of it, just wondering since skygear.yaml is client facing, would client confuse? Or do we want to be a bit more error-tolerance, that when user type a major.minor.patch in API version, we will ignore the patch? (Or just show an error?)

louischan-oursky: The error produced by skycli is

```json
{
  "name": "Invalid",
  "reason": "ValidationFailed",
  "message": "invalid request body",
  "info": {
    "causes": [
      {
        "kind": "Enum",
        "message": "api_version must be one of the following: \"v2.0\", \"v2.1\"",
        "pointer": "/api_version",
        "details": {
          "expected": [
            "v2.0",
            "v2.1"
          ]
        }
      }
    ]
  }
}
```

I guess it is simple enough for the developer to correct it?

louischan-oursky:
@kiootic This PR should be ready for merge.

PR #406, opened 2020-01-13
Support CDN & Static Assets Hosting

### LINKED ISSUE #375: Static Website Hosting

kiootic:
[Rendered document](https://github.com/kiootic/skygear-features/blob/375-static-cdn/features/375-static-cdn/design.md)

fix #375

### DIFF
--- a/features/375-static-cdn/cdn.md
+++ b/features/375-static-cdn/cdn.md
@@ -33,17 +33,6 @@ When custom domain is added, the cluster controller would provision required
 CDN resources before providing verification instruction (with DNS records) to
 developers.
 
-## CDN DNS Configuration
-
-Most CDN requires the domain DNS to use CNAME record to point to CDN servers.
-However, root domain cannot use CNAME records, and support for DNS Alias is not
-consistent across providers. Therefore:
-- For root domain, an A record would be provided, and CDN cannot be enabled.
-- For subdomains, a CNAME record would be provided, and CDN would be used.
-
-Developers are advised to redirect users to a `www` subdomain in order to use
-CDN.
-
 ## Considerations
 
 ### Traffic Cost
@@ -62,3 +51,31 @@ following reasons:
   use single distribution for all tenant domains.
 - Azure does not support wilcard domains / TLS certificates.
 - CloudFlare requires enterprise subscription to use CDN with wildcard domain.
+
+### Provider-specifc Network Architecture
+For AWS CloudFront, we will provision a Distribution for each custom domain.
+All provisioned Distribution would have the gateway as origin server.
+Custom domains would have a CNAME record points to the Distribution domain.
+
+For GCP Cloud CDN, we will provision a CDN-enabled load balancer for the
+whole cluster.
+The provisioned load balancer would have the gateway as upstream server.
+Custom domains would have a A record points to the load balancer.
+
+For Azure CDN, we will provision a CDN Profile for the whole cluster, and an
+Endpoint for each custom domain.
+All provisioned Endpoints would have the gateway as origin server.
+Custom domains would have a CNAME record points to the Azure CDN endpoint.
+
+Some CDN requires the domain DNS to use CNAME record to point to CDN endpoint.
+However, root domain cannot use CNAME records, and support for DNS Alias is not
+consistent across providers. Therefore:
+- For root domain, an A record pointing to gateway would be provided,
+  and CDN cannot be enabled.
+- For subdomains, a CNAME record would be provided, and CDN would be used.
+
+Developers are advised to redirect users to from root domain to a
+`www` subdomain in order to use CDN.
+
+CDN-managed certificates are not used. Instead, TLS certificates are
+provisioned by the controller, and synchronize to CDN certificate store.
--- a/features/375-static-cdn/cdn.md
+++ b/features/375-static-cdn/cdn.md
@@ -1,0 +1,64 @@
+# Static Asset CDN
+
+## Architecture
+
+### Network Diagram
+
+[![](https://mermaid.ink/img/eyJjb2RlIjoiZ3JhcGggVERcbiAgSU5FVFtDbGllbnRzXSAtLT58Q3VzdG9tIERvbWFpbnwgQ0ROXG4gIElORVQgLS0-IHxEZWZhdWx0IERvbWFpbnwgSUdcbiAgQ0ROW0NETiBOZXR3b3JrXSAtLT4gSUdcbiAgSUdbSW5ncmVzc10gLS0-IEdXe0dhdGV3YXl9XG4gIEdXIC0tPnxzdGF0aWN8IE9bT2JqZWN0IFN0b3JhZ2VdXG4gIEdXIC0tPnxodHRwLXNlcnZpY2V8IE1bTWljcm8tc2VydmljZXNdXG5cdFx0IiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0)](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbiAgSU5FVFtDbGllbnRzXSAtLT58Q3VzdG9tIERvbWFpbnwgQ0ROXG4gIElORVQgLS0-IHxEZWZhdWx0IERvbWFpbnwgSUdcbiAgQ0ROW0NETiBOZXR3b3JrXSAtLT4gSUdcbiAgSUdbSW5ncmVzc10gLS0-IEdXe0dhdGV3YXl9XG4gIEdXIC0tPnxzdGF0aWN8IE9bT2JqZWN0IFN0b3JhZ2VdXG4gIEdXIC0tPnxodHRwLXNlcnZpY2V8IE1bTWljcm8tc2VydmljZXNdXG5cdFx0IiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0)
+
+### CDN-enabled Custom Domain Flow
+
+[Graph](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgREVWIGFzIERldmVsb3BlclxuICAgIHBhcnRpY2lwYW50IEROUyBhcyBETlNcbiAgICBwYXJ0aWNpcGFudCBDVFJMIGFzIENvbnRyb2xsZXJcbiAgICBwYXJ0aWNpcGFudCBDRE4gYXMgQ0ROIFByb3ZpZGVyXG4gICAgcGFydGljaXBhbnQgSzhTIGFzIEt1YmVybmV0ZXNcbiAgICBERVYtPj4rQ1RSTDogQWRkIGN1c3RvbSBkb21haW5cbiAgICBDVFJMLT4-K0NETjogUHJvdmlzaW9uIENETiByZXNvdXJjZXNcbiAgICBDRE4tPj4tQ1RSTDogQ05BTUUgcmVjb3JkXG4gICAgQ1RSTC0-Pi1ERVY6IFZlcmlmaWNhdGlvbiBjaGFsbGVuZ2UgKEEvQ05BTUUgJiBUWFQgcmVjb3JkKVxuICAgIERFVi0-PkROUzogVXBkYXRlIEROUyBSZWNvcmRcbiAgICBERVYtPj4rQ1RSTDogVHJpZ2dlciB2ZXJpZmljYXRpb25cbiAgICBDVFJMLT4-RE5TOiBWZXJpZnkgRE5TIHJlY29yZHNcbiAgICBDVFJMLT4-SzhTOiBDcmVhdGUgSW5ncmVzcyAmIFByb3Zpc2lvbiBUTFMgQ2VydFxuICAgIEs4Uy0-PkNUUkw6IFRMUyBDZXJ0XG4gICAgQ1RSTC0-PkNETjogU3luY2hyb25pemUgVExTIENlcnRcbiAgICBDVFJMLT4-LURFVjogQ3VzdG9tIGRvbWFpbiBpcyByZWFkeSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In19)
+
+## CDN Provider
+
+For simplicity, we will only support GCP Cloud CDN at the moment.
+
+*For the record, CloudFlare only allows up to 20 domains per account.*
+
+## TLS Termination
+
+To use HTTPS with CDN, we will need to terminate TLS connection at CDN network.
+The cluster controller would synchronize the TLS certificates in k8s cluster
+to provider-specific certificate store.
+
+## Supported Domain
+
+For some providers, it takes non-trivial time to provision a CDN-enabled domain.
+Since we do not use wildcard domains, enabling CDN for default domain would
+cause considerable delay in app creation due to CDN provisioning. Therefore,
+we only enable CDN on added custom domain.
+
+When custom domain is added, the cluster controller would provision required
+CDN resources before providing verification instruction (with DNS records) to
+developers.
+
+## CDN DNS Configuration
+
+Most CDN requires the domain DNS to use CNAME record to point to CDN servers.
+However, root domain cannot use CNAME records, and support for DNS Alias is not
+consistent across providers. Therefore:
+- For root domain, an A record would be provided, and CDN cannot be enabled.
+- For subdomains, a CNAME record would be provided, and CDN would be used.
+
+Developers are advised to redirect users to a `www` subdomain in order to use
+CDN.
+
+## Considerations
+
+### Traffic Cost
+
+Since supported providers is limited, it is possible that there exists some
+clusters where the CDN network and cluster network is of different
+providers. In this case, the traffic cost need to be considered, since
+inter-network traffic is usually billed.
+
+### Wildcard Domains & Certificates
+
+Some providers support wildcard domains & TLS certificates. This allows reduced
+provider-specific logic in cluster controller. However, it is not used for
+following reasons:
+- AWS CloudFront use path as cache key per distributions. This means we cannot
+  use single distribution for all tenant domains.
+- Azure does not support wilcard domains / TLS certificates.
+- CloudFlare requires enterprise subscription to use CDN with wildcard domain.
--- a/features/375-static-cdn/config.md
+++ b/features/375-static-cdn/config.md
@@ -10,6 +10,7 @@ deployments:
     path: /
     context: ./frontend/build
     fallback: /index.html
+    expires: 3600
 ```
 
 ## Updated Fields
@@ -17,3 +18,5 @@ deployments:
 - **type**: A new type `static` is added.
 - **context**: If type is `static`: The static content directory.
 - **fallback**: Valid only if type is `static`. The fallback absolute path.
+- **expires**: Valid only if type is `static`. The cache duration in seconds.
+               Default to 3600 (1 hour). Valid values are 0 to 604800 (7 days).
--- a/features/375-static-cdn/config.md
+++ b/features/375-static-cdn/config.md
@@ -10,6 +10,7 @@ deployments:
     path: /
     context: ./frontend/build
     error_page: /404.html
+    index_file: index.htm
     expires: 3600
 ```
 
@@ -18,6 +19,8 @@ deployments:
 - **type**: A new type `static` is added.
 - **context**: If type is `static`: The static content directory.
 - **error_page**: Valid only if type is `static`. Optional.
-                 The 4xx error page absolute path.
+                  The 4xx error page absolute path.
+- **index_file**: Valid only if type is `static`. Optional. Default to `index.html`.
+                  File name of the index page.
 - **expires**: Valid only if type is `static`. Required.
                The cache duration in seconds. Valid values are 0 to 604800 (7 days).
--- a/features/375-static-cdn/config.md
+++ b/features/375-static-cdn/config.md
@@ -9,7 +9,7 @@ deployments:
     type: static
     path: /
     context: ./frontend/build
-    fallback: /index.html
+    error_page: /index.html
     expires: 3600
 ```
 
@@ -17,7 +17,7 @@ deployments:
 
 - **type**: A new type `static` is added.
 - **context**: If type is `static`: The static content directory.
-- **fallback**: Valid only if type is `static`. Optional.
-                The fallback absolute path.
+- **error_page**: Valid only if type is `static`. Optional.
+                 The 4xx error page absolute path.
 - **expires**: Valid only if type is `static`. Required.
                The cache duration in seconds. Valid values are 0 to 604800 (7 days).
--- a/features/375-static-cdn/config.md
+++ b/features/375-static-cdn/config.md
@@ -1,0 +1,22 @@
+# Configuration
+
+## Example
+```yaml
+# skygear.yaml
+
+deployments:
+  - name: assets
+    type: static
+    path: /
+    context: ./frontend/build
+    fallback: /index.html
+```
+
+## Updated Fields
+
+- **type**: A new type `static` is added.
+- **context**: If type is `static`, it represents the local directory containing
+               static assets.
+- **fallback**: Valid only if type is `static`. If the request path failed to
+                match a file in the served asset directory, the URL would be
+                rewritten with this path to be routed again.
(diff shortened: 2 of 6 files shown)


features/375-static-cdn/cdn.md
```diff
@@ -0,0 +1,64 @@
...
+
+### CDN-enabled Custom Domain Flow
+
+[Graph](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgREVWIGFzIERldmVsb3BlclxuICAgIHBhcnRpY2lwYW50IEROUyBhcyBETlNcbiAgICBwYXJ0aWNpcGFudCBDVFJMIGFzIENvbnRyb2xsZXJcbiAgICBwYXJ0aWNpcGFudCBDRE4gYXMgQ0ROIFByb3ZpZGVyXG4gICAgcGFydGljaXBhbnQgSzhTIGFzIEt1YmVybmV0ZXNcbiAgICBERVYtPj4rQ1RSTDogQWRkIGN1c3RvbSBkb21haW5cbiAgICBDVFJMLT4-K0NETjogUHJvdmlzaW9uIENETiByZXNvdXJjZXNcbiAgICBDRE4tPj4tQ1RSTDogQ05BTUUgcmVjb3JkXG4gICAgQ1RSTC0-Pi1ERVY6IFZlcmlmaWNhdGlvbiBjaGFsbGVuZ2UgKEEvQ05BTUUgJiBUWFQgcmVjb3JkKVxuICAgIERFVi0-PkROUzogVXBkYXRlIEROUyBSZWNvcmRcbiAgICBERVYtPj4rQ1RSTDogVHJpZ2dlciB2ZXJpZmljYXRpb25cbiAgICBDVFJMLT4-RE5TOiBWZXJpZnkgRE5TIHJlY29yZHNcbiAgICBDVFJMLT4-SzhTOiBDcmVhdGUgSW5ncmVzcyAmIFByb3Zpc2lvbiBUTFMgQ2VydFxuICAgIEs4Uy0-PkNUUkw6IFRMUyBDZXJ0XG4gICAgQ1RSTC0-PkNETjogU3luY2hyb25pemUgVExTIENlcnRcbiAgICBDVFJMLT4-LURFVjogQ3VzdG9tIGRvbWFpbiBpcyByZWFkeSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In19)
```
carmenlau: The sequence diagram is great! But maybe we will want to add one more section about the implementation details that how we synchronize the TLS cert to the cdn provider? We will do it in the current controller or another k8s crd controller?

kiootic: I was leaving it intentionally vague, since it is an implementation details that would not have visible external effect. In the diagram, 'Controller' entity represents the logical cluster controller, i.e. existing controller and k8s CRD controller (if we opt to use this approach).

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,77 @@
...
+  cache duration of 24 hour.
+
+### Static Deployment Routing
+
+When a request matched a `static` deployment, only its path would be considered
+during routing. The request would be routed as followed:
+
+1. Let `U` be the request URI, let `P` be the routing path.
+2. If file `P` exists, return `P`.
+3. If file `Q` = `P/index.html` exists, return `Q`.
```
louischan-oursky: I think it is equivalent to making `fallback = /index.html` by default? Why do we want to make this a special case? A generalization of `fallback` may be the `try_files` directive of nginx. How do we handle trailing slash?

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,77 @@
...
+4. If `fallback` is configured, set `U` = `U` with path `fallback` and route again.
+5. Otherwise, no file is matched.
+
+**Example**:
+
+Suppose a `static` deployment with name `assets` is configured to serve file in `static` local directory, and a `fallback` path of `/index.html` is configured:
```
louischan-oursky: It would be easier to understand if this line represented in an example configuration that the developer would specify.

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,77 @@
...
+path (`fallback`).
+
+When `static` deployment type is used:
+- The deployment context would be persisted to object storage.
+- A request with matching path would be routing using static deployment routing
+  algorithm (described below).
+- If a file is matched, the gateway would proxy the request to the backing
+  object storage file.
+- If no file is matched, a generic 404 page would be returned.
+- The response would have appropiate cache control header, allowing maximum
```
louischan-oursky: A typical deployment of single page application looks like

```
index.html
assets/js/main.688787d8ff144c502c7f5cffaafe2cc588d86079f9de88304c26b0cb99ce91c6.js
assets/css/main.cc00113f2c36b6d0e501dd832771ce5db7b41c1d77edb22b5fde8baf1c5e449b.css
```

where we want to cache the files in `assets/` for a very long period time while cache `index.html` for a very short period time. How can we support this?

features/375-static-cdn/design.md
kiootic: It's like using nginx directive `index index.html` and `try_files $uri <fallback>`. The step 3 is mainly for the convention of index file, and fallback path is an absolute path. Trailing slashes are ignored.

features/375-static-cdn/design.md
kiootic: For now, we have a fixed cache duration of 24 hours, so changes to `index.html` would takes 24 hours to roll out completely. I'm not sure if how flexible we should allow configuring cache duration, since misconfiguration can cause prolonged stale cached content/performance issues.

To support your use case, we can have cache duration configuration on static deployment, so developer can have two static deployment:
- `path=/, context=./build, fallback=/index.html, duration=1h`
- `path=/assets, context=./build/assets, duration=30d`

Do you think it is flexible enough?

louischan-oursky: This design allows a path-prefix based caching configuration and should be good enough. We may follow nginx to name `duration` `expires`. I guess there is no cache control headers if the developer does not specify it, right?

features/375-static-cdn/design.md
louischan-oursky: Can we document the behavior in terms of nginx directives as in your comment? It would be helpful for some readers who are familiar with nginx.

So `fallback` must be an absolute path, right? Should we clarify this?

kiootic: Clarified routing behavior and configuration.

features/375-static-cdn/design.md
kiootic: I think we'll have a sensible default of 1 hour cache duration. Developer can disable caching by setting it explicitly to 0, but the maximum cache duration is 7 days.

kiootic: Added `expires` configuration.

louischan-oursky: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

The related directives are `public`, `no-cache`, `must-revalidate` and `immutable`. For path that contains hash, we should specify `public, immutable, max-age=<max-age>`.
For index.html, we should specify `public, no-cache, max-age=<max-age>` or `public, must-revalidate, max-age=<max-age>`. Should we support this as well?

kiootic: - `immutable`: It seems quite dangerous. If I understand correctly, it would poison the cache forever if used incorrectly, therefore I don't suggest supporting this.
- `no-cache`/`must-validate`: I think `no-cache` would be used automatically if duration is zero, otherwise `must-validate` would be used.

This should cover most use cases without major problems. I'll update to include the concrete injected cache control headers. Do you have any other suggestions?

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,138 @@
...
+    `https://example.com/index.html` and route again. (step 4)  
+    Route to `assets` and serve file at `./static/index.html`. (step 1)
+
+### Caching
+
+Developer can configure cache duration per `static` deployment. The gateway
+would attach appropiate headers to response according to configuration.
+
+By default, the cache duration is 1 hour. Developers can disable caching by
+setting it explicitly to zero, but the maximum cache duration is 7 days.
```
carmenlau: I think the 1 hour default cache duration may surprise developer that, when they start using custom domain, they may not aware that there is CDN in between and found their site won't update after deploy. 

1. I think maybe we can invalidate the cache after deployment. Since we don't have idea about the user content, so we can only invalidate all. It is not good for performance, cdn will need to fetch all the content again.
2. @louischan-oursky suggest that maybe we can change expires field to mandatory, I also think it is good too. So when user fill in this number they will know that there is CDN in the setup.

Thought?

Also, I think we cannot avoid we need to have a separate command to allow developer invalidate the cache, may the simplest version is just allow invalidate all?

kiootic: > they may not aware that there is CDN in between and found their site won't update after deploy.

I don't think it is surprising: this behavior can be understood quickly by inspecting the response headers, and many major static hosting sites require some times for the static content to update after deployment (e.g. GitHub pages).

Also, making `expires` field mandatory could also confuse the developer: if developers changed the expiry time (e.g. from 1 hour to zero), it would not be reflected immediately, and allow configuring expiry time does not implies the presence of CDN.

In my opinion, I think the current behavior as specified is acceptable, and I think it is a common expectation that it takes some time to update cached static content after deployment.

carmenlau: Maybe we can have copywriting in the successful message (maybe after deploy or configure custom domain), and said there is 1 hour cache for static asset by default?

But do you think we should have a way for user to invalidate cache? In case the developer find some serious mistake after deployment and want to update the cdn immediately.

kiootic: > Maybe we can have copywriting in the successful message (maybe after deploy or configure custom domain), and said there is 1 hour cache for static asset by default?

I think this is good enough, something like 'Your deployment is successful. It may take some time to update cached CDN content'.

> But do you think we should have a way for user to invalidate cache? In case the developer find some serious mistake after deployment and want to update the cdn immediately.

I'd say this is not a must-have feature that we have to implement now. We can revisit it later when enough users request this.

louischan-oursky: I still think invalidation is an important feature. If the developer make a mistake, they can help themselves instead of looking for support from us.

kiootic: > I still think invalidation is an important feature. If the developer make a mistake, they can help themselves instead of looking for support from us.

As mentioned before, many major static website hosts with CDN (e.g. Github Pages, now.sh, surge.sh) does not offer this option, so I don't think it is important.

For example, If I made a mistake when deploying on GitHub Pages, I would expect it takes up to an hour for the CDN cache to update.

On the other hand, some CDN (now.sh, surge.sh) supports cache invalidation on deployment, we can research on how to support this in a later iteration?

Also, for built-in CDN of cloud providers, cache invalidation seems to be rather restricted: GCP Cloud CDN limits at most one invalidation per minute, and Azure CDN have low concurrent invalidation request limit (~50).

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,138 @@
...
+## Design
+
+### Static Deployment
+
+A new type of deployment, `static`, would be supported. This deployment type
+supports configurating a deployment context (`context`) and fallback absolute
+path (`fallback`).
+
+When `static` deployment type is used:
+- The deployment context would be persisted to object storage.
```
louischan-oursky: We need to specify how do we handle symlinks. Some options:

- Silently ignore them (not recommended) 
- Error on encountering them
- Warn on encountering them
- Follow them
- Research on how popular object storage CLI handle them and follow their approach

kiootic: I would prefer to follow the symlink at client side before putting it into archive. It seems S3 CLI do the same thing (follow symlink) by default.

louischan-oursky: Great. On implementation side, we need to detect circular symlinks.

kiootic: I think we can just leave it to the runtime, trying to read content of circular symlink would throw ELOOP error.

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,140 @@
...
+
+**Example**:
+
+Suppose developer configured a `static` deployment as follow:
+```yaml
+  - name: assets
+    type: static
+    path: /
+    context: ./static
+    fallback: /index.html
```
chpapa: I think the term `fallback` is a bit misleading. Instead, maybe better call it "index", and also add configuration for 404 (4xx class of error codes) which is commonly needed too?

kiootic: I think `fallback` is actually quite clear on what it does: if the file is not found, fallback to this file. This config is mainly used for 404 handling and SPA routing, and we do not support other 4xx HTTP code handling (it seems to me there's no need to support so for static assets).

features/375-static-cdn/design.md
chpapa: For serious use case some mistakes can't wait an unknown length of time. Be more specific / have a way to invalidate is better.

features/375-static-cdn/design.md
chpapa: I think we use follow tradition. It actually took me like 0.5 second to realize what fallback could meant. By tradition I think index is the word used by various services from apache/nginx to similar configuration at s3. Who use fallback in similar services?

Moreover, fallback here serve two purpose: both display index page and 4xx error. Those are definitely different. Most static website / asset hosting will need a different design of index page and 404 not found. In fact it would be even better if it is easy to define a different page for 4xx vs 5xx error.

features/375-static-cdn/design.md
kiootic: Can you give examples of what use cases and mistakes we should be considering in the discussion? Given that there exists at least one major production-grade static page host (GitHub Page) that does not support invalidation, I don't think it is essential to support it now.

features/375-static-cdn/design.md
chpapa: p.s. use 4xx instead of specific 404 here just coz S3 do it.

kiootic: > By tradition, I think index is the word used by various services from apache/nginx to the similar configuration at s3.

This configuration is called error document/page in S3, nginx, and apache. I can change its name to `errorPage` if you find it easier to understand.

> Moreover, fallback serves two purposes: both display index page and 4xx error.

That's not correct. In fact, index page are always served from `index.html` if exists. `fallback` configuration is only for 404.

> In fact it would be even better if it is easy to define a different page for 4xx vs 5xx error.

Static routing would only produce 'not found' errors. Other 4xx errors (e.g. 401/403), is not produced. 5xx errors are not always recoverable, and I don't think there are providers supporting custom 5xx pages.

louischan-oursky: I guess S3 allows us to specify the status code because S3 objects may not be publicly accessible. So the status code could be 403 instead of 404.

features/375-static-cdn/design.md
chpapa: Github Page is definitely not very production considering its a free services and most usage are from personal webpage to random note.

An example could be, someone posted a mobile phone no, or a wrong cost on the website, and need to be invalidated immediately, or each minute of delay would result is monetory loss

features/375-static-cdn/design.md
chpapa: @kiootic, in that case, I think we should have `error` and `index`; We could have `index.html` as default but it should be changeable.

chpapa: @kiootic in that case I think we should have, ideally:

* An option to change the index page's name. We should default it to `index.html`, but it should be changeable in case some poor dev is stuck with something else...
* We should have an error page for 404
* Ideally, we should also have an error page config for 5xx errors. 5xx errors are not recoverable, but for UX purposes, say Skygear screw up something and need to display 5xx, some business may prefer to have its own "We're sorry" page, instead of a Skygear Branded or a plain HTML text of 5xx error page. However, if there is some reason make it difficult to implement I can live without it for now.

features/375-static-cdn/design.md
kiootic: > Github Page is definitely not very production considering its a free services and most usage are from personal webpage to random note.

I disagree on this. Free services can be production grade, and many large projects uses GitHub page as main site host (e.g. https://angular.io/, https://facebook.github.io/react-native/, https://getbootstrap.com/).

> An example could be, someone posted a mobile phone no, or a wrong cost on the website, and need to be invalidated immediately, or each minute of delay would result is monetory loss

Cache invalidation is also not an immediate operation; for CloudFront, it takes 10 - 15 minutes to complete as mentioned in documentation. Also, we cannot invalidate client-side/intermediate proxy cache. If developer is concerned about this, they can disable caching on the page or use a short cache duration (e.g. 10 mins).

features/375-static-cdn/design.md
kiootic: > An option to change the index page's name. We should default it to index.html, but it should be changeable in case some poor dev is stuck with something else...

I don't think it is a required feature. Using `index.html` is a convention common enough that there should be no need cater for this at the moment. We can revisit this when there's concrete scenario that request this feature.

> We should have an error page setting for 404

I'll change the `fallback` config to `errorPage` instead then.

> Ideally, we should also have an error page config for 5xx errors. 5xx errors are not recoverable, but for UX purposes, say Skygear screw up something and need to display 5xx, some business may prefer to have its own "We're sorry" page, instead of a Skygear Branded or a plain HTML text of 5xx error page. However, if there is some reason make it difficult to implement I can live without it for now.

If a 5xx error is produced when serving static asset, it is very likely that serving another static asset (i.e. custom error page) would fail too. Therefore it is very likely that the custom 5xx error page would not be served for all 5xx errors, and developers would be confused on why sometimes the custom 5xx error page is not used.

chpapa: @kiootic let's stick with a configuration for `index` and one for `error` then. I think there are not much overhead to maintain an `index` configuration option, this configuration is also very common (S3, apache, nginx, anything I can think of have this option); Given there are also known usage of this (legacy one like `index.htm`, `index.xhtml`, or sub-folder, etc), I don't see why we shouldn't do it now vs have to re-visit it in future.

Good point on 5xx error might not served at all so render the feature not very useful.

features/375-static-cdn/design.md
chpapa: > I disagree on this. Free services can be production-grade, and many large projects use GitHub page as main site host (e.g. https://angular.io/, https://facebook.github.io/react-native/, https://getbootstrap.com/).

Well but all those are not very mission-critical to the business, mostly simple landing page. e.g. I haven't encountered any e-commerce or serious stuff running on it. That's what I meant

> Cache invalidation is also not an immediate operation; for CloudFront, it takes 10 - 15 minutes to complete as mentioned in documentation. Also, we cannot invalidate client-side/intermediate proxy cache. If developer is concerned about this, they can disable caching on the page or use a short cache duration (e.g. 10 mins).

That's why I said "For serious use case some mistakes can't wait an unknown length of time. Be more specific / have a way to invalidate is better."; Even without cache invalidation, nothing is "immediate"...

The unknown is a configuration with a default long values, someone will bombard us for support someday for this, I think we should either:

1. Make the default value short, maybe 10 mins as you proposed is good enough
2. Netlify's approach to this is it adds a fingerprint to all files, but I don't think we can replicate that...?
3. Make the value explicit
4. Enable cache invalidation feature

features/375-static-cdn/cdn.md
```diff
@@ -0,0 +1,81 @@
...
+
+## CDN Provider
+
+For simplicity, we will only support GCP Cloud CDN at the moment.
+
+*For the record, CloudFlare only allows up to 20 domains per account.*
+
+## TLS Termination
+
+To use HTTPS with CDN, we will need to terminate TLS connection at CDN network.
```
chpapa: Does it meant other API calls to MicroServices etc will be in plain text HTTP after CDN? (e.g. from CDN to ingress)

features/375-static-cdn/config.md
```diff
@@ -0,0 +1,23 @@
...
+## Example
+```yaml
+# skygear.yaml
+
+deployments:
+  - name: assets
+    type: static
+    path: /
+    context: ./frontend/build
+    error_page: /index.html
```
chpapa: use something like `/404.html` to avoid misleading sample

features/375-static-cdn/design.md
kiootic: > Well but all those are not very mission-critical to the business, mostly simple landing page. e.g. I haven't encountered any e-commerce or serious stuff running on it. That's what I meant

I think we have different definitions on what is 'production-grade'.

I'll make the cache expiry duration explicit for now.

features/375-static-cdn/design.md
kiootic: There are some additional considerations if we want to include this index page configuration:
- Do we want to support multiple custom index page file name? (e.g. use `index.html` if exist, otherwise `index.htm`)
- Do we want to include a default set of common index page names?

features/375-static-cdn/cdn.md
kiootic: The external TLS connection would be terminated at the CDN, and CDN would initiate another TLS connection to gateway, so plain-text HTTP is not used.

features/375-static-cdn/design.md
chpapa: I think default to `index.html` is good enough and let the user pick those obsoleted one is good enough. Set is a bit too much in our case?

However if we're going deep into it, one common use case is support multiple languages index page depends on Accept-Language and IP address geo... however I think that's a rabbit hole we don't want to dig into it yet...?

features/375-static-cdn/cdn.md
chpapa: Great

features/375-static-cdn/config.md
```diff
@@ -10,6 +10,7 @@ deployments:
     path: /
     context: ./frontend/build
     error_page: /404.html
+    index_file: index.htm
```
louischan-oursky: Not sure if [error_page](http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page) is following nginx's convention. If yes, we should also follow nginx to call this [index](http://nginx.org/en/docs/http/ngx_http_index_module.html#index) instead of `index_file`

features/375-static-cdn/design.md
```diff
@@ -0,0 +1,142 @@
...
+    File is not found, so instead rewrite URL with error page path
+    `https://example.com/index.html` and route again. (step 4)  
+    Route to `assets` and serve file at `./static/index.html`. (step 1)
+
+### Caching
+
+Developer can configure cache duration per `static` deployment. The gateway
+would attach appropiate headers to response according to configuration.
+
+Developers can disable caching by setting it explicitly to zero, but the
```
louischan-oursky: > The unknown is a configuration with a default long values, someone will bombard us for support someday for this, I think we should either:

> 1. Make the default value short, maybe 10 mins as you proposed is good enough
> 2. Netlify's approach to this is it adds a fingerprint to all files, but I don't think we can replicate that...?
> 3. Make the value explicit
> 4. Enable cache invalidation feature

This sentence seems to me it means option 1. Am I correct?

features/375-static-cdn/config.md
kiootic: Not intentionally, and since providers have different name for this configuration (e.g. ErrorDocument for S3/Apache, error_page for nginx), I don't think it is necessary to follow any one of them.

features/375-static-cdn/design.md
kiootic: I intend it to be option 3, but I can see that the wordings can be confusing. I'll amend it further.

features/375-static-cdn/config.md
chpapa: ummm, I do want to clarify (or maybe we need it on the doc), does `error_page` and `index_file` accept same type of parameters, reading it feel like "page" meant a URL (and the example is `/404.html`, while `file` is a file name... is it intended? Maybe need to clarify it and consider it with the naming too...

kiootic: Yes, they accept different value type. `error_page` is an absolute path to the error page, and `index_file` is the file name of the index page.

PR #407, opened 2020-01-16
Clarify the behavior of delete-k8s-credentials

louischan-oursky:
ref #384

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -283,7 +283,7 @@ Run kubectl config use-context skygear-skygear-production-myapp to switch to the
 
 `skycli app delete-k8s-credentials`
 
-Delete the credentials remotely if exists. Then delete the config from the machine's kubeconfig.
+Delete the config from the local machine's kubeconfig.
 
 This command requires the presence of kubectl in PATH.
 


features/270-next-skycli/commands.md
```diff
@@ -283,7 +283,7 @@ Run kubectl config use-context skygear-skygear-production-myapp to switch to the
 
 `skycli app delete-k8s-credentials`
 
-Delete the credentials remotely if exists. Then delete the config from the machine's kubeconfig.
+Delete the config from the local machine's kubeconfig.
```
kiootic: Why not delete the credentials remotely too?

louischan-oursky: The TokenRequest is actually a JWT token. [client-go does not expose a high level API to delete TokenRequest.](https://godoc.org/k8s.io/client-go/kubernetes/typed/core/v1#ServiceAccountExpansion)

kiootic: I suppose we can delete the service account, and it would have the same effect?

louischan-oursky: I have thought about it. See the concerns in the commit message.

PR #409, opened 2020-02-07
Update static asset & CDN spec

kiootic:
ref #375


features/375-static-cdn/design.md
```diff
@@ -60,7 +70,7 @@ Suppose developer configured a `static` deployment as follow:
     type: static
     path: /
     context: ./static
-    error_page: /index.html
+    fallback_page: /index.html
```
chpapa: Add error_page and fallback_page in example?

features/246-custom-domain/index.md
```diff
@@ -76,6 +76,9 @@ Allow developer to setup custom domain for their application, basically followin
 - `A` record is used for setting up DNS record, so both apex domain and subdomain can be supported.
 - After developer verified a domain, Skygear Cluster will verify the root domain by checking `TXT` record. Once the root domain is verified, the root domain will be added to that specific app. Developer will not need to add `TXT` record again for the new subdomains in that app, only `A` record will be checked.
 - Configure custom domain requires root domain ownership verification, so it is not supported if developer can control the subdomain only (e.g. free domain).
+- If developer wants to use CDN with custom domain, the CDN DNS records may be used instead of the provided
+`A` record. In this case, developer can pass a flag to the verification command to skip verification of
```
louischan-oursky: I guess this should be an persistent attribute of custom domain like `redirect_domain` instead of a flag of the verify command. When we later have periodic verification, we will not know whether A record should be skipped.

kiootic:
@louischan-oursky @chpapa 
After further consideration, now I think it's better to check for TXT record only, and gives a warning if other records (CNAME/A/etc.) are not set up properly, instead of giving an error and having a flag to bypass:
- Most other providers verify TXT records only.
- Even if the A record is not present, it does not affect the correctness of system, and it may be intentional by developer, so reporting as error is not appropriate.
- This concern happens if developers do not follow our instructions fully, so reporting it as either warning or error would not matter much, since they're already disregarding the instructions.

What do you think about this?

carmenlau:
> it's better to check for TXT record only

Originally we check the A record since we request the letsencrypt cert for the custom domains. If users don't add the A record, cert manager will keep trying and fail. Is it irrelevant now? As we will have cert in cdn level and cdn provider handle it?

> For GCP Cloud CDN, we will provision a CDN-enabled load balancer every 15 domains, since each load balancer can serve at most 15 TLS certificates.

1. Does CDN enable by default when custom domain is added?

thanks @kiootic

kiootic:
@carmenlau 
> Originally we check the A record since we request the letsencrypt cert for the custom domains. If users don't add the A record, cert manager will keep trying and fail. Is it irrelevant now? As we will have cert in cdn level and cdn provider handle it?

It is developer's responsibility to ensure our load balancer can be reached at the custom domain. Otherwise, Let's Encrypt would not issue the HTTPS certificate.

> Does CDN enable by default when custom domain is added?

Since there is severe limitations on Cloud CDN on their quota, I've discussed with @louischan-oursky and @chpapa and we think dropping the CDN feature is the simplest for now.

Issue #382, opened 2019-11-25
Remove verification page & link


kiootic:
Currently, the verification email includes a link, which mark the user as verified when opened. This is problematic for security reasons:
- [GET request should not cause side effects](https://security.stackexchange.com/a/197005)
- [User should log-in before verification to ensure a typo in email would not cause account takeover](https://security.stackexchange.com/a/71714)

Therefore, to support a secure verification page, we need to require manual action & authentication before verification. However, due to complex login behaviors & methods, and the use of header session transport, it is impractical to implement a generic login page at the moment.

For now, we'll mitigate the existing problem by:
- remove verification GET endpoint
- remove the link in email and show only the code

kiootic:
The team had discussed on that, and we think that:
- removal of these behavior will indeed make default flow incomplete, but we are unable provide a reasonable UX for it without lots of work, otherwise there would be dead-ends in the verification flow.
- it is expected every apps would customize this flow, therefore the impact is minimal, even if the default flow is somewhat incomplete.
- we do not expect the verification page to be gone forever, and we can revisit it when the proposed AuthKit is completed.

kiootic:
For reference, here's some problems we need to address if we want to keep a complete default flow:
- We'll need to read the login state at the verification form page, however we can't do this reliably if header session transport is used. If we only allow cookie transport to be used with default flow, we'll have two UX, which makes an inconsistent developer experience (i.e. 'if you use header transport, you make the verification flow yourself')
- We'll need to allow user to login if not already logged in, however we don't have a generic login page, so we'll have to redirect to developer controlled login page.
- We can add a config option for URL of login page, however most likely developers would customize UX of email template and verification page, so we'll be adding an option which would be unused most of the time. Personally speaking, I think the maintenance burden would outweigh the usefulness.
- After redirecting to login page and user is logged in, where should the user be redirected to? Most likely developer would want to redirect back to verification form page, yet the form page is not a public API at the moment. We'll need to publicly document this form page and committed to maintain this page, which as mentioned before, would most likely went unused.

chpapa:
The problem is the current way we propose to just send an email with verification code, is very half-baked and make the whole experience very incomplete for both type of developers: Developers who don't want to customize the UX or developers who want to.

* It is incomplete for people who don't care the details of UX. Some MVP apps or software don't care too much and would be happy to use the default UX, no customization at all. If we just pass the verification code in the email, it is incomplete flow and they must develop their own verification page.
[...]

Issue #370, opened 2019-10-28
Authorization + Security Setting page for Users


chpapa:
## Description

Support an end-point for login/signup + security setting which users can use directly for their web / mobile apps.

The experience should be similar with how Google different product support the same login and security setting page, or [Auth0's Universal Login](https://auth0.com/docs/universal-login)

Features:

* [skygear.or.custom.endpoint]/authorize:
    * login w/email, phone, username
    * signup
    * SSO login/signup
    * MFA support
    * forgot password
* During signup, if configured:
    * verify by emails/sms
* [skygear.or.custom.endpoint]/settings:
    * Change metadata of user (name, etc)
    * Change password
    * Change email / phone / username securely
    * Enable / change 2FA, generate recovery code
        * CRD of Security Key (in future when the feature is available)
    * Manage SSO linked accounts
    * List all devices(sessions), revoke sessions
    * Check audit log (in future when the feature is available)
    * Check other signed in sites/app (in future when the feature is available)

## Portal Design

**Remove this section if the feature have no Portal**

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

louischan-oursky:
Moved the discussion from Basecamp to here.

---

Ben wrote

> Thanks  Ten Tang Ten  I think it provide some useful insights, here are some of my response. I will add some of these to Github Feature Issue so  Louis Chan Louis  could take reference too.
> - Username and Password in two steps
>   - I think it is preferred to have username and password in two pages, as it provides the opportunities in UX, for Skygear to determine what type of login it is. Some login type may require a different flow (e.g. SAML?)
> - User should be able to add custom signup form fields such as First Name and Last Name to metadata
>   - I think this is an interesting idea, maybe can't have it on phase 1 but seems nice to make it possible to configure custom fields at Sign up form and as you said, put them in MetaData of the signed up user
> - Show the requirements of the password like Nvidia Design
>   - Agree, I think that is something we should ask Frank to address on the design
> - The Terms & Condition Popup
>   - I think we can only accommodate it with HTML template customization (and a good use case)
We’ll ask the user to login again after verifying their email. It’s a more secure practice. Is that okay?
Actually, we won't ask the user to login "again", they just need the sessions to verify the email, and if they don't have it will be asked to login again

Louis wrote

[...]

PR #411, opened 2020-02-10
Auth UI

louischan-oursky:
ref #370 
ref #382 

[Rendered document](https://github.com/louischan-oursky/features/blob/370-auth-ui/features/370-auth-ui/design.md)

### DIFF
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -7,7 +7,6 @@ Sessions of auth gear should be configurable and manageable.
 
 - App Developer/User should be able to manage sessions.
 - App Developer should be able to configure session lifetime.
-- App Developer should be able to configure session token transport.
 
 
 # Proposed Design
@@ -56,32 +55,29 @@ Update session API allows updating name of session. Custom attributes can also
 be updated using master key.
 
 
-## Access / Refresh Token
+## Access token, refresh token and session token
 
-Each session would have an access token, and optionally a refresh token. Both
-type of tokens are opaque: token formats are unspecified and developer should
-not attempt to interpret the content of tokens.
+Each session at least has an access token and a refresh token, and optionally a session token.
+The format of all 3 tokens are opaque. The developer should not attempt to interpret the content of tokens.
 
-Access token would always have a lifetime; access token would be treated as
-invalid after its expiry.
+Access token would always have a lifetime; access token would be treated as invalid after its expiry. The access token must be used in bearer token in `Authorization:` header.
 
 Refresh token can be used to obtain a new access token. If the old access token
 is still valid, the old access token is invalidated; there would be at most one
-valid access token for a session at any time.
+valid access token for a session at any time. The lifetime of the refresh token is the lifetime of the session.
 
-Refresh token would always have a maximum lifetime.
+Session token must be used in cookie. It identifies the session. The lifetime of the session token is the lifetime of the session.
 
-Optionally, a session idle timeout can be specified: refresh token (or access
-token if refresh token is disabled) must be used at least once before the
-timeout, otherwise the session would be expired.
+Optionally, a session idle timeout can be specified: either one token must be used at least once before the timeout, otherwise the session would be expired.
 
 A session is invalidated if its identity is deleted, or its user is disabled.
-A session is expired if its refresh token is expired, or its access token is
-expired (if refresh token is disabled).
-If a session is invalidated, expired, revoked, or logout, its associated access
-token / refresh token would be treated as invalid.
+A session is expired if it reaches its maximum lifetime.
+If a session is invalidated, expired, revoked, or logout, its associated tokens
+would be treated as invalid.
 
 A session is invalid if it is invalidated, expired, revoked or logged out.
+
+### Legacy refresh token flow
 
 For auth gear, if the endpoint requires authentication and the session is detected as invalid,
 the HTTP response includes a header `x-skygear-try-refresh-token: true`.
@@ -102,32 +98,25 @@ if refresh token is available; SDKs would not expose concept of access token /
 refresh token; however, developers should still handle situation where session
 is no longer valid due to various reasons.
 
+### New refresh token flow
 
-## Session Cookie
+The flow is the same but the condition to happen is changed. The SDK must be configured to use `Authorization:` in order to trigger the flow. It is because the session token cannot be refreshed.
 
-Optionally, developer can choose to use cookies to transport session tokens, in
-order to facilitate server-side rendering.
+## Session token
 
-If developer choose to enable session cookie:
-- Refresh token would not be issued. (See [Appendix](#session-cookie-and-refresh-token))
-- Access token would be transported in session cookie.
-- Access token would not be returned from APIs.
-- Gateway would read access token from cookie only, ignoring `Authorization`
-  header. If a token is present in both header and cookie, it would be treated
-  as if session is not found.
-- To prevent unrecoverable failure (i.e. site unaccessible until cookies are cleared),
-  the session cookie would be cleared when the session is not found or invalid,
-  instead of producing error response.
-- Some SDKs (e.g. mobile SDKs) would not be able to consume it.
+The session has a session token that must be used in cookie. Session token is essential for
+server-side rendering app.
+
+To obtain a session token for the app domain, the SDK calls a special endpoint in the app domain. The endpoint accepts the access token and the refresh token and return the session token in cookie.
 
 ### Security Consideration
 
 - HTTPS is required (can be disabled for development purpose).
-- Session cookie is HTTP-only; web app and web SDK would not have access to the
+- Session token is HTTP-only; web app and web SDK would not have access to the
   session tokens.
-- Session cookie is set with `SameSite=lax` by default; generally, cross-domain
+- Session token is set with `SameSite=lax` by default; generally, cross-domain
   requests would not contain the cookie.
-- Session cookie would not be shared across custom domains.
+- Session token is host-only.
 - App developers are responsible to prevent CSRF in their services.
 - For untrusted multi-tenant hosting where sub-domain is allocated to each
   tenant, the top-level domain must be included in Public Suffix List.
@@ -156,12 +145,7 @@ Developer can create a API key for each client. Each API key identifies a
 specific client, and different attributes can be configured for the client:
 - Name
 - Enabled status
-- Session token transport (cookie / `Authorization` header)
 - Session token properties (e.g. lifetime)
-
-Session token transport must be specified at client creation. Developer should
-not change it after creation.
-
 
 # Future Works
 - Audit log should include session attributes for historical records (#340)
@@ -176,6 +160,7 @@ not change it after creation.
 
 [Use cases](./use-cases.md)
 
+[Breaking Changes](./breaking-changes.md)
 
 ## Session Cookie and Refresh Token
 
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -72,7 +72,7 @@ When the RP is a third-party app that has its session management.
 When it receives a ID token from Auth Gear it creates its own session.
 Revoking the OIDC session has no direct effect on the third-party app session, unless both Auth Gear and the third-party app supports OIDC Session Management which offers single logout.
 
-If the RP is the same Skygear App, then revoking the OIDC session works seamlessly. It is because Skygear App does maintain a separate session.
+If the RP is the same Skygear App, then revoking the OIDC session works seamlessly. It is because Skygear App does not maintain a separate session.
 
 ## Access token, refresh token and session token
 
--- a/features/202-session-management/design.md
+++ b/features/202-session-management/design.md
@@ -54,6 +54,25 @@ to be revoked.
 Update session API allows updating name of session. Custom attributes can also
 be updated using master key.
 
+## Different kinds of session
+
+### OIDC session
+
+The sessions created by JSON API or OIDC flow have `client_id`.
+
+### OP session
+
+The sessions created by the web UI of Auth Gear does not have `client_id` because Auth Gear is acting as an OP.
+
+### Revoking sessions
+
+The user can revoke session of any kinds. However, revoking a OIDC session does not necessarily cause logout in the RP.
+
+When the RP is a third-party app that has its session management.
+When it receives a ID token from Auth Gear it creates its own session.
+Revoking the OIDC session has no direct effect on the third-party app session, unless both Auth Gear and the third-party app supports OIDC Session Management which offers single logout.
+
+If the RP is the same Skygear App, then revoking the OIDC session works seamlessly. It is because Skygear App does maintain a separate session.
 
 ## Access token, refresh token and session token
 
--- a/features/370-auth-ui/configuration.md
+++ b/features/370-auth-ui/configuration.md
@@ -1,0 +1,13 @@
+# Configuration
+
+```yaml
+auth_ui:
+  css: |
+    .primary-btn {
+      background-color: purple;
+    }
+  content_secure_policy: "frame-ancestors https://thirdpartyapp.com 'self'"
+```
+
+- `css`: A string of CSS to be included in all pages of Auth UI.
+- `content_secure_policy`: If present, override the default HTTP Header `Content-Secure-Policy` set by Auth UI.
--- a/features/370-auth-ui/design.md
+++ b/features/370-auth-ui/design.md
@@ -1,0 +1,331 @@
+# Auth UI
+
+## Background
+
+Auth UI is a web application in Auth Gear. It aims to become a OpenID Provider.
+
+## Auth Gear acting as OP
+
+After the introduction of Auth UI, Auth Gear acts as an OpenID Provider (OP). API Client is now treated as Relying Party (RP).
+
+Authorization in the context of OpenID Connect is the authorized scopes the user has granted to the RP. Only the value `openid` is officially defined. In the context of Skygear Auth, possible values of scope include `email` and `phone_number`. They correspond to the value of the login ID. If the `email` scope is granted and the identity the user is using does have the `email` claim, then the value is included in the UserInfo Response.
+
+Auth Gear does not support Authorization at the moment. To be backward compatible, existing access tokens can access Data without any restriction. That means the user can change password, list sessions of **ALL** API clients, etc, via Auth Gear. The new access tokens issued by Auth Gear are intended to be consumed in the same fashion so the new and the old behave the same.
+
+In the future Auth Gear can act as an OP for another Skygear App. At that moment Authorization must be introduced. Existing API clients are considered as first party and have the largest scopes. The RP Skygear App must register itself as a new kind of API client with the OP Skygear App. This is analogous to using access token issued by Google cannot access Google API to change Google account password unless the user authorize the corresponding scope.
+
+## Domain
+
+Auth Gear is served in its own domain. It also also be served in a designated custom domain.
+
+### Default domain
+
+If the app is using default domain, then Auth Gear is served at `https://accounts.<appname>.skygearapp.com`. Note that this domain format may conflict with deployment version. Reserved label has higher precedence than deployment version. The choice of `accounts` is borrowed from some other internet services.
+
+### Custom domain
+
+If the app is using custom domain, then Auth Gear is served at the developer's wish.
+
+### Within the same domain of the app
+
+Auth Gear is never be served in the app domain.
+
+## Logout
+
+See [Revoking sessions](/features/202-session-management/design.md#revoking-sessions).
+
+## Page templates
+
+All pages served by Auth UI are rendered with Go standard library package [html/template](https://golang.org/pkg/html/template/).
+This is also the package we use to render HTML email template. So the developer only needs to learn the syntax once.
+
+The content of the template is mainly HTML with some CSS.
+It also contains some JavaScript to add some extra functionalities such as dynamically checking password requirements and toggling of password visibility.
+All pages must be function properly even without JavaScript.
+
+### Customization of page templates
+
+The developer can customize the pages at 2 levels.
+They can keep using the builtin templates and customize the theme.
+Or they can replace the builtin ones with their own ones.
+
+#### Customizing the theme with CSS
+
+By studying the UI design of Auth UI, a few class names can generalized:
+
+- `.primary-txt`: The primary text.
+- `.secondary-txt`: The secondary text.
+- `.primary-btn`: The primary button.
+- `.secondary-btn`: The secondary button.
+- `.anchor`: The anchor.
+- `.text-input`: The text input.
+
+There are a few CSS pseudo classes that are applicable to the UI elements.
+
+- `:link`: Unvisited links.
+- `:visited`: Visited links.
+- `:hover`: Mouse hover state.
+- `:active`: Button being clicked.
+
+Buttons also have two variants.
+The primary button has background color while the secondary one has border.
+
+So the following CSS properties should be able to customize.
+
+- `color`
+- `background-color`
+- `border*`
+
+To customize all combinations, the developer can provide the CSS directly.
+The developer is strongly recommended to write CSS covering only
+the above mentioned classes, pseudo classes and properties.
+There is no validation to enforce this restriction.
+
+#### Customizing with own templates
+
+The developer can also provide their own templates via [the existing template mechanism](../251-templates/index.md).
+
+## The Authorization Endpoint
+
+The Authorization Endpoint `/authorize` allows the user to authenticate with configured authentication methods. It handles MFA challenge during authentication. If the user does not have an account yet, the user can sign up directly.
+
+### The authorization page
+
+The authorization page is the landing page of the authorize endpoint. It lists out the configured IdPs. It shows a text field for login ID. The login ID field is either a plain text input or a phone number input, depending on the type of the first login ID key.
+
+```
+|---------------------------|
+| Login with Google         |
+|---------------------------|
+| Login with Facebook       |
+|---------------------------|
+
+              Or
+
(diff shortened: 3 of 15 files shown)


features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+If the user selects an IdP, the user is redirected to the authorization endpoint of that IdP. If the user continues with a login ID, the user is directed to the password page.
+
+Separating the login ID and the password into two pages allow future improvements, such as [Open ID Provider Issuer Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery). For example, the user can type in their Google account name and then be redirected to the authorization endpoint of Google.
+
+### The signup page
+
+The signup page displays the first login ID key default. Other login ID keys are available to choose.
+
+```
+Sign with email
```
kiootic: Typo: should be sign up.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+- [ ] At least one digit
+- [ ] At least one uppercase English character
+- [ ] At least one lowercase English character
+- [ ] At least one symbols ~`!@#$%^&*()-_=+[{]}\|;:'",<.>/?
+- [ ] At least 8 characters long
+```
+
+### The verification page
+
+The verification page allows the user to verify their login ID. This page requires authenticated user. Unauthenticated access is redirected to the authorization page. Verification is not enforced by the authentication flow. For example, if auto send verification email is enabled, the email is sent behind the scene and the user is redirected to the application immediately after authentication success.
```
kiootic: I think this may need to be a documented entry point to Auth UI, since developers may want to use it in email templates.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+
+#### Example of Token Response with cookie transport
+
+```json
+{}
+```
+
+## Security considerations
+
+### Being embedded in an HTML iframe
```
kiootic: I'm not sure if safari's anti tracking cookie feature would affect this. We may need to look into this.

louischan-oursky: Just tested with the following program

```golang
package main

import (
	"fmt"
	"math/rand"
	"net/http"
	"strconv"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		value := strconv.Itoa(rand.Int())
		cookie := &http.Cookie{
			Name:     "foobar",
			Value:    value,
			Path:     "/",
			HttpOnly: true,
		}
		body := fmt.Sprintf(`
<!DOCTYPE html>
<html>
<head>
<title>/</title>
<body>
/ %s
<iframe src="http://iframe.testsafari.localhost:9999/iframe" />
</body>
</head>
</html>
		`, value)
		http.SetCookie(w, cookie)
		_, _ = w.Write([]byte(body))
	})
	http.HandleFunc("/iframe", func(w http.ResponseWriter, r *http.Request) {
		cookie, err := r.Cookie("foobar")
		value := "<none>"
		if err == nil {
			value = cookie.Value
		}
		body := fmt.Sprintf(`
<!DOCTYPE html>
<html>
<head>
<title>/</title>
<body>
iframe %s
</body>
</head>
</html>
		`, value)
		_, _ = w.Write([]byte(body))
	})
	_ = http.ListenAndServe(":9999", nil)
}

```

Open Safari and go to `http://root.testsafari.localhost:9999`. The iframe cannot read the cookie. However, if the origin is the same (`http://iframe.testsafari.localhost:9999`), cookie can be read in the iframe. I guess as long as we keep the origin being the same, the cookie is considered first party and can be read without problems.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+As of the time of writing, the conformance of OpenID Provider is not a goal.
+However, the implementation must be prepared for conformance in the future.
+
+## Where Auth UI is served
+
+It is served in the same domain as Auth Gear and Microservices.
+
+### Reason 1: Individual gears should be unaware of Custom Domain
+
+Auth UI is going to be responsible for user verification. The verification email includes an URL to Auth UI. This implies Auth Gear has to know where Auth UI is served. When Custom Domain was introduced, it is a feature solely handled by Gateway. Individual gears such as Auth Gear are unaware of Custom Domain. Serving Auth UI in a different domain will break this property. If Auth UI is served in the same domain, it is trivial for Auth Gear to know where it is.
```
chpapa: > The verification email includes an URL to Auth UI.

I'm thinking maybe it is not a must? Auth Gear itself already have a feature to display the page after verification success, maybe we can just reuse that instead of having "yet another customizable page" on auth UI?

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+
+Some example endpoints:
+
+- `/_auth/_/authorize`
+- `/_auth/_/signup`
+- `/_auth/_/forgot_password`
+- `/_auth/_/reset_password`
+- `/_auth/_/verify`
+
+As a side note, Universal Login of Auth0 is served at `https://<app>.auth0.com/u/{login,signup,...}`
```
chpapa: Auth0 support custom domain too (https://auth0.com/docs/custom-domains)

Anyway... but I feel like it is werid to include it here..

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Reason 2: Cookies written by Auth Gear is host-only
+
+A host-only cookie only matches the domain by simple string comparison. While a non host-only cookie matches the domain via the [domain-matches](https://tools.ietf.org/html/rfc6265#section-5.1.3) algorithm. To allow Auth UI to be embedded within an iframe in a Skygear App, the session and its cookie must be shared by both. This means the domain has to be either exact match or domain-matches.
+
+If cookies written by Auth Gear have to be not host-only, then the cookie attribute `Domain` must be specified. If the custom domain is `www.example.com`, then the value of the attribute should be the apex domain `example.com`. So Auth Gear cannot avoid being aware of Custom Domain.
+
+### The actual location of Auth UI
+
+Auth UI is served at the path prefix `/_auth/_/`
```
chpapa: So can we support, as we said, having something like `auth.oursky.com/authorize` when custom domain name need to be supported?

My understanding here is not planned -- but I think this is a required features. The main reason is for user experience, the URL have to be easily recognizable to avoid phishing problem and avoid trust problem with users.

chpapa: and I think it kind of implies AuthUI can't live under `/_auth/`

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+|--------------------------------------|  |----------|
+| Enter in your email or username here |  | Continue |
+|--------------------------------------|  |----------|
+
+Login with a phone number instead.
+```
+
+If the user selects an IdP, the user is redirected to the authorization endpoint of that IdP. If the user continues with a login ID, the user is directed to the password page.
+
+Separating the login ID and the password into two pages allow future improvements, such as [Open ID Provider Issuer Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery). For example, the user can type in their Google account name and then be redirected to the authorization endpoint of Google.
```
chpapa: A side note... have we created the feature issue for the things we have discussed / our Lxxx clients need? Which is to force loginID to use SSO if configured...? And maybe we should mention that here.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,393 @@
...
+- Update metadata
+- Change password
+- Configure MFA
+- Add / Update / remove login ID
+- Manage identities
+- Manage sessions
+
+## The verification endpoint
+
+The verification endpoint `/_auth/_/verify` allows the user to verify their login ID. This page requires authenticated user. Unauthenticated access is redirected to the authorization page. Verification is not enforced by the authentication flow. For example, if auto send verification email is enabled, the email is sent behind the scene and the user is redirected to the application immediately after authentication success.
```
chpapa: As mention above, if there are no advantages, maybe we can merge this with the "classic" verify feature?

features/370-auth-ui/configuration.md
```diff
@@ -0,0 +1,3 @@
+# Configuration
+
+Auth UI shares the configuration of SSO. Both of them need to configure a set of allowed redirect URLs so they should be configured in one place.
```
chpapa: Ummm, but at least CSS are some sort of styling?

And also (as mentioned in feature issue):
- Default callback URL for login / logout?
- Specify the domain for iframe?

features/370-auth-ui/design.md
chpapa: So in consideration with the cookies thing, I'm thinking is it better if we...

A. force certain rules on the domain, if skygear custom domain is `example.com`, then auth UI custom domain must be `*.example.com`; OR
B. we go directly make Auth UI totally independent, so after AuthUI authenticated the users, it returns a token to call back Auth Gear to generate the real session.

features/370-auth-ui/sdk.md
```diff
@@ -0,0 +1,166 @@
...
+// Open a WebView to login. If it is canceled, the promise will be rejected.
+function loginWithWebUI(redirectURI: string, options?: LoginOptions): Promise<User>;
+
+// Open a WebView to access the given Auth UI path.
+// It interacts with the native cookie storage to ensure
+// the session is attached to the WebView correctly.
+function openWebUI(path: string): Promise<void>;
+```
+
+## Web
```
chpapa: Actually if it is a simple URL with parameters, do we really need to provide Javascript API? Is it really better for developers?

louischan-oursky: The actual URL looks something like `http://192.168.1.127:8081/_auth/_/authorize?client_id=api_key&code_challenge=code_challenge&code_challenge_method=S256&redirect_uri=REDIRECT_URI&response_type=code&scope=openid`. It is not simple enough to let the developer to craft themselves.

features/370-auth-ui/design.md
louischan-oursky: That feature is independent of Auth UI. I will create another feature issue for that.

features/370-auth-ui/configuration.md
louischan-oursky: Will do. Supporting default redirect URI means relaxing a required parameter to a optional parameter. I guess it is OK to deviate from the spec by making required parameter optional.

chpapa: Yes I was thinking about the required to optional parameter... but actually I'm not sure if it is a good idea, so consider it a suggestion (Sorry didn't make it clear in first message)

features/370-auth-ui/design.md
chpapa: Thanks yes (sort of independent, but after we implemented that feature we will have some code change on AuthUI to handle the redirect?)

features/370-auth-ui/sdk.md
chpapa: Noted XD

chpapa: I think documentation-wise, we can still explain both the SDK and the URL.. not "that" complicated and some people may prefer one less JS dependency?

features/370-auth-ui/configuration.md
louischan-oursky: I thought about it again and found that supporting default redirect URI is not very useful. An example authorize URL looks like `http://192.168.1.127:8081/_auth/_/authorize?client_id=api_key&code_challenge=code_challenge&code_challenge_method=S256&redirect_uri=REDIRECT_URI&response_type=code&scope=openid`. Even if we allow default redirect URI, the query is still very long. SSO does not support default redirect URI neither.

features/370-auth-ui/design.md
louischan-oursky: https://github.com/SkygearIO/features/issues/412

features/370-auth-ui/design.md
louischan-oursky: I have the following proposal:

1. Auth UI is deployed standalone. It shares the same code base of Auth Gear.
1. Without custom domain Auth UI is served at `https://accounts.<appname>.skygearapp.com`. The choice of `accounts` is borrowed from Google. Later on when we have deployment version, Auth UI has higher precedence. In practice it should not be a problem.
1. Add a new custom domain feature to allow configure a custom domain for Auth UI.
1. Auth UI and Auth Gear somehow can detect if the request domain is a custom domain. If it is a custom domain, use public suffix list to infer the eTLD+1 and write the cookie with `Domain` attribute. The cookies for custom domain are no longer host only and are shared by **all** subdomains including those unknown to Skygear Controller. The cookie name Auth Gear and Auth UI use for storing the access token is called `session`. It may conflict with other systems running other unknown subdomains.
1. If the request domain is not a custom domain, then the domain of the cookie is derived by removing the leftmost label. 
1. `Content-Security-Policy` default value is still `frame-ancestors 'self'`. It means must explicitly set it to something else if they want to use Auth UI in an iframe.

One downside is that it is impossible to make Auth UI share the domain with other gears and microservices.

cc @kiootic

features/370-auth-ui/design.md
louischan-oursky: Let's continue the discussion in https://github.com/SkygearIO/features/pull/411#discussion_r378191375

features/370-auth-ui/sdk.md
louischan-oursky: I guess the developer cannot get rid of the SDK. The SDK [generates the code challenge](https://tools.ietf.org/html/rfc7636#section-4.2) which is not a trivial task.

Of course the documentation should explain the URL parameters, specifically the difference from the OIDC.

features/370-auth-ui/design.md
louischan-oursky: > (sort of independent, but after we implemented that feature we will have some code change on AuthUI to handle the redirect?)

Edited the feature issue to include this suggestion.

louischan-oursky:
The draft was rewritten. The new design is more coherent with OpenID Connect.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,529 @@
...
+                       |                 +---+---+                 |
+                       +-----------------+  SDK  +-----------------+
+                                         +-------+
+```
+
+## Session compatibility with Auth Gear and the SDK
+
+In handling the Authentication Request in the Authorization Endpoint, Auth UI creates a session and store it in cookie. Finally it redirects the user agent back to the app with a authorization code.
+
+In a typical Authorization Code Flow, the app should send a Token Request to the Token Endpoint of Auth UI to obtain a Token Response. However, Token Response is incompatible with Auth Gear and the SDK, where the session transport can either be cookie or header.
```
chpapa: > Token Response is incompatible with Auth Gear and the SDK, where the session transport can either be cookie or header

Why? Token Response normally give a JWT token and after the JWT token is validated. it should be up to the RP to use whatever way to maintain the session transport? Or I didn't see something here?

chpapa: Ah I think I got what it meant after looking at the flow chart. Are we assuming the Auth UI is the OP? If yes the Token Endpoint should be at the Auth UI instead of Auth Gear?

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,529 @@
...
+- Add custom pages after authentication, and finally redirect the user back to the application.
+- Redirect to an external page after authentication, the external page redirects back to Auth UI, and finally redirect the user back to the application.
+
+Introducing the concept of [authentication flow](https://www.keycloak.org/docs/6.0/server_admin/#_authentication-flows) may fulfill some of the above use cases.
+
+## OpenID Provider conformance
+
+Some of the terms used in this session are defined in [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html#Terminology).
+
+Auth UI implements the [Authorization Code Flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) with [PKCE](https://tools.ietf.org/html/rfc7636)
```
chpapa: Totally random, but maybe we can bear in mind we may want to support more OpenID features in future, such as the standard claims (as metadata?), and other optional endpoints like UserInfo or Session management (actually for OpenID standard I think UserInfo is not optional btw)

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,529 @@
...
++-----------------+          +-------+
+```
+
+### New Relationship
+
+After the introduction of Auth UI, it acts an OpenID Provider (OP). API Client is now treated as Relying Party (RP). Microservices and Auth Gear use the API Key of a API Client to identify themselves as RP.
+
+Auth UI does not support Authorization at the moment. To be backward compatible, existing access tokens can access Data without any restriction. That means the user can change password, list sessions of **ALL** API clients, etc, via Auth Gear. The new access tokens issued by Auth UI are intended to be consumed in the same fashion so the new and the old behave the same.
+
+In the future Auth UI can act as an OP for another Skygear App. At that moment Authorization must be introduced. Existing API clients are considered as first party and have the largest scopes. The RP Skygear App must register itself as a new kind of API client with the OP Skygear App. This is analogous to using access token issued by Google cannot access Google API to change Google account password unless the user authorize the corresponding scope.
```
chpapa: `In the future Auth UI can act as an OP for another Skygear App`

Thought experiment — Can 3rd parties Skygear App support “the Facebook Messenger” flow?

It works something like this (assume Skygear-App2 using Skygear-App1 auth):
* Visit www.skygear-app2.com, it display “Ben Cheng From skygear-app1? Click here to Login”
* Redirect to AuthUI of skygear-app1, login directly without input of password

louischan-oursky: [The OpenID Connect spec has something similar but not exactly the same](https://openid.net/specs/openid-connect-session-1_0.html#OPiframe).

Basically iframes and postMessage are used to detect session status in the OP. This technique enables your Point 1.

Point 2 can be achieved with `prompt=none`. (prompt=none is mandatory anyway If OP supports the above iframe)

features/370-auth-ui/design.md
louischan-oursky: > Totally random, but maybe we can bear in mind we may want to support more OpenID features in future, such as the standard claims (as metadata?)

We DO want to support more OpenID features. In fact, [Terry considered standard claims](https://github.com/SkygearIO/features/blob/master/features/323-login-id/design.md#standard-keys). 

> actually for OpenID standard I think UserInfo is not optional btw

Correct. But Auth Gear does not need it at the moment. Implementing UserInfo also requires Authorization which we do not implement now.

features/370-auth-ui/design.md
louischan-oursky: All OIDC endpoints implemented by Auth UI should be conforming to the spec. The Token endpoint is implemented by Auth UI.

The problem is the Token Response is JSON and the access token is inside the JSON body. If the SDK directly calls the Token endpoint of Auth UI, it is impossible to set access token in the cookie. The presence of the forwarding token endpoint in Auth Gear is solely for setting cookie and returning a backward compatible response the SDK expects. Of course, third party RP should call the Token endpoint of Auth UI directly.

features/370-auth-ui/design.md
chpapa: Nice, got another feature issue for it? I think we better cover it in another feature issue.

features/370-auth-ui/design.md
chpapa: icic, that meant if we do it in future it will be a proprietary implementation instead of a standard one... maybe worth to open a feature issue about it too? I don't think it is a must for now. More like good to have...

features/370-auth-ui/design.md
chpapa: Possible to have the Auth Gear call token endpoint instead of client side SDK? seems more "normal" to the normal code flow.

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,529 @@
...
+
+Auth UI as a OP has its own session management. Auth UI session is stateful and is stored in HTTP cookie. The session management API of Auth Gear cannot list, get or delete Auth UI session. Auth UI session and Auth Gear session are not interchangeable. This is analogous to a website utilizing Google as an OP cannot list the sessions the user have on `https://google.com`.
```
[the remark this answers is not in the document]
louischan-oursky: I think we should list the following things in Auth UI

- List all sessions of Auth UI (1)
- List all authorizations. An authorization is per API Client. (2)
  - List all sessions of this authorization (3)

However, should we change the behavior of session management API of Auth Gear? If we do not change anything, the behavior is that listing all sessions is the same thing as the concatenation of all (3)

features/370-auth-ui/design.md
louischan-oursky: Opened https://github.com/SkygearIO/features/issues/413

features/370-auth-ui/design.md
louischan-oursky: Created https://github.com/SkygearIO/features/issues/415

features/370-auth-ui/design.md
louischan-oursky: The flow is documented in [RFC7636 PKCE](https://tools.ietf.org/html/rfc7636#section-1.1). Only the client has the `code_verifier` so the following flow is impossible

```
 SDK                                                     Auth UI                                       Auth Gear
  +                                                         +                                                +
  |               Send Authentication Request               |                                                |
  +--------------------------------------------------------->                                                |
  |                                                         |                                                |
  |                           +-----------------------------+                                                |
  |                           |    Authenticate the user    |                                                |
  |                           +----------------------------->                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                      Write cookie                       |                                                |
  | Redirect to Auth Gear with Authentication Response      |                                                |
  |                                                         |                                                |
  |                                                         |             Follow HTTP redirect               |
  +--------------------------------------------------------------------------------------------------------->+
  |                                                         |                                                |
  |                                                         |              Stuck at the point                |
  |                                                         |  because Auth Gear does not have code verifier |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  |                                                         |                                                |
  +                                                         +                                                +
```

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,531 @@
...
+It also contains some JavaScript to add some extra functionalities such as dynamically checking password requirements and toggling of password visibility.
+All pages must be function properly even without JavaScript.
+
+### Customization of page templates
+
+The developer can customize the pages at 2 levels.
+They can keep using the builtin templates and customize the theme.
+Or they can replace the builtin ones with their own ones.
+
+#### Customizing the theme with CSS
```
chpapa: What about the logo?

louischan-oursky: Added in https://github.com/SkygearIO/features/pull/411/commits/a3a962547bd82dce0c7a3ee11b1668e80baf208c

features/370-auth-ui/design.md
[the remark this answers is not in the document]
kiootic: Conceptually, the three things mentioned are:
- My login sessions in this app (1)
- Third-party apps authorized to access this app on my behalf (2)
- The times that a specific third-party app is using the authorization to access this app (3)

It is obvious that (1) and (2) must be listed. However for (3), users would care neither how many concurrent access from the third-party apps (i.e. number of authorization sessions), nor how the third-party app is accessing the app (i.e. user agent, IP, etc.). Most likely user would only care about the last access time of the third-party app, therefore I suggest omitting (3) instead.

For first-party apps, as a user, I'd expect it to be treated as part of (1), so the session of Auth IO and first-party apps may need to be shared.

louischan-oursky:
@carmenlau @kiootic @chpapa 

I have rewritten this PR. Various specs have been updated to reflect the new design.

features/202-session-management/design.md
```diff
@@ -55,34 +54,50 @@ to be revoked.
...
+
+### Revoking sessions
+
+The user can revoke session of any kinds. However, revoking a OIDC session does not necessarily cause logout in the RP.
+
+When the RP is a third-party app that has its session management.
+When it receives a ID token from Auth Gear it creates its own session.
+Revoking the OIDC session has no direct effect on the third-party app session, unless both Auth Gear and the third-party app supports OIDC Session Management which offers single logout.
+
+If the RP is the same Skygear App, then revoking the OIDC session works seamlessly. It is because Skygear App does maintain a separate session.
```
carmenlau: `Skygear App does maintain a separate session`?  Per our discussion, Skygear App should be the first party app and using the same session?

features/370-auth-ui/design.md
```diff
@@ -0,0 +1,331 @@
...
+
+Auth Gear is served in its own domain. It also also be served in a designated custom domain.
+
+### Default domain
+
+If the app is using default domain, then Auth Gear is served at `https://accounts.<appname>.skygearapp.com`. Note that this domain format may conflict with deployment version. Reserved label has higher precedence than deployment version. The choice of `accounts` is borrowed from some other internet services.
+
+### Custom domain
+
+If the app is using custom domain, then Auth Gear is served at the developer's wish.
```
carmenlau: No sure how custom domain looks like after this change? Originally we only have one domain for all gears and Skygear App. 

After auth ui is supported, developers can configure gears domain and skygear app domain separately?

features/202-session-management/design.md
louischan-oursky: Typo. Should be `does not maintain`

features/370-auth-ui/design.md
louischan-oursky: Yes. I should update the spec of custom domain to reflect this.

Issue #400, opened 2020-01-02
Support multiple contexts in skycli


louischan-oursky:
## Description

Just like kubeconfig, we want to support multiple contexts in skycli.
A common use case is that the developer want to simultaneously login both the staging and the production cluster.

In the discussion with the team, we tend to use the cluster name for both the name of the user and the context. So skycli only support one user per cluster. The developer can still edit the config file manually to achieve multiple users per cluster.

The current format of the config file is still using dictionary while [we agreed on using lists](https://github.com/SkygearIO/features/blob/master/features/324-config/conventions.md). We need to migrate the old config into new one and add versioning.

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #417, opened 2020-02-20
Add skycli commands for multiple clusters

kiootic:
ref #400

### DIFF
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -66,13 +66,20 @@ View the current config.
 #### Example
 ```
 $ skycli config view
-[cluster]
-email=user@skygear.io
-cluster-server=https://1.2.3.4
-admin-key=
-
-[app]
-app=myapp
+api_version: v1
+clusters:
+  - name: skygeario
+    cluster:
+      env: cloud
+      endpoint: 'https://controller.example.com'
+      api_key: api-key
+users: []
+contexts:
+  - name: skygeario
+    context:
+      cluster: skygeario
+      user: skygeario
+current_context: skygeario
 ```
 
 ### skycli config set-cluster
--- a/features/270-next-skycli/commands.md
+++ b/features/270-next-skycli/commands.md
@@ -7,6 +7,7 @@
 #### Flags
 
 - `--app=APP_NAME` Provide app name for commands. Overrides the app provided in `skygear.yaml`.
+- `--context=CONTEXT_NAME` Provide context name for commands. Overrides the current context in config.
 
 #### Example
 ```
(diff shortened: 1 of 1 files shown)


kiootic:
So, we'll have three commands?
- skycli config list-contexts
- skycli config use-context
- skycli config set-cluster

It seems a bit inconsistent to me, since set-cluster would manipulate contexts implicitly, and there is no set-context. That is why I choose to expose clusters in the commands only.

louischan-oursky:
The problem is that what `skycli config set-cluster` is doing now is that it clears the config file and set up a context for the cluster.

skycli should always has a context to work with. A context is a combination of cluster and user. A context can without a user.

`skycli config set-cluster` adds a new cluster. If the cluster already exists, it is an error. Otherwise create a new context with the new cluster. (Should we switch to the newly created context?)

`skycli auth login` adds a new user. If the current context has no user, the context is updated in place. If the current context has user, create a new context with the cluster of the current context and the logged in user, switch to the new context.

skycli will have a global flag `--context` to override `current_context` in the config file.

Thoughts?

kiootic:
The issue described that we'll support only one user per cluster for now, i.e. only one context per cluster. Is that correct?

Anyways, `set-cluster` need to be changed. I'm thinking of having `set-cluster` always update the context and cluster with specified name according to the input parameters, so it's an upsert operation to clusters and contexts with cluster name as key.

kiootic:
I'll add the three commands mentioned before (`get-contexts`, `use-context`, set-cluster`). Contexts can be added/updated through `set-cluster`, and logging in would affect the user of current context.

louischan-oursky:
Thanks. Can you also introduce the global `--context` flag? Actually it is original use case I want to support. It is particularly useful for `skycli app get-k8s-credentials`.

louischan-oursky:
Maybe we should also review whether we still keep `skycli config view`

kiootic:
> Maybe we should also review whether we still keep skycli config view

I would suggest making `skycli config view` output the config yaml verbatim.

features/270-next-skycli/commands.md
```diff
@@ -66,13 +66,20 @@ View the current config.
...
-[app]
-app=myapp
+api_version: v1
+clusters:
+  - name: skygeario
+    cluster:
+      env: cloud
+      endpoint: 'https://controller.example.com'
+      api_key: api-key
+users: []
```
louischan-oursky: For completeness, either include the user `skygeario` here or omit `user` in the context `skygeario`.

kiootic: This output is possible, if there is no user associated with the `skygeario` context. In that case, the context would refer to an nonexistent user and `auth login` would create/update it.

features/270-next-skycli/commands.md
```diff
@@ -64,26 +66,35 @@ View the current config.
...
+Configure Cluster endpoint and API key, it is provided by cluster admin.
+If there is no current context set, the configured Cluster would be set as the
+current context.
+If there exist a cluster with same name, it is updated with provided
+information.
 
 #### Flags
 
-- `--cluster=[CLUSTER_NAME]` Provide cluster name, currently only `skygeario` is supported.
+- `--cluster=[CLUSTER_NAME]` Provide cluster name.
```
carmenlau: After this update, if user want to set their own cluster without prompt, they will need to provide `cluster` flag? If yes, please update the example below. Maybe `skycli config set-cluster --cluster=mycluster --endpoint=https://mycluster-controller --api-key=api_key` ?

PR #421, opened 2020-03-05
Add spec for Skygear Auth OIDC auth flow

carmenlau:
[Rendered Document](https://github.com/carmenlau/features/blob/370-auth-ui/370-auth-ui/oidc.md)


370-auth-ui/oidc.md
```diff
@@ -0,0 +1,351 @@
+# OpenID Connect (OIDC) Identity Provider
+
+## Overview
+
+Skygear Auth offers authentication flow which conforms OIDC specification. You can integrate Skygear Auth as OIDC auth provider in your application or use Skygear Platform with Skygear Microservices directly.
```
louischan-oursky: Standardize the terms.

"OIDC auth provider" -> "OpenID Provider" or "OP" (This is the term used in OIDC spec)
"auth provider" -> "OP"

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,351 @@
...
...udEFwcFxuICBwYXJ0aWNpcGFudCBBcHBCYWNrZW5kXG4gIHBhcnRpY2lwYW50IFNreWdlYXJBdXRoXG4gIENsaWVudEFwcC0-PkFwcEJhY2tlbmQ6IFVzZXIgY2xpY2sgbG9naW5cbiAgQXBwQmFja2VuZC0-PlNreWdlYXJBdXRoOiBBdXRob3JpemF0aW9uIGNvZGUgcmVxdWVzdFxuICBTa3lnZWFyQXV0aC0-PkNsaWVudEFwcDogUmVkaXJlY3QgdG8gYXV0aG9yaXphdGlvbiBwcm9tb3RlXG4gIENsaWVudEFwcC0-PlNreWdlYXJBdXRoOiBBdXRob3JpemF0aW9uIGFuZCBjb25zZW50XG4gIFNreWdlYXJBdXRoLT4-QXBwQmFja2VuZDogQXV0aG9yaXphdGlvbiBjb2RlXG4gIEFwcEJhY2tlbmQtPj5Ta3lnZWFyQXV0aDogQXV0aG9yaXphdGlvbiBjb2RlICsgY2xpZW50IGlkICsgY2xpZW50IHNlY3JldFxuICBTa3lnZWFyQXV0aC0-PlNreWdlYXJBdXRoOiBWYWxpZGF0ZSBhdXRob3JpemF0aW9uIGNvZGUgKyBjbGllbnQgaWQgKyBjbGllbnQgc2VjcmV0XG4gIFNreWdlYXJBdXRoLT4-QXBwQmFja2VuZDogVG9rZW4gcmVzcG9uc2UgKElEIHRva2VuICsgYWNjZXNzIHRva2VuICsgcmVmcmVzaCB0b2tlbilcbiAgQXBwQmFja2VuZC0-PlNreWdlYXJBdXRoOiBSZXF1ZXN0IHVzZXIgZGF0YSB3aXRoIGFjY2VzcyB0b2tlblxuICBTa3lnZWFyQXV0aC0-PkFwcEJhY2tlbmQ6IFJlc3BvbnNlIHVzZXIgZGF0YVxuICBBcHBCYWNrZW5kLT4-QXBwQmFja2VuZDogQ3JlYXRlIEFwcEJhY2tlbmQgbWFuYWdlZCBzZXNzaW9uXG4gIEFwcEJhY2tlbmQtPj5DbGllbnRBcHA6IFJldHVybiBBcHBCYWNrZW5kIG1hbmFnZWQgc2Vzc2lvbiIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0Iiwic2VxdWVuY2UiOnsic2hvd1NlcXVlbmNlTnVtYmVycyI6dHJ1ZX19LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)
```
louischan-oursky: Typo. "Redirect to authorization promote" -> "Redirect to authorization endpoint". Apply to other flow chart as well.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,351 @@
...
+### Parameters
+
+#### OIDC client Metadata
+
+Some parameters are defined in OIDC, see [ClientMetadata](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata).
+
+They are
+
+- `client_name`
+- `application_type`
```
louischan-oursky: This is included in OIDC spec but [removed in OAuth](https://tools.ietf.org/html/draft-ietf-oauth-dyn-reg-30). Actually we do not need it?

For the portal use case, we offer a "profile" for the developer to choose from. The profiles include

Server side relying party

```yaml
grant_types: ["authorization_code", "refresh_token"]
response_types: ["code"]
```

Native applications (including mobile applications and desktop applications)

```yaml
grant_types: ["authorization_code", "refresh_token"]
response_types: ["code"]
```

Single page application

```yaml
grant_types: ["authorization_code"]
response_types: ["code"]
```

Server side rendering application

```yaml
grant_types: []
response_types: ["none"]
```

carmenlau: Yes, OP actually don't read it . I removed it

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,341 @@
...
+- `login_uri`
+- `grant_types`
+- `response_types`
+
+#### Skygear-specific
+
+- `client_id`: OIDC client id.
+- `client_secret`: OIDC client secret, used in RP integration.
+- `access_token_lifetime`: OIDC access token lifetime in seconds, default to 1800.
+- `session_lifetime`: OIDC refresh token lifetime in seconds, default to max(access_token_lifetime, 86400). Must greater than or equal to `access_token_lifetime`.
```
kiootic: I think we don't need configuration of session here? Since there is no concept of session in OIDC, we should instead have configuration of refresh token here?

carmenlau: Yes, you are right. I changed the name back to `refresh_token_lifetime`, that would be clearer.

carmenlau:
> We also mention we support Authorization Server Metadata 

Updated. I defined some of the metadata values, see if it looks good, specially the `claims_supported`.

> The ID token must be signed. I guess we will have a per app key pair to sign ID token?

I added a section for id token signing, not sure if it is a good idea to store the keys in app config. If it is not good, I think we can store it in `TenantConfiguration` level. But controller may needs to add apis to manage the keys. Open to discuss.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+
+No difference from the spec.
+
+#### redirect_uri
+
+No difference from the spec.
+
+#### client_id
+
+It is always required. The value must be a valid Skygear API Key.
```
kiootic: We should clarify that we use OIDC client ID as API key, or just omit the description on API key.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+
+No difference from the spec.
+
+#### expires_in
+
+No difference from the spec.
+
+#### refresh_token
+
+No difference from the spec. Absent in [SPA](#case-2-single-page-web-app-spa) flow.
```
kiootic: I think the link should be updated?

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+
+`RS256` is supported.
+
+#### claims_supported
+
+The values will be `["sub", "iss", "aud", "exp", "iat"]`.
+
+### code_challenge_methods_supported
+
+The values will be `["plain", "S256"]`
```
kiootic: We only support `S256`.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+- `grant_types`
+- `response_types`
+
+#### Skygear-specific
+
+- `client_id`: OIDC client id.
+- `client_secret`: OIDC client secret, used in RP integration.
+- `access_token_lifetime`: OIDC access token lifetime in seconds, default to 1800.
+- `refresh_token_lifetime`: OIDC refresh token lifetime in seconds, default to max(access_token_lifetime, 86400). Must greater than or equal to `access_token_lifetime`.
+- `session_idle_timeout_enabled`: Indicate whether OIDC session idle timeout is enabled, default to `false`.
```
kiootic: Session idle timeout is not client-specific, so the config can be removed here.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+    - "https://app-backend-endpoint.com"
+    grant_types:
+    - "authorization_code"
+    - "refresh_token"
+    response_types:
+    - "code"
+```
+
+- Should include app backend endpoint in `redirect_uris` which handle code to access token exchange
+- Generic OIDC RP run authentication code flow, and token endpoint will return refresh token. So `grant_types` should be [`authorization_code`, `refresh_token`] and `response_types` should be [`code`].
```
kiootic: The OIDC flow should be 'Authorization' Code flow, please use the correct name for the other cases as well.

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,455 @@
...
+  clients:
+  - redirect_uris:
+    - "https://client-app-endpoint.com"
+    grant_types: []
+    response_types:
+    - "none"
+```
+
+- Should include client app endpoint in `redirect_uris`. When Auth Gear set the Idp Session, it will redirect back to client app with empty result. Only authorized uris can be redirected.
+- First party web app use Idp session for authentication, no OIDC session will be created. So `grant_types` should be [] and `response_types` should be ["none"].
```
kiootic: It should be 'no OIDC grants would be returned', since there is no concept of session in OIDC.

370-auth-ui/oidc.md
louischan-oursky: But to be fully compliant with the spec, we must support `plain`.

370-auth-ui/oidc.md
```diff
@@ -290,6 +290,10 @@ The values will be `["sub", "iss", "aud", "exp", "iat"]`.
 
 The values will be `["plain", "S256"]`
 
+### revocation_endpoint
+
+No difference from the spec. See [Revocation Request](https://tools.ietf.org/html/rfc7009#section-2.1) for detail.
```
louischan-oursky: It's better to document the endpoint. The authorization endpoint is `/oauth2/authorize` so the revocation endpoint could be `/oauth2/revoke`.

We should also document that the token endpoint is `/oauth2/token`.

370-auth-ui/oidc.md
kiootic: I think PKCE spec does not require support of `plain` transformation. Quoting from spec:
- 4.2: 'Clients are permitted to use "plain" only if they cannot support "S256" for some technical reason and know via out-of-band configuration that the server supports "plain".'
- 4.4.1: 'If the server supporting PKCE does not support the requested
   transformation, the authorization endpoint MUST return the
   authorization error response with "error" value set to
   "invalid_request".'

This shows the spec allow server to return error response in case "plain" transformation is not supported.

370-auth-ui/oidc.md
carmenlau: Yes, you are right. I removed the session idle config.

But would like to clarify the session idle behavior. So if user access any api with Idp session / refresh token / access token, the last access at of session will be updated. If last access at pass the session idle timeout, all those session and tokens will be invalidate?

370-auth-ui/oidc.md
carmenlau: Maybe update to support `S256` first? And we will only use `S256` actually :)

370-auth-ui/oidc.md
kiootic: Allow me to clarify:
- IdP session: have session cookie
- Offline grant: have refresh token & access tokens

IdP sessions and offline grants are independent of each other (even if user logged out from IdP session, the offline grant is still valid). Session idle behavior only applies to IdP sessions.

louischan-oursky: Should we have idle timeout for refresh token? Access Token should be kept simple and does not have idle timeout.

kiootic: We do not plan to expose manual refresh function in SDK, and it'll use refresh token regularly according to the expiry of access token, so I think it's mostly meaningless.

louischan-oursky: How should a third party banking mobile application configure the settings if it wants idle timeout?

kiootic: I suppose the use case is that refresh token should expire soon if the app is inactive (i.e. in background). In this case, we can separate refresh token lifetime into two config: one for maximum lifetime of the offline grant, one for the lifetime of individual refresh token.

ref: http://docs.identityserver.io/en/stable/topics/refresh_tokens.html

louischan-oursky: So the refresh token itself is also refreshed?

kiootic: Yes, it is a possible solution. However, I think we don't need to include these mechanism for now.

louischan-oursky: So the conclusion is that both access token and refresh token has maximum lifetime only. Refresh token is static and never refresh.

carmenlau: thanks @louischan-oursky @kiootic :)

370-auth-ui/oidc.md
```diff
@@ -0,0 +1,452 @@
...
+
+```json
+{
+  "token_type": "bearer",
+  "access_token": "access.token",
+  "refresh_token": "refresh.token"
+}
+```
+
+## Authorization Server Metadata Document
```
kiootic: I think we should also document every endpoints (e.g. userinfo, jwks_uri), and note down which fields belongs to which document (I think both document can have same fields?)

louischan-oursky:
Combined here https://github.com/SkygearIO/features/pull/425

PR #422, opened 2020-03-05
Move obsoleted specs into /features/obsoleted

louischan-oursky:


louischan-oursky:
Updated. Please tell me if I classified the specs incorrectly.

chpapa:
Looking at it I'm thinking maybe it is time to re-organize this repo to have folders based on features instead of feature issues number? 

So the repo become a state of capture of latest Skygear's design? 

Opened a feature issue for this: https://github.com/SkygearIO/features/issues/428

PR #423, opened 2020-03-06
Add spec for Auth UI, gateway and SDK

louischan-oursky:


370-auth-ui/sdk.md
```diff
@@ -0,0 +1,306 @@
...
+      <button onClick={onClickOpenSettings}>Open Settings</button>
+      <iframe src={url} />
+    </div>
+  );
+}
+```
+
+## The open source AppAuth library
+
+The open source AppAuth library has three variants, namely Android, iOS and JS. The JS one is for web. However, our JS SDK must support three common JS environments browser, Node and React Native. Therefore AppAuth is not being used.
```
kiootic: I think AuthAuth for JS can be used in Node too? Quoting from their README:
> The library is designed for use in `Web Apps`, `Node.js` CLI applications,
`Chrome Apps` and applications that use `Electron` or similar frameworks.

370-auth-ui/web-ui.md
```diff
@@ -0,0 +1,171 @@
...
+- `background-color`
+- `border*`
+
+### Customizing with own templates
+
+The developer can also provide their own templates via [the existing template mechanism](../251-templates/index.md).
+
+## The authorization endpoint
+
+The authorization endpoint `/authorize` is the entry point of OIDC. The user can sign up, sign in, perform multi-factor authentication, reset forgotten password and authorize access.
```
kiootic: We may want to add a version prefix here? e.g. `/v1/authorize`.

370-auth-ui/web-ui.md
```diff
@@ -0,0 +1,171 @@
...
+- Manage OAuth authorizations
+- Manage OAuth grants
+- Update metadata
+- Change password
+- Configure multi-factor authentication
+- Manage identities (e.g. login IDs and connections with external IdPs)
+
+### The verification endpoint
+
+The verification endpoint `/verify` allows the user to verify their login ID. This page requires authenticated user. Unauthenticated access is redirected to the root endpoint. Verification is not a strict requirement. For example, if auto send verification email is enabled, the email is sent behind the scene and the user is redirected to the application immediately after performing the Authorization Code Flow successfully.
```
kiootic: If an unauthenticated user goes to the verification endpoint from email, do we redirect the user back to the verification endpoint after authentication, or the user need to open the link again from email?

370-auth-ui/gateway.md
```diff
@@ -0,0 +1,211 @@
...
+
+The resolution of custom domain is purely table-driven. The developer is free to assign domain to their app and gears.
+
+## Delegating request authentication to Skygear Auth
+
+Skygear Gateway initiate subrequest to the resolve endpoint of Skygear Auth after resolving the app. The subrequest does not contain body. The gateway must remove `x-skygear-*` headers from the original request before initiating the subrequest. The gateway must copy `x-skygear-*` headers from the response of the subrequest to the original request. If `x-skygear-session-valid` is `false` and `x-skygear-session-transport` is `cookie`, then the gateway must clear the cookie named by `x-skygear-session-cookie-name`.
+
+## Authenticating the request
+
+The resolve endpoint `/resolve` looks at `Authorization:` and `Cookie:` to authenticate the request. `Cookie:` has higher precedence. Regardless of the result, the status code is always 200. If the request is anonymous, the response does not contain any `x-skygear-*` headers. Otherwise the response contains the following headers.
```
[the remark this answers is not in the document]
louischan-oursky: OK. Let me remove this requirement and `x-skygear-session-cookie-name`.

370-auth-ui/sdk.md
louischan-oursky: But not React Native :(

370-auth-ui/web-ui.md
louischan-oursky: But the Authorization Server Metadata is a well known static endpoint. The metadata itself does not support versioning.

370-auth-ui/sdk.md
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// The promise resolves to nothing. It is because after calling this function,
+// the URL is changed. The app is destoryed so the return value is not usable at all.
+function authorize(options: AuthorizeOptions): Promise<void>;
+
+// exchangeToken looks at window.location
+// It checks if error is present and rejects with OAuthError.
+// Otherwise if code is present, make a token request.
+function exchangeToken(): Promise<User>;
```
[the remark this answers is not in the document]
louischan-oursky: `state` is always there.

`state=state&code=code` or `state=state&error=invalid_request`.

If we only return `state` for the success case, it is a bit weird.

370-auth-ui/web-ui.md
kiootic: For example, [Google's discovery document](https://accounts.google.com/.well-known/openid-configuration) has a versioned authorization endpoint: `https://accounts.google.com/o/oauth2/v2/auth`

The protocol itself may be relatively stable, but the behavior may have breaking changes, so adding API versioning would be a good practice.

370-auth-ui/sdk.md
kiootic: I found this: https://github.com/FormidableLabs/react-native-app-auth

Is it suitable?

370-auth-ui/sdk.md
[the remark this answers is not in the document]
kiootic: Yes, we can return both. I mean `exchangeToken` would returns `{ user: User, state: unknown}`, or throws `class OAuthError { message: string; state: unknown; }`, so that developer don't need to extract the state from the URI manually.

370-auth-ui/web-ui.md
louischan-oursky: I mean the `/.well-known/openid-configuration` endpoint is not versioned at all. Today the `authorization_endpoint` is `https://accounts.google.com/o/oauth2/v2/auth` but tomorrow it may be `https://accounts.google.com/o/oauth3/v3/auth`. However as long as it is a OAuth 2.0 or OpenID Connect 1.0 metadata document, all endpoints should behave according to the spec.

The OIDC client should not interpret the versioning scheme of the endpoint URL. Instead the client look at other metadata to find out what features the AS supports. If the OAuth were 3, I guess the standard committee will introduce another well known endpoint for the AS to advertise its capabilties?

370-auth-ui/gateway.md
```diff
@@ -0,0 +1,205 @@
...
+
+If this header is absent, it means that the original request does not contain any session.
+
+If the value is `true`, it indicates the original request has valid session. More headers will be included.
+
+If the value is `false`, it indicates that the original request has an invalid session.
+
+### x-skygear-session-transport
+
+Tell the session transport used to resolve the session when `x-skygear-session-valid` is present.
```
kiootic: Is this useful? I don't think the backend service would need to be aware of where the token is in the request.

The only case I can think of is to use the cookie/header in another sub-request in backend. However, the backend should already know where to look for the token, so there is no need to read this header.

370-auth-ui/web-ui.md
kiootic: Yes, but we cannot just throw away OAuth2 & OIDC when we support OAuth3. Then, we'll have two authorize endpoint for each protocols, and their respective discovery document would refer to individual versioned authorized endpoint: e.g. `/v1/authorize` for OAuth2, `/v2/authorize` for OAuth3.

louischan-oursky: I prefer the prefix `/oauth2/authorize`, what do you think?

370-auth-ui/gateway.md
louischan-oursky: This header is only useful if we allow the upstream server to easily reuse the cookie or authorization header. Currently we do not ban the upstream server from doing this. The proper way is to support client credentials with admin scope in the future.

Let me remove this header from the spec first.

370-auth-ui/web-ui.md
kiootic: I think it's good enough for authorization endpoint.

370-auth-ui/sdk.md
louischan-oursky: The AppAuth library is not particularly useful in our use case. The most challenge part of the SDK IMO is the refresh flows, which are not covered by AppAuth. The PKCE utilities are already implemented in our SDK. POSTing form request can be done with `fetch` easily.

kiootic: Okay I understand. Then, we can just note down there exists this popular official library, and we may want to test our server-side implementation against it.

370-auth-ui/web-ui.md
louischan-oursky: Addressed by introducing a redirect mechanism in the root endpoint.

370-auth-ui/sdk.md
```diff
@@ -0,0 +1,309 @@
...
+```typescript
+import React, { useCallback } from "react";
+import skygear from "@skygear/react-native";
+import {Button, View} from "react-native";
+function SettingsScreen() {
+  const onPressEditProfile = useCallback(() => {
+    // Application-specific settings
+  }, []);
+  const onPressSecuritySettings = useCallback(() => {
+    skygear.auth.openURL("https://accounts.myapp.com").catch(e => {
```
carmenlau: Maybe we should teach developer to go to `https://accounts.myapp.com/settings` instead of `https://accounts.myapp.com`? Although `/` will redirect to `/settings` if user logged in, but it will show non OIDC login page if the Idp session is not valid.

370-auth-ui/web-ui.md
```diff
@@ -0,0 +1,183 @@
...
+- Manage OAuth authorizations
+- Manage OAuth grants
+- Update metadata
+- Change password
+- Configure multi-factor authentication
+- Manage identities (e.g. login IDs and connections with external IdPs)
+
+### The verification endpoint
+
+The verification endpoint `/verify` allows the user to verify their login ID. This page requires authenticated user. Unauthenticated access is redirected to the root endpoint with `redirect_uri=/verify`. Verification is not a strict requirement. For example, if auto send verification email is enabled, the email is sent behind the scene and the user is redirected to the application immediately after performing the Authorization Code Flow successfully.
```
carmenlau: Would like to know how user do verification for native app? User will complete the authorization flow without the app, and user will need to login again outside the app in browser?

370-auth-ui/web-ui.md
```diff
@@ -0,0 +1,183 @@
...
+- Add custom pages after authentication, and finally redirect the user back to the application
+- Redirect to an external page after authentication, the external page redirects back to Skygear Auth, and finally redirect the user back to the application
+
+An introduction of workflow may be required to support some of the above use cases.
+
+## Security considerations
+
+### The session cookie is SameSite=Lax by default
+
+To make cookie visible in third-party context, the developer must set it to None.
```
kiootic: I think we don't have this configuration anymore.

370-auth-ui/sdk.md
```diff
@@ -0,0 +1,309 @@
...
+interface OAuthError {
+  // The state from AuthorizeOptions.
+  state?: string;
+  error: string;
+  error_description?: string;
+  error_uri?: string;
+}
+
+// The promise rejects with an JavaScript Error that looks like a OAuthError.
+function authorize(options: AuthorizeOptions): Promise<User>;
```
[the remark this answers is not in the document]
louischan-oursky: But the state is already given in `options`, returning it seems a bit redundant.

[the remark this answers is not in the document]
louischan-oursky: Anyway let me return state as well.

louischan-oursky:
Combined here https://github.com/SkygearIO/features/pull/425

PR #425, opened 2020-03-10
Auth UI

louischan-oursky:
ref #370

### DIFF
--- a/features/370-auth-ui/gateway.md
+++ b/features/370-auth-ui/gateway.md
@@ -1,0 +1,210 @@
+# Gateway
+
+Skygear Gateway is responsible for:
+
+- Resolve the app and the upstream server from the request host.
+- Delegate request authentication to Skygear Auth.
+- Proxy legacy request convention.
+
+## Resolving the app and the upstream server from the request host
+
+### Default domain
+
+The domain label pattern is `https://[<deployment_version_or_gear_name>.]<app_name>.<skygear_cluster_domain>`.
+
+The defined gear name are `accounts` and `assets`.
+
+Examples:
+
+- `https://698d0e9.myapp.skygearapp.com` resolves to the app `myapp` in deployment version `698d0e9`.
+- `https://accounts.myapp.skygearapp.com` resolves to Skygear Auth of the app `myapp`.
+
+### Custom domain
+
+The resolution of custom domain is purely table-driven. The developer is free to assign domain to their app and gears.
+
+## Delegating request authentication to Skygear Auth
+
+Skygear Gateway initiate subrequest to the resolve endpoint of Skygear Auth after resolving the app. The subrequest does not contain body. The gateway must remove `x-skygear-*` headers from the original request before initiating the subrequest. The gateway must copy `x-skygear-*` headers from the response of the subrequest to the original request.
+
+## Authenticating the request
+
+The resolve endpoint `/resolve` looks at `Authorization:` and `Cookie:` to authenticate the request. `Cookie:` has higher precedence. Regardless of the result, the status code is always 200. If the request is anonymous, the response does not contain any `x-skygear-*` headers. Otherwise the response contains the following headers.
+
+## HTTP Headers
+
+### x-skygear-session-valid
+
+Tell whether the session of the original request is valid.
+
+If this header is absent, it means that the original request does not contain any session.
+
+If the value is `true`, it indicates the original request has valid session. More headers will be included.
+
+If the value is `false`, it indicates that the original request has an invalid session.
+
+### x-skygear-user-id
+
+The user id.
+
+Example
+
+```
+x-skygear-user-id: 87dfaacf-a872-444a-948a-1497c6bb2a03
+```
+
+### x-skygear-user-verified
+
+Whether the user is considered verified or not.
+
+The value is either `true` or `false`.
+
+Example
+
+```
+x-skygear-user-verified: true
+```
+
+### x-skygear-user-disabled
+
+Whether the user is disabled or not.
+
+The value is either `true` of `false`.
+
+Example
+
+```
+x-skygear-user-disabled: false
+```
+
+### x-skygear-session-identity-id
+
+The ID of the identity the user authenticated as.
+
+Example
+
+```
+x-skygear-session-identity-id: 87dfaacf-a872-444a-948a-1497c6bb2a03
+```
+
+### x-skygear-session-identity-type
+
+What kind of identity the user authenticated as.
+
+The value is `password`, `oauth` or `custom_token`.
+
+Example
+
+```
+x-skygear-session-identity-type: password
+```
+
+### x-skygear-session-identity-updated-at
+
+The last time the identity was authenticated in this session. The application can use this to trigger reauthentication if it considers the freshness of the identity is stale.
+
+The value is in RFC3339 format.
+
+Example
+
+```
+x-skygear-session-identity-updated-at: 2019-09-17T00:00:00.000Z
+```
+
+### x-skygear-session-authenticator-id
+
+The ID of the authenticator the user used.
+If the session did not involve MFA, this header is absent.
+If the application has enabled MFA and enforcement is `off`, the presence of this header could mean the user has performed step-up MFA.
+
+Example
+
+```
+x-skygear-session-authenticator-id: 87dfaacf-a872-444a-948a-1497c6bb2a03
+```
+
+### x-skygear-session-authenticator-type
+
+What kind of authenticator the user used.
+If the session did not involve MFA, this header is absent.
+
+The value is `totp`, `oob`, `bearer_token` or `recovery_code`.
+
+Example
+
+```
+x-skygear-session-authenticator-type: oob
+```
+
+### x-skygear-session-authenticator-oob-channel
+
+The channel of the OOB authenticator.
+If the session did not involve MFA or the authenticator is not OOB, this header is absent.
+
+The value is `sms` or `email`.
+
+Example
+
+```
+x-skygear-session-authenticator-oob-channel: sms
+```
+
+### x-skygear-session-authenticator-updated-at
+
+The last time MFA was used in this session.
+The application can use this to trigger reauthentication if it considers the freshness of the authenticator is stale.
+
+The value is in RFC3339 format.
+
+Example
+
+```
+x-skygear-session-identity-updated-at: 2019-09-17T00:00:00.000Z
+```
+
+## Example
+
+### No session associated with the original request
+
+No headers are added.
+
+### Valid session associated with the original request
+
+```
+x-skygear-session-valid: true
+x-skygear-user-id: a
+x-skygear-user-verified: true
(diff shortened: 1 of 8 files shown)


features/370-auth-ui/overview.md
```diff
@@ -0,0 +1,97 @@
+# Auth UI, OIDC, and Sessions
```
carmenlau: @kiootic In order to make reader easier to follow maybe we could 

1. Add TOC in overview to list different sections?
2. Should we have section to highlight changes? For user who used skygear already, after reading the whole documents they may not realize that
    - api key is removed. we use client id for auth endpoint and api key is no required for other authorized request(e.g. update password, settings...etc). For user who used api key to take down the app, they should implement their own logic to force update app.
    - Custom token requires setup domain for auth gear too. We suggest user use session cookie for web app which is more secure, so custom domain with same etdl+1 domain for auth gear is required. That also implies that same etdl+1 domain cannot hold 2 app with different sessions

features/370-auth-ui/oidc.md
```diff
@@ -0,0 +1,452 @@
...
+
+```json
+{
+  "token_type": "bearer",
+  "access_token": "access.token",
+  "refresh_token": "refresh.token"
+}
+```
+
+## Authorization Server Metadata Document
```
kiootic: I think we should also document every endpoints (e.g. userinfo, jwks_uri), and note down which fields belongs to which document (I think both document can have same fields?)

features/370-auth-ui/web-ui.md
```diff
@@ -0,0 +1,179 @@
...
+- Manage OAuth authorizations
+- Manage OAuth grants
+- Update metadata
+- Change password
+- Configure multi-factor authentication
+- Manage identities (e.g. login IDs and connections with external IdPs)
+
+### The verification endpoint
+
+The verification endpoint `/verify` allows the user to verify their login ID. This page requires authenticated user. Unauthenticated access is redirected to the root endpoint with `redirect_uri=/verify`. Verification is not a strict requirement. For example, if auto send verification email is enabled, the email is sent behind the scene and the user is redirected to the application immediately after performing the Authorization Code Flow successfully.
```
louischan-oursky: > Carmen: Would like to know how user do verification for native app? User will complete the authorization flow without the app, and user will need to login again outside the app in browser?

# Option 1: Make the authorization endpoint enforce verification with an option

1. To let the developer know whether the user is considered verified or not, we should introduce a new claim `is_verified` to UserInfo.
2. In addition to [accepting Access Token in the UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo), the endpoint also accepts IdP session cookie.
3. Add SDK method to call the UserInfo endpoint. If Access Token is present, use it as bearer token, otherwise assume IdP session cookie is present.
4. The developer can then call the UserInfo endpoint via SDK to obtain a UserInfo Response, which contains `is_verified`.
5. If `is_verified` is false, the developer call `/oauth2/authorize?verify=true`, which enforce verification

# Option 2: Verification is mandatory

The only difference from Option 1 is that `verify=true` in Step 5 is unnecessary.

# Whether addLoginID and login ID verification is atomic

In the Auth API, the two are separate operations. An unverified login ID can be used in the same way as verified login ID without any restriction. Should we make adding login ID more similar to adding an authenticator? That is, the newly added login ID must be "verified" before it is usable. But only `email` and `phone` login ID type can be verified. `username` and `raw` can never be verified.

---

Thoughts? @carmenlau @kiootic

kiootic: I think in native app, after authentication, there would be a pair of access/refresh token in the app, and a session token in cookies of external browser. In this case, developer can simply call userinfo endpoint to check whether user is verified, then open verify endpoint in external browser directly if not verified.

louischan-oursky: > I think in native app, after authentication, there would be a pair of access/refresh token in the app, and a session token in cookies of external browser.

Step 2 is for web application, where no access token is obtained after authentication. But the developer still need to initiate a CORS request to the UserInfo endpoint.

> In this case, developer can simply call userinfo endpoint to check whether user is verified, then open verify endpoint in external browser directly if not verified.

So the authorization endpoint does not handle verification at all?

features/370-auth-ui/overview.md
kiootic: - I'll add a TOC linking the other documents
- I'll mention the removal of API key in overview document, but I don't think we need to mention the take down app use case, since it is not intended to use API key like this.
- I'll mention the expected custom domain setup in overview document.

louischan-oursky: I used to use [this](https://github.com/ekalinin/github-markdown-toc) to generate TOC

features/370-auth-ui/web-ui.md
kiootic: I think if we want to integrate user verification into authentication flow, we should require it automatically depending on server-side configuration, so both option 1 and 2 is not good enough. I think we don't need to support it at the moment?

louischan-oursky: Option 2 integrates user verification into the authentication flow so it is solely by server-side configuration. The problem is, the user can open the settings endpoint and add a unverified login ID. This could make the user become unverified. So even if we have option 2, the developer still have to check UserInfo in suitable places to ensure the user is verified.

kiootic: I think the role of OP is to assert the identity of user to RP. To avoid complicated logic at RP side and mistaken interpretation of claims, we should have a config flag for user verification (name for reference only):
- always_do_verification: always verify user in authentication flow, `*_verified` claims are always true.
- no_verify: don't verify user in authentication flow, disable verify APIs, `*_verified` claims are always false.
- use_api_to_verify: don't verify user in authentication flow, enable verify APIs, `*_verified` claims can be true/false.

I think we don't need to support it at the moment?

chpapa: I didn't read the discussion in details, only saw this at Slack... I think we can just let users use (admin?) API to verify first? And incorporate verification into auth flow later?

features/370-auth-ui/gateway.md
```diff
@@ -0,0 +1,210 @@
...
+
+```
+x-skygear-session-identity-id: 87dfaacf-a872-444a-948a-1497c6bb2a03
+```
+
+### x-skygear-session-identity-type
+
+What kind of identity the user authenticated as.
+
+The value is `password`, `oauth` or `custom_token`.
```
kiootic: Let's remove custom token.

features/370-auth-ui/oidc.md
carmenlau: 1. Updated to define endpoint for `userinfo ` and `jwks_uri `
1. Most of the fields are common in both documents, but there are some fields that only define in one of the spec.
    - [OIDC](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) only fields
        - `userinfo_endpoint`
        - `subject_types_supported`
        - `id_token_signing_alg_values_supported`
        - `claims_supported`
    - [OAuth2](https://tools.ietf.org/html/rfc8414#section-2) only fields
        - `code_challenge_methods_supported`
        - `revocation_endpoint`

In OAuth2 spec
>  Additional authorization server metadata parameters MAY also be used.
   Some are defined by other specifications, such as OpenID Connect
   Discovery 1.0 [OpenID.Discovery].

So for those OIDC only fields, we can include them in the OAuth2 metadata. For those OAuth2 only fields, we actually need them in oidc discovery document too (I assume we only use oidc discovery document in SDK), google one also have those fields.

```
https://accounts.google.com/.well-known/openid-configuration
```

So the content of two documents should be the same, thought?

carmenlau:
@chpapa Can you review the whole documents when have time? We are still updating but should be some minor comments which won't affect the main flow. Would like to ensure we don't have big problem, and we started implementing already.

features/370-auth-ui/oidc.md
louischan-oursky: OIDC is based on OAuth 2 so its metadata document could contains all OAuth 2 fields.

features/370-auth-ui/web-ui.md
louischan-oursky: @kiootic The reason I proposed option 1 is because the verification endpoint is like the authorization endpoint plus verification. It requires `redirect_uri` so it can redirect back to the application.

kiootic: I think option 1 would mix up the concerns of OAuth/OIDC and verification. The `authorize` endpoint is for obtaining consent from end-user to access protected resource (i.e. APIs in OAuth and identity information in OIDC); performing authentication is just a side-effect. It would be better to have a proprietary endpoint for presenting the verification UI if it does not fit into the authentication flow.

features/370-auth-ui/gateway.md
```diff
@@ -0,0 +1,210 @@
...
+
+### No session associated with the original request
+
+No headers are added.
+
+### Valid session associated with the original request
+
+```
+x-skygear-session-valid: true
+x-skygear-user-id: a
```
kiootic: I'm thinking of adding a header for session ID here, so backend services can call APIs (in future) to manipulate sessions. Any thoughts?

louischan-oursky: Maybe add it in the future when we actually have admin API.

features/370-auth-ui/web-ui.md
louischan-oursky: See if https://github.com/SkygearIO/features/pull/425/commits/b2f509b807dd7cf5e9deb53bd805f21d2344fec0 captures the conclusion here.

chpapa:
@carmenlau sorry I am very busy these few days (like 1am most days) for business work... I probably don't have 3 hours to sit down and read things bit by bit.

Do you want to say have a 30 mins walk through with me tomorrow (11 Mar)? And see if @kiootic and @louischan would like to join?

carmenlau:
We need to add a logout function for user who used authui. We cannot use the same logout function, since the implementation is totally different. We have 2 ideas,

1. auth ui and auth api are 2 different packages.
2. auth ui and auth api are in different namespace. e.g. 

    auth ui functions:
    ```
    skygear.auth.startAuthorization
    skygear.auth.finishAuthorization
    skygear.auth.logout
    ```
    
    auth api functions:
    ```
    skygear.classicAuth.login
    skygear.classicAuth.signup
    skygear.classicAuth.logout
    ```

We had a discussion and due to time being, we will pick approach 2 first. We will move auth ui related functions in a new namespace `authui`. We can discuss and decide in production meeting.

@louischan-oursky @kiootic @chpapa @keithtkl

chpapa:
@carmenlau I think the answer depends on if developer will use the auth api functions with ui a lot? Keep 2 packages will make it simpler for people who only use auth ui or auth api, less distraction. I don't have a good answer now... so maybe just go with the easier solution first.

louischan-oursky:
I added a commit to specify how do the developer can configure country calling codes.

PR #431, opened 2020-04-02
Enhanced template

louischan-oursky:


features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
+
+All templates have default value so template resolution always succeed.
+
+The templates are first resolved by matching the type and the key. And then select the best language according to the user preferred languages.
+
+## Component templates
+
+Some template may depend on other templates which are included during rendering. This enables customizing a particular component of a template.
+
+For example, `auth_ui_login.html` depend on `auth_ui_header.html` and `auth_ui_footer.html` to provide the header and footer. If the developer just wants to customize the header, they do not need to provide customized templates for ALL pages. They just need to provide `auth_ui_header.html`.
```
kiootic, line 51: How to prevent excessive recursion (e.g. billion laugh attack for XML) if we allow user template to include another template?

features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
-If the developer has configured `email` and `secondary_email` as login IDs. User Verification supports verifying both with different templates. Therefore, `key` is required to differentiate different instances of the same type of the template.
+In addition to the template language tag, sometimes it is preferred to localize the text of the template rather the whole template.
+
+For example, `auth_ui_login.html` defines the HTML structure and is used for all languages. What the developer wants to localize is the text.
+
+A special function named `localize` can be used to format a localized string.
+
+```html
+<input type="password" placeholder="{{ localize "enter.password" }}">
+<!-- <input type="password placeholder="Enter Password">
```
kiootic, line 63: The comment tag is not closed and messed up the highlighting...

features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
+```
+
+```html
+<p>{{ localize "email.sent" .email .name }}</p>
+<!-- <p>Hi John, an email has been sent to john.doe@example.com</p -->
+```
+
+`localized` takes a translation key, followed any arguments required by that translation key. If the key is not found, the key itself is returned.
```
kiootic, line 71: I suppose in future, we need some mechanism to inform developer of the mistake (e.g. mistyped key, missing arguments).

features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
+The translation file itself is a template so it is loaded as template as usual. For example:
+
+```
+templates/auth_ui_translation.json
+templates/zh-Hant/auth_ui_translation.json
+templates/zh-Hans/auth_ui_translation.json
+templates/ja/auth_ui_translation.json
+```
+
+The translation file is simply a flat JSON object with string keys and string values. The value is in ICU MessageFormat. Not all ICU MessageFormat arguments are supported. The supported are `select`, `plural` and `selectordinal`.
```
kiootic, line 82: I remember we have fallback mechanism for a whole template (i.e. if `/zh-Hant/template.html` does not exists, fallback to `/template.html`), but do we want to have fallback mechanism per-key and per-language?
- e.g. if `email.sent` is not found in `/zh-Hant/translation.json`, fall back to `/translation.json`
- e.g. if `email.sent` is not found in `/zh-Hant/translation.json`, fall back to `/zh/translation.json`

features/251-templates/go-i18n-research.md
```diff
@@ -0,0 +1,63 @@
+# How could one do i18n in Go
+
+This document concludes the research on how to do i18n in Go as of 2020-03.
```
kiootic, line 3: Do we have a conclusion on how to do it at the moment?

features/251-templates/index.md
louischan-oursky: Internally, the template spec must declare the templates they depend on. Since the template spec is written by us, we can check the AST to ensure user cannot include an unlisted template. We should also ensure template do not have cycle.

features/251-templates/go-i18n-research.md
louischan-oursky: Sorry. I should have mentioned that I tend to pick the last choice: Implement MessageFormat ourselves.

features/251-templates/index.md
louischan-oursky: Good point! Translation resolution should be different from template resolution. Let me clarify it.

features/251-templates/index.md
louischan-oursky: The best way is to let the developer to "preview" the rendered template.

features/251-templates/index.md
kiootic: I understand it correctly, that means each template type have a whitelisted of template types that it could include. That is, `auth_ui_login.html` can import only `auth_ui_header.html` and `auth_ui_footer.html`, and `auth_ui_header.html`/`auth_ui_footer.html` cannot import any template?

features/251-templates/index.md
louischan-oursky: which could be done by creating a staging app and test their templates there...

features/251-templates/index.md
louischan-oursky: They can, as long as there is no cycle.

kiootic: It's just that we'll need to inspect the content of all templates to check whether the template reference graph is a DAG.

louischan-oursky: I think we do not need to detect cycle with graph. The dependency is static and is code reviewed. At parsing time, we just need to check if the template has `{{template name}}` where `name` is not listed.

kiootic: But template is customizable? e.g. developer can update `auth_ui_login.html` to include `auth_ui_header.html`, and update `auth_ui_header.html` to include `auth_ui_login.html`? Or do we maintain a whitelist of which template can include what templates?

louischan-oursky: > Or do we maintain a whitelist of which template can include what templates?

Yes. The template spec would have its `Defines` replaced by `Dependencies []TemplateItemType`.

louischan-oursky: If the developer wants to change the structure of `auth_ui_login.html` entirely, like instead of `auth_ui_header.html` and `auth_ui_footer.html`, their custom template expects `auth_ui_sidebar.html` and `auth_ui_nav.html` then dependencies must not be declared in the template spec and we need to use graph algorithm to detect cycles.

I think we can leave this to future.

features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
+## Component templates
+
+Some template may depend on other templates which are included during rendering. This enables customizing a particular component of a template.
+
+For example, `auth_ui_login.html` depend on `auth_ui_header.html` and `auth_ui_footer.html` to provide the header and footer. If the developer just wants to customize the header, they do not need to provide customized templates for ALL pages. They just need to provide `auth_ui_header.html`.
+
+## Localization of the text of the template
 
-If the developer has configured `email` and `secondary_email` as login IDs. User Verification supports verifying both with different templates. Therefore, `key` is required to differentiate different instances of the same type of the template.
+In addition to the template language tag, sometimes it is preferred to localize the text of the template rather the whole template.
```
carmenlau, line 55: If there is a case that user want to have different logo per language. So they can provide `auth_ui_header.html` only for different languages? e.g. 
- `/auth_ui_header.html`
- `/zh-Hant/auth_ui_header.html`

I think it is great enough :)

features/251-templates/index.md
```diff
@@ -2,57 +2,92 @@
...
+
+A special function named `localize` can be used to format a localized string.
+
+```html
+<input type="password" placeholder="{{ localize "enter.password" }}">
+<!-- <input type="password placeholder="Enter Password"> -->
+```
+
+```html
+<p>{{ localize "email.sent" .email .name }}</p>
```
carmenlau, line 67: We assume the localize json should only have text. So `localize` here will do html escape right?

features/251-templates/index.md
louischan-oursky: The logo is localized differently. `logo_uri` is a OIDC client metadata and we definitely should support something like `logo_uri#zh`, `client_name#zh`.

features/251-templates/index.md
louischan-oursky: Good point! I guess html/template has handled this for us. Let me double check.

louischan-oursky: html/template is smart enough to escape function output for us :)

Issue #430, opened 2020-04-01
Password-less login


kiootic:
## Description

Developer may want to enable user to login without using password. Instead, an OTP would be sent to user through email/SMS as temporary login credentials.

- Refactor login ID & password
- Password-less authentication method
    - Activation
    - Message template
    - OTP verification
- Interactions
    - with password login
    - with authentication session & MFA
    - with user verification

## Portal Design

**Remove this section if the feature have no Portal**

## Blog Post Specification

**Blog Post of the Feature Release**

## Open Questions

**Put a list of open questions here before a complete design / specification is decided**

# Related Issues

- Server Issues
- Client Issues
- Guides Issues

PR #432, opened 2020-04-02
Passwordless authentication

kiootic:
ref #430


features/430-passwordless/design.md
```diff
@@ -0,0 +1,204 @@
...
+
+For examples:
+- Simple password login:
+    1. Enter Login ID.
+    2. Authenticate with password.
+    3. Done.
+- Password/SSO with email/TOTP MFA:
+    1. Select Login ID or OAuth identity.
+        - Login ID: enter email.
+        - OAuth: redirect to OAuth provider and wait for return.
```
louischan-oursky, line 135: Ben in the past told me a feature like this. After entering the email, if the email matches one of the OAuth identity the user has, redirect the user to the OAuth provider. So what the user really enters is a hint which can be interpreted as login ID or a reference to one of the OAuth identity they have. Anyway it is an improvement and (I think) no need to be done at this stage.

louischan-oursky: Should primary authenticator have priority? A recurring requirement I have been told is that the user initially signed up with `user@gmail.com`, later on the user connected their gmail account. From now on the user must use google to login and cannot use password anymore.

features/430-passwordless/implementation.md
```diff
@@ -0,0 +1,40 @@
...
+      identity. These authenticators should always fail the authentication.
+- Perform secondary authentication: `mfa:authn`
+    - A list of usable secondary authenticators (set up for the user) would be
+      available for selection.
+    - At this step, we have sufficient confidence about user's identity to list
+      usable authenticators of the user.
+- Setup secondary authentication: `mfa:setup`
+    - User must setup one secondary authenticator to proceed.
+
+For Auth API, client will always recieve authentication session after
```
louischan-oursky, line 33: Not sure what this means. I think this statement is only applicable to app that has MFA turned on?

features/430-passwordless/implementation.md
```diff
@@ -0,0 +1,40 @@
...
+`identity:authn` step. i.e. `mfa:authn` or `mfa:setup`.
+
+## API models
+
+Previously, we have two types of identities: `password` and `oauth`. Under the
+new design, `password` type should be renamed to `login_id`. However, for
```
louischan-oursky, line 39: I think we should better rename this. Let's do a search to see if any project depend on the value of `password`.

features/430-passwordless/design.md
kiootic: > After entering the email, if the email matches one of the OAuth identity the user has, redirect the user to the OAuth provider.

Yes, it is intended to be possible to support that. A more generalized formalization is that, in this step user gives a set of claims (e.g. `{email: user@example.com}` for email login ID, `{oauth_provider: google}` for OAuth), then the system would match the claims to find a candidate identity, and it is up to the frontend to interpret user input into this set of claims.

We definitely do not need this right now.

> A recurring requirement I have been told is that the user initially signed up with user@gmail.com, later on the user connected their gmail account. From now on the user must use google to login and cannot use password anymore.

We should not have this problem, since user can continue to enter login ID and password to login, as long as the login ID identity and password authenticator is intact, and system is configured to allow these identity and authenticator types to be used.

features/430-passwordless/implementation.md
kiootic: Yes, I should clarify it to be `if client received an authentication session, it'll always have steps after...`.

features/430-passwordless/design.md
louischan-oursky: Sorry, I meant the requirement is:

As long as the user has an OAuth Identity whose claims match one of their login ID, then the login ID must not be used anymore. Further login must use the OAuth identity.

kiootic: If so, I think using priority is not appropriate in this scenario: the use case wants to 'forbid' login ID identity to be used, not 'prioritize' OAuth identity. Misusing semantics of features can lead to future breaks.

For this use case, I'll suggest the developer to delete the login ID identity/password authenticator in post create hook of OAuth identity.

features/430-passwordless/implementation.md
carmenlau: I think we used it in `Identity` object and the gateway header? Maybe we can just ask if anyone is using it, if no then we change it?🙈

features/430-passwordless/design.md
```diff
@@ -0,0 +1,204 @@
...
+To indicate authenticator used in authentication, `amr` claim is used in OIDC
+ID token.
+
+`amr` claim is an array of string. It includes authentication method used:
+- If password authenticator is used: `pwd` is included.
+- If any OTP (TOTP/OOB-OTP) is used: `otp` is included.
+- If WebAuthn is used: `hwk` is included.
+
+If no authentication method is to be included in `amr` claim, `amr` claim would
+be omitted from the ID token.
```
carmenlau, line 183: Would like to know which case the id token will not have the amr?

features/430-passwordless/implementation.md
carmenlau: Also for the current headers, we had authenticator. We will need to update those headers too to support list?

features/430-passwordless/design.md
louischan-oursky: For the hook approach to work, the developer need to has access to Admin API that allows them to query additional information about the user. For example in `after_identity_create`, the developer needs to query all identities the user, to see if the identities match the condition.

kiootic: Yes, that's correct. The simplest approach is to get the list of identities and delete login ID identity, if the created identity is an OAuth identity. We're going to have proper admin APIs right?

features/430-passwordless/implementation.md
kiootic: It is possible to maintain full backward compatibility, by keeping the 'password' identity type and fill in authenticator headers using secondary authenticator only. The question is do we want to keep it that way.

features/430-passwordless/design.md
kiootic: When user is logged in with OAuth only, there will be no `amr` claim.

features/430-passwordless/design.md
louischan-oursky: > We're going to have proper admin APIs right?

Yes

---

Just got this idea in my mind. We could have some implemented hooks to solve some common problems. The developer can just enable the hooks in config.

This feature and the OnUserDuplicate flag could be implemented by hook.

kiootic: > Just got this idea in my mind. We could have some implemented hooks to solve some common problems. The developer can just enable the hooks in config.
> 
> This feature and the OnUserDuplicate flag could be implemented by hook.

I'm not sure if that's a good idea:
- Hooks are supposed to be extension point for external services, and handling hook events by ourselves seems strange.
- Once implemented, we cannot have breaking changes to the behavior of built-in hooks.
- If there is a bug in built-in hook handlers, either user should update to latest version with the fix, which may have undesired changes, or we back-port the fix to previous versions, and neither one seems ideal.
- It would tend to be misused, if it is the only endpoint exposing certain behaviors.
- We need to have a clear criteria on what problems are 'common' enough to have the hooks included in-repo, since everybody would certainly want us to maintain the hooks.

features/430-passwordless/config.md
```diff
@@ -1 +1,71 @@
...
+
+- `identities`: string list, must be non-empty.
+                List of usable identity types.
+- `primary_authenticators`: string list, must be non-empty.
+                            List of primary authenticator types.
+- `secondary_authenticators`: string list, can be non-empty.
+                              List of secondary authenticator types.
+
+Identity and authenticator types must be included in one of the lists to be
+enabled. There must be no duplicated authenticators across primary and
```
louischan-oursky, line 73: Do we need to define the authentication factors?

Suppose we agree that there are 3 authentication factors

- something the user has (A)
- something the user knows (B)
- something the user is (C)

The next step is to classify our authenticators into one of these factors

- oob: A
- totp: A
- bearer_token: A
- recovery_code: A
- anonymous: A
- webauthn: A
- password: B
- biometric: C
- oauth: C

The next question is should we enforce the two authenticators must be of different factors? Or we allow authenticators of the same factor as long as the authenticators are distinct?

If we allow the latter case then the configuration should allow duplicated authenticator types.

features/430-passwordless/design.md
```diff
@@ -169,19 +183,97 @@ For examples:
...
+authorize endpoint.
+
+### `login_hint` parameter
+
+Developer can optionally pre-select the identity to use using `login_hint`
+parameter. For examples:
+- To authenticate with email, `login_hint` should be
+  `login_id:email` or `login_id:email:<email address>`.
+- To authenticate with Google OAuth, `login_hint` should be `oauth:google` or
+  `oauth:google:<subject ID>`.
```
louischan-oursky, line 278: Instead of `login_id:` and `oauth:` we should have `https://auth.skygear.io/login_hint?type=claims&email=email` and `https://auth.skygear.io/login_hint?type=attestation&data=...`. Using URL allow us to add extra fields in the future. Can we specify the exact format we are going to support?

features/430-passwordless/config.md
kiootic: I don't think we need to enforce this; the three factor kinds are not defined strictly (e.g. a password can be considered something user knows, but a written down password can be considered something user has), and mostly used for explanatory purpose.

kiootic:
@louischan-oursky Updated, thanks.

features/430-passwordless/implementation.md
```diff
@@ -38,14 +38,18 @@ one of the steps after `identity:authn` step. i.e. `mfa:authn` or `mfa:setup`.
...
+
 ## Session Headers
 
 The authenticator headers would be removed and replaced with the following
 headers:
 - `X-Skygear-Session-Acr`: `acr` claim, URI value.
 - `X-Skygear-Session-Amr`: `amr` claim, space separated values.
 
-Also, `X-Skygear-Session-Identity-Updated-At` header would be removed.
+Also, `X-Skygear-Session-Identity-Id` and `X-Skygear-Session-Identity-Updated-At`
```
louischan-oursky, line 51: Rebase onto master and

- Update `370-auth-ui/gateway.md` to remove the mentioned headers
- Grep `currentIdentity` and update it as well. Not sure if we should rename it or simply remove it. If we remove it, we can remove `Identity` from `AuthResponse`.

kiootic: I don't think we should modify existing specs at the moment, otherwise we'll have no context on the changes?

I think we can keep the `Identity` object and just remove the `id` field.

louischan-oursky: > I think we can keep the Identity object and just remove the id field.

One type or two types? `listIdentities` still return identities with IDs? Do we remove the `id` field at HTTP API level or only hide it in the SDK?

kiootic: Remove it everywhere, including all APIs, hooks, HTTP headers, and any other places. Since we're not accepting identity ID in our APIs (they're just for convience of clients), we should be able to remove them without breaking API input interfaces.

kiootic:
Added detailed design for anonymous user authentication & promotion.

features/000-anonymous-user/design.md
```diff
@@ -0,0 +1,112 @@
...
+either have no anonymous identity, or have exactly one anonymous identity
+without other types of identity.
+
+Anonymous identity has fields:
+- public key: the public key represented as a JWK.
+- key ID: a random string (`^[-\w]{8,64}$`), used for efficient lookup.
+
+### Authentication
+
+To signup as anonymous user:
```
louischan-oursky, line 22: Is this flow compatible with WebAuthn register flow?

features/000-anonymous-user/design.md
```diff
@@ -0,0 +1,112 @@
...
+    https://accounts.skygear.test/oauth2/authorize?xxx&redirect_uri=app://on_authorized&login_hint=https%3A%2F%2Fauth.skygear.io%2Flogin_hint%3Ftype%3Danonymous%26jwt%3DeyJhbGciOiJSUzI1Nxxx
+    ```
+- Client SDK redirects users to the authorization endpoint.
+- Server validates the self-signed JWT in login hint parameter, and validate
+  the action in JWT payload (signup).
+- Server creates a new user with anonymous identity, using the specified
+  public key.
+- Server rediercts to the redirect URI with an authorization code.
+
+To authenticate as anonymous user:
```
louischan-oursky, line 39: Is this flow compatible with WebAuthn login flow?

features/000-anonymous-user/sdk.md
```diff
@@ -0,0 +1,27 @@
+# SDK
+
+## Authentication
+
+APIs would be added to perform authentication:
+```ts
+function authenticateAnonymously(): Promise<{ user: User }>;
+```
+
+This function would log in as an existing anonymous user if a local key-pair is
```
louischan-oursky, line 10: What would happen if there is a non-anonymous session?

features/000-anonymous-user/sdk.md
```diff
@@ -0,0 +1,27 @@
+# SDK
+
+## Authentication
+
+APIs would be added to perform authentication:
+```ts
+function authenticateAnonymously(): Promise<{ user: User }>;
```
carmenlau, line 7: If I understand correctly, we will use browser-less flow for signup and with browser flow for promote. So this function should be step of browser-less flow signup, right?

Would like to know will this function behaves differently when the client app is first party or third party app? `response_type=none` for first party and `response_type=code` for third party app?

features/000-anonymous-user/design.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Anonymous identity
+
+A new identity type `anonymous` would be added. It is expected that users would
+either have no anonymous identity, or have exactly one anonymous identity
+without other types of identity.
+
+Anonymous identity has fields:
+- public key: the public key represented as a JWK.
+- key ID: a random string (`^[-\w]{8,64}$`), used for efficient lookup.
```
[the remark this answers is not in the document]
kiootic: The key ID is generated client side. Most likely it would be implemented as a UUID.

The key ID is in `kid` field of the JWK. You can see it in the example JWTs.

features/000-anonymous-user/design.md
kiootic: Not necessarily. WebAuthn has different security properties as anonymous user, most notably the key pair may not be stored securely on web platform. Also, WebAuthn can be used on web platform only.

features/000-anonymous-user/sdk.md
kiootic: The existing session would be overwritten by the new session without logging out using token invalidation API.

features/000-anonymous-user/sdk.md
kiootic: > So this function should be step of browser-less flow signup, right?

Yes, the SDK only uses browser-less flow.

> Would like to know will this function behaves differently when the client app is first party or third party app? response_type=none for first party and response_type=code for third party app?

Yes, the behavior is same as normal authorization.

kiootic:
@carmenlau @louischan-oursky I've simplified the browser-less flow to use OAuth token endpoint directly, please check if it makes sense to you.

features/000-anonymous-user/design.md
louischan-oursky: My main concern is that we do not invent a protocol ourselves but reference to a similar and existing protocol which has been studied, especially in the security aspect.

features/000-anonymous-user/sdk.md
louischan-oursky: After the simplification, native application SDK will use the browser-less flow for both signup and login and web application SDK will use the authorization code flow, am I correct?

features/000-anonymous-user/design.md
kiootic: The anonymous user flows are largely taken from WebAuthn. However it is not fully compatible with WebAuthn for simplicity, e.g. using JSON instead of CBOR, not distinguishing between attestation and assertion, etc.

features/000-anonymous-user/sdk.md
kiootic: Yes, native application would use browser-less flow. However, for web platform, it is not yet fully specified as I don't have a concrete use-case to demonstrate how would users like the flow to be.

